PR #26, opened 2022-01-15
extend the transaction type to handle plutus scripts, datum, etc

Benjmhart:
fixes https://github.com/Plutonomicon/cardano-browser-tx/issues/3


src/Types/Transaction.purs
```diff
@@ -20,27 +29,129 @@ newtype TxBody = TxBody
...
+
+newtype ScriptHash = ScriptHash String
+
+newtype Costmdls = Costmdls (Map Language CostModel)
+
+data Language = PlutusV1
+
+newtype CostModel = CostModel (Array Int)
+
+type ProtocolVersion =
```
ngua, line 116: Would it make sense to use `UInt`s for these fields? Assuming we want to track the serialization-lib's representation as closely as possible

src/Types/Transaction.purs
```diff
@@ -114,61 +249,9 @@ newtype Address = Address
   }
 
 newtype BaseAddress = BaseAddress
-  { network :: Int, -- u8,
+  { network :: UInt, -- UInt8
```
ngua, line 252: Since we already depend on `arraybuffer-types`, perhaps we could make this a `UInt8` as well

src/Types/Transaction.purs
```diff
@@ -20,27 +29,129 @@ newtype TxBody = TxBody
...
+  { proposed_protocol_parameter_updates :: ProposedProtocolParameterUpdates
+  , epoch :: Epoch
+  }
+
+newtype ProposedProtocolParameterUpdates 
+  = ProposedProtocolParameterUpdates (Map GenesisHash ProtocolParamUpdate)
+
+newtype GenesisHash = GenesisHash String
+
+type ProtocolParamUpdate =
```
ngua, line 60: Perhaps these `Int`s could be `UInt`s, going off of serialization-lib

ngua (commented):
There are a few places where we might want to use an unsigned integer type instead of an `Int` to more closely track serialization-lib's representations. What do you think?

Benjmhart:
Agree!  I'll make the adjustment.  I wish the libraries offered safe UInt8,
32 representations. - I feel like we just have thin wrappers here.  Math
might screw us.   But in the short Term this is fine.

src/Types/Transaction.purs
Benjmhart: https://pursuit.purescript.org/packages/purescript-arraybuffer-types/3.0.1/docs/Data.ArrayBuffer.Types#t:Uint8

in short, UInt8 isn't of kind `Type` it's of kind `ArrayViewType` and `ArrayView` is the only thing of kind `ArrayViewType -> Type`.

we're stuck with UInt

vvtran:
Will `BaseAddress` and `Address` need realigning with https://github.com/Plutonomicon/cardano-browser-tx/pull/31?

ngua:
@vvtran I don't think so. The address types in #31 are the representations in the serialization-lib. They're opaque and can only be used across the FFI

ngua (approved):
Time to merge this I think

Issue #9, opened 2021-12-27
provide the correct way to go from addr1 format to pubkeyhash


Benjmhart:
nami wallet provides addr1-formatted addresses

in previous attempts to integrate with Nami, we had difficulty getting the _correct_ pubkeyhash out of this address.

we may need to get the right library functions to do this correctly.

mikekeke:
A I understand, Nami gives address that built from `Credential` and `StakingCredential`: https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Address.html#Address

`Credential` could be `PubKeyHash` or `ValidatorHash`: https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Credential.html#Credential

So, when we parse `Address`, we can get `PubKeyHash` from it, but we can not build back same `Address` from `PubKeyHash` without `StakingCredential`, it it wasn't `Nothing`. So if we are using address with `StakingCredential` like Nami provides, we can't build that address only from `PubKeyHash` .

adamczykm:
Can't yet assign myself.

nrutledge:
Yeah, there is no issue going from addr1 format to pubKeyHash. Going from pubKeyHash back to addr1 is the issue due to loss of StakingCredential as Misha stated. Solution seems to be to always store the full address as opposed to just the PubKeyHash.

adamczykm:
### **Getting PubKeyHash**

There exists a function in nami-wallet api that allows to extract a PKH from BaseAddress:

```jsx
export const extractKeyHash = async (address) => {
  await Loader.load();
  //TODO: implement for various address types
  if (!(await isValidAddressBytes(Buffer.from(address, 'hex'))))
    throw DataSignError.InvalidFormat;
  try {
    const baseAddr = Loader.Cardano.BaseAddress.from_address(
      Loader.Cardano.Address.from_bytes(Buffer.from(address, 'hex'))
    );
    return baseAddr.payment_cred().to_keyhash().to_bech32('hbas_');
  } catch (e) {}
  try {
    const rewardAddr = Loader.Cardano.RewardAddress.from_address(
      Loader.Cardano.Address.from_bytes(Buffer.from(address, 'hex'))
    );
    return rewardAddr.payment_cred().to_keyhash().to_bech32('hrew_');
  } catch (e) {}
  throw DataSignError.AddressNotPK;
};
[...]

PR #31, opened 2022-01-18
Proppose API for BaseAddress and PubKeyHash conversion.

adamczykm:
Addresses #9 
Please let me know your thoughts. 
What next steps to consider this PR ready to merge?


adamczykm:
cc. 
@ngua @Benjmhart

Benjmhart:
looks good on my end

ngua:
Looks good to me too

vvtran:
A question, are we using the convention to prefix the module name by the folder structure or not? It doesn't make it any less usable of course. Here we have `module Address` instead of `module Types.Address`.

adamczykm:
> A question, are we using the convention to prefix the module name by the directory structure or not? It doesn't make it any more usable of course. Here we have `module Address` instead of `module Types.Address`.

I like `Types.Address` better. Will change.

src/Types/Address.js
```diff
@@ -0,0 +1,55 @@
...
+// addressPubKeyHash :: BaseAddress -> PubKeyHash
+exports.addressPubKeyHash = baseAddr => {
+    // i've chosen a prefix that Nami uses for payment_creds
+    return baseAddr.payment_cred().to_keyhash().to_bech32('hbas_');
+};
+
+//addressStakeKeyHash :: BaseAddress -> StakeKeyHash
+exports.addressStakeKeyHash = baseAddr => {
+    // i've chosen an arbitrary prefix here
+    return baseAddr.stake_cred().to_keyhash().to_bech32('hstk_');
```
klntsky, line 54: `to_keyhash()` has a possibility of failing. Should be inside Effect I think. Same as above.

src/Serialization/Address.js
```diff
@@ -0,0 +1,63 @@
+const CardanoWasm = require("@emurgo/cardano-serialization-lib-nodejs");
+
+
+// newBaseAddress :: NetworkId -> PubKeyHash -> StakeKeyHash -> BaseAddress
+exports.newBaseAddress = netId => pkhBech32 => skhBech32 => {
```
ngua, line 5: This is also partial, so I think that we should reflect that in the type signature. `from_bech32()` might throw several different exceptions if the input if malformed (we don't have a smart constructor for `Bech32String` or another guarantee that it will be correct). We could just make this an `Effect` and do away with the checks in JS, but you took the approach of returning `Maybe`s in some implementations below, which I think is fine, so maybe we can do that here too

Perhaps a similar approach to the one you take in `fromBech32Impl` would work here, e.g.

```js
// newBaseAddress :: 
//   (forall a. a -> Maybe a)
//   (forall a. Maybe a)
//   NetworkId -> 
//   PubKeyHash -> 
//   StakeKeyHash -> 
//   Maybe BaseAddress
exports.newBaseAddress =
  just => nothing => netId => pkhBech32 => skhBech32 => {
    try {
      const pkh = CardanoWasm.Ed25519KeyHash.from_bech32(pkhBech32);
      const skh = CardanoWasm.Ed25519KeyHash.from_bech32(skhBech32);
      // null checks... (what if both are null?)
      const baseAddr = CardanoWasm.BaseAddress.new(
        netId,
        CardanoWasm.StakeCredential.from_keyhash(pkh),
        CardanoWasm.StakeCredential.from_keyhash(skh)
      );
      return just(baseAddr);
    } catch (err) {
      // maybe log it
      return nothing;
    }
  };
```

src/Serialization/Address.js
```diff
@@ -0,0 +1,63 @@
...
+    pkh = CardanoWasm.Ed25519KeyHash.from_bech32(pkhBech32);
```
ngua, line 6: Missing a declaration here

src/Serialization/Address.js
```diff
@@ -0,0 +1,63 @@
...
+        return nothing;
+    }
+};
+
+// addressBech32 :: BaseAddress -> Bech32String
+exports.addressBech32 = baseAddr => {
+    return baseAddr.to_address().to_bech32();
+};
+
+// addressPubKeyHash :: (forall x.x -> Maybe x) -> (forall x.Maybe x) -> BaseAddress -> PubKeyHash
```
ngua, line 45: Missing a `Maybe` in the PS type sig comment

src/Serialization/Address.js
```diff
@@ -0,0 +1,63 @@
...
+exports.addressPubKeyHashImpl = just => nothing => baseAddr => {
+    // i've chosen a prefix that Nami uses for payment_creds
+    const kh = baseAddr.payment_cred().to_keyhash();
+    if(kh==null){
+        return nothing;
+    }
+    return just(kh.to_bech32('hbas_'));
+};
+
+// addressStakeKeyHash :: (forall x.x -> Maybe x) -> (forall x.Maybe x) -> BaseAddress -> StakeKeyHash
```
ngua, line 55: Same as above

ngua (commented):
Looks pretty good, I'm not sure how you want to approach the partiality in `NewBaseAddress`, so I'd like to hear your thoughts on that. There are a few places where you're missing a declaration before assignment (`newBaseAddress` and `fromBech32Impl`)

adamczykm:
@vvtran @ngua @klntsky 
I have thought about it a little bit and instead propose something like this:

```purescript
---- assumed this data types and that they are correct by construction
data Ed25519KeyHash
data ScriptHash
newtype Bech32String = Bech32String String
data ByteArray
----

data NetworkTag = TestnetTag | MainnetTag

data StakeCredPart
foreign import data Payment :: StakeCredPart
foreign import data Delegation :: StakeCredPart

newtype StakeCred p t = StakeCred t

type PaymentKeyHash = StakeCred Payment Ed25519KeyHash
type StakeKeyHash = StakeCred Delegation Ed25519KeyHash

-- | Shelley-era address where payment and delegation parts
-- | are each either a `Ed25519Keyhash` or a `ScriptHash`
newtype BaseAddress (p :: Type) (d :: Type) = BaseAddress
  { network :: NetworkTag
  , payment :: StakeCred Payment p
  , delegation :: StakeCred Delegation d
  }

newtype CredType (p :: Type) = CredType {bitSet :: Maybe Boolean}

---- address <->

addressBech32 :: forall p d. BaseAddress p d -> Bech32String
addressBech32 = undefined

addressFromBech32 :: forall p d. {payment :: CredType p, delegation :: CredType d} -> Bech32String -> Maybe (BaseAddress p d)
addressFromBech32 = undefined

addressBytes :: forall p d. BaseAddress p d -> ByteArray
addressBytes = undefined

addressFromBytes :: forall p d. {payment :: CredType p, delegation :: CredType d} -> ByteArray -> Maybe (BaseAddress p d)
addressFromBytes = undefined

---- stake creds

stakeCredFromKeyHash :: forall (p :: StakeCredPart). Proxy p -> Ed25519KeyHash -> StakeCred p Ed25519KeyHash
stakeCredFromKeyHash = undefined

stakeCredFromScriptHash :: forall (p :: StakeCredPart). Proxy p -> ScriptHash -> StakeCred p ScriptHash
stakeCredFromScriptHash = undefined

unStakeCred :: forall p t. StakeCred p t -> t
unStakeCred = undefined


---- cred types

anyCredType :: CredType (forall a. a)
anyCredType = CredType {bitSet: Nothing}

ed25519KeyHashCredType :: CredType Ed25519KeyHash
ed25519KeyHashCredType = CredType {bitSet: Just false}

scriptHashCredType :: CredType ScriptHash
scriptHashCredType = CredType {bitSet: Just true}

```

ngua:
@adamczykm I actually like that proposal more than the existing encoding of `BaseAddress`, for a couple of reasons

- we can keep all or almost all FFI calls to the serialization-lib in a single place (i.e. Vladimir's work)
- it is closer to the specification on addresses
- it's safer and more extensible, as you point out

I assume that the constructor for `BaseAddress` won't be exported, right? Otherwise we could construct something non-sensical

How long do you think it would take to complete that? It would be nice to merge this and @vvtran's PR, which depends on this as well

vvtran:
@adamczykm The proposal looks fine to me.

adamczykm:
I don't know how long it'd take to get it done. It won't be anything problematic if not for these functions:
```purescript
addressBech32 :: forall p d. BaseAddress p d -> Bech32String

addressFromBech32 :: forall p d. {payment :: CredType p, delegation :: CredType d} -> Bech32String -> Maybe (BaseAddress p d)

addressBytes :: forall p d. BaseAddress p d -> ByteArray

addressFromBytes :: forall p d. {payment :: CredType p, delegation :: CredType d} -> ByteArray -> Maybe (BaseAddress p d)
```
which might be time-consuming to implement by hand, we could use CSL but they don't expect their inputs to be valid and therefore throw errors. To make it faster to merge I could implement tests in another PR.

ngua:
OK, if you feel comfortable implementing those functions I'd so go for the new proposal

package.json
```diff
@@ -13,9 +13,15 @@
...
   "dependencies": {
     "@cardano-ogmios/client": "^4.2.1",
+    "@emurgo/cardano-serialization-lib-nodejs": "^9.1.2",
+    "bech32": "^2.0.0",
     "big-integer": "^1.6.51",
     "bufferutil": "^4.0.5",
     "uniqid": "^5.4.0",
     "ws": "^8.4.0"
+  },
+  "devDependencies": {
```
ngua, line 23: `purescript-language-server` and `pscid` should be provided by `easy-purescript-nix`

package.json
```diff
@@ -13,9 +13,15 @@
   "license": "ISC",
   "dependencies": {
     "@cardano-ogmios/client": "^4.2.1",
+    "@emurgo/cardano-serialization-lib-nodejs": "^9.1.2",
```
ngua, line 16: I've forked sundae's fork of CSL, so we should switch this once I've uploaded the NPM packages. According to @klntsky, their APIs have diverged, hopefully there won't be any issues though

src/Serialization/Address.js
```diff
@@ -0,0 +1,148 @@
...
+
+exports.toAddressCslUnsafe = addresType => {
+    return addresType.to_address();
+};
+
+exports.addressBytesImpl = baseAddr => {
+    return baseAddr.to_bytes();
+};
+
+exports.addressBech32Impl = baseAddr => {
```
klntsky, line 146: Let's stick to `_` at the end of FFI implementations, as suggested [here](https://github.com/Plutonomicon/cardano-browser-tx/pull/51#discussion_r791712120)

src/Serialization/Address.js
```diff
@@ -0,0 +1,148 @@
...
+    //
+    // reward addresses:
+    // bits 7-5: 111
+    // bit 4: credential is keyhash/scripthash
+    // bits 3-0: network id
+    //
+    // byron addresses:
+    // bits 7-4: 1000
+    const paymcheck = maybe.from(null, checks.payment);
+    if (paymcheck != null && !checkBitSet(bytes, 4, paymcheck)) {
```
klntsky, line 131: What is `checkBitSet`?

src/Serialization/Address.purs
```diff
@@ -0,0 +1,348 @@
...
+  , delegation: unit
+  }
+
+
+-- | Convert address to its byte representation
+addressBytes
+  :: forall a p d
+   . Address a p d
+  => a
+  -> Uint8Array
```
klntsky, line 235: I suggest using recently merged `ByteArray` everywhere, it has the same runtime representation as Uint8Array

test/Serialization/Address.purs
```diff
@@ -0,0 +1,44 @@
...
+
+import Data.Maybe (Maybe, maybe, fromJust)
+import Effect.Class (liftEffect)
+import Effect.Exception (error, throwException)
+import Partial.Unsafe (unsafePartial)
+import TestM (TestPlanM)
+import Serialization.Address (Bech32String(..), addressBech32, addressNetworkId, addressPubKeyHash, addressStakeKeyHash, fromBech32, newBaseAddress)
+
+
+errBool :: String -> Boolean -> TestPlanM Unit
```
klntsky, line 13: `assertTrue` would be a better name I think. Can we plug `Test.Unit.Assert.*` assertions here BTW?

ngua:
@adamczykm I just merged #58 into master, so you should be able to switch to my fork of CSL (`@ngua/cardano-serialization-lib-nodejs`) now. Hopefully it doesn't cause any issues

src/Serialization/Bech32.purs
```diff
@@ -0,0 +1,36 @@
...
+
+-- | Represents strings that are in Bech32 BIP173 format but with length limit
+-- | extended to 1023 characters.
+newtype Bech32String = Bech32String String
+
+derive instance Eq Bech32String
+derive newtype instance Show Bech32String
+
+verifyBech32 :: String -> Maybe Bech32String
+verifyBech32 = verifyBech32Impl maybeFfiHelper bech32Limit
```
klntsky, line 30: I doubt this approach is what we should follow. This wrapper only guarantees that we have a valid Bech32 data. But what we *really* are interested in is to check if a given string is a valid bech32 encoding of some *particular* data type from serialization-lib.

Here we are pulling a new dependency just to add a new option to fail the decoding which we can't meaningfully handle. Which is completely unnecessary, because `serialization-lib` contains its own bech32 implementation anyway.

adamczykm:
> @adamczykm I just merged #58 into master, so you should be able to switch to my fork of CSL (`@ngua/cardano-serialization-lib-nodejs`) now. Hopefully it doesn't cause any issues

merged in

src/Serialization/Bech32.purs
adamczykm: Even if it does, it doesn't expose it.
In case of Bech32String being just another name for a String i'd give up on the type entirely. WDYT?

src/Serialization/Address.purs
adamczykm: Okay

src/Serialization/Address.js
adamczykm: Oh I forgot about it. It will check if n'th bit is as expected

package.json
adamczykm: ok

src/Serialization/Address.js
adamczykm: ok

klntsky: Sorry, I messed up this suggestion. The original comment by @vvtran was refering to `src/Ogmious.js`

> Minor nitpick, the rest of the codebase prepends FFIs with an underscore, we should probably stick to one convention.

adamczykm: ^^

adamczykm: Reverted to the original for now.

src/Serialization/Hash.js
```diff
@@ -0,0 +1,61 @@
...
+    try {
+        ret = hashClassFrom(input);
+    }
+    catch (e) {
+        console.log(e);
+    }
+    if (ret == null) {
+        return maybe.nothing;
+    }
+    return maybe.just(ret);
```
klntsky, line 14:
```suggestion
    try {
        return maybe.just(hashClassFrom(input));
    }
    catch (e) {
        console.log(e);
        return maybe.nothing;
    }
```

I'd remove logging as well.

src/Serialization/Bech32.purs
klntsky: Yeah a type synonym is enough, as I said in slack.

test/Serialization/Address.purs
```diff
@@ -0,0 +1,45 @@
...
+
+import Prelude
+
+import Data.Maybe (Maybe, maybe, fromJust)
+import Effect.Class (liftEffect)
+import Effect.Exception (error, throwException)
+import Partial.Unsafe (unsafePartial)
+import TestM (TestPlanM)
+import Types.Aliases (Bech32String)
+import Serialization.Address (addressBech32, addressNetworkId, addressPubKeyHash, addressStakeKeyHash, fromBech32, newBaseAddress)
```
ngua, line 11: Several of these are missing from `Serialization.Address`

src/Serialization/Hash.js
```diff
@@ -0,0 +1,61 @@
+const CardanoWasm = require("@emurgo/cardano-serialization-lib-nodejs");
```
ngua, line 1: I think you'll have to change this to the forked version

src/FfiHelper.purs
```diff
@@ -0,0 +1,8 @@
+module FFiHelpers(MaybeFfiHelper, maybeFfiHelper) where
```
ngua, line 1: Is the module name a typo (`Ffi` vs `FFi`)?

src/Serialization/Address.purs
```diff
@@ -0,0 +1,348 @@
...
+-- | Parse `BaseAddress` from its byte representation
+baseAddressFromBytes
+  :: forall p d
+   . { payment :: CredType p
+     , delegation :: CredType d
+     }
+   -> ByteArray
+   -> Maybe (BaseAddress p d)
+baseAddressFromBytes checks bts = do
+  addrCsl <- baseAddressFromBytesImpl maybeFfiHelper bts
```
ngua, line 258: I think it would be nice to have an example showing how the `checks` argument should be used. For example, say I have a `ByteArray` that I want to get a `PubKeyAddress` from. What should I provide as the `checks` argument?

adamczykm: PubkeyAdress does not require checks argument, there's a dedicated function that does it for you. I will extend docs for the base function.

src/Serialization/Address.js
adamczykm: applied prefix for functions not intented for export

ngua (approved):
Looks good. We should get this merged

Issue #29, opened 2022-01-17
JSON / BigInt preprocessor


Benjmhart:
haskell serializes arbitrary-precision numbers as Integers.

when javascript decodes them as IEEE Floats, we can lose precision.

the `argonaut` library can't really be used here since it converts all json values into JS values and THEN parses them to PS values.

so we need to roll over json as a string, and whenever we encounter an integer larger than 32 bits, wrap it in quotes so it can be parsed correctly as a BigInt.

adamczykm:
So in jsons we get from outside we expect values such as: `{"bigint": 1000000000000000}`?

adamczykm:
I see here two options.
First is to provide custom JSON parser and then come up with a way to decode it to a custom types based on existing libraries or fork existing library injecting our custom parser (easy for the `simple-json` lib).
Second will be to replace all numbers in JSON for strings and then parse it as such as with existing solutions, decoding BigInts at the end of the pipeline.
Which should I pursue?

Benjmhart:
honestly replacing all numbers in json for strings and then parse it with existing solutions is what i had in mind.

Benjmhart:
> Doesn't it use doubles?
> 
> ```js
> > Number.MAX_VALUE
> 1.7976931348623157e+308
> > Number.MAX_SAFE_INTEGER
> 9007199254740991
> ```

yeah, IEEE Double-precision Floating point numbers,

PR #34, opened 2022-01-20
Parse json from string using a function turning numbers to strings.

adamczykm:
**Addresses #29** 
Parse json from string only via using a function turning numbers to strings.

- [x] Tests to be added for the helper function.


package.json
```diff
@@ -17,5 +17,10 @@
     "bufferutil": "^4.0.5",
     "uniqid": "^5.4.0",
     "ws": "^8.4.0"
+  },
+  "devDependencies": {
+    "pscid": "^2.9.3",
+    "purescript-language-server": "^0.16.3",
```
ngua, line 23: `purescript-language-server` and `purty` should already be available in the shell, provided by `spago2nix`. Was there an issue with them (we might want to try to fix that instead of getting them from NPM)?

src/Helpers.js
```diff
@@ -0,0 +1,44 @@
+
```
ngua, line 1: I'm not sure of the best way to approach this. I think it might be possible to use `JSON.parse()` with the optional second "reviver" argument. Then we could use a regex to find the original number in the JSON string and quote it. I think that would communicate the intent of the function more clearly. However I'm not sure if that would be a workable approach, so I'm very much open to discussion about it

ngua: There's also this library: https://www.npmjs.com/package/json-bigint. Usually I'd be hesitant to add another dependency, but it seems it would only add a single transitive dependency

ngua: It also has `options.storeAsString` to convert the bigint's to strings

adamczykm: Afaik revivers work on parsed values. They'd allow pretty easily to quote numbers or convert to bigint numbers that are already in IEEE754 format

src/Helpers.js
```diff
@@ -0,0 +1,44 @@
...
+    var arr = [];
+    var prev_in_number = false;
+    var in_number = false;
+    var in_string = false;
+
+    for (var i = 0, n = s.length; i < n; ++i) {
+        const c = s[i];
+        if (c == '"') {
+            if( i ==0 ||  (i>0 && s[i-1] != "\\" )){
```
klntsky, line 14: Looking back by 1 character is not enough, you need a flag:

```
> JSON.parse('{"a":"\\\\", "b": 123}')
{ a: '\\', b: 123 }
> jsonTurnNumbersToStrings('{"a":"\\\\", "b": 123}')
'{"a":"\\\\", "b": 123}'
```

klntsky (changes requested):

src/Helpers.js
klntsky: Would still be scary though!

I'm also not in favor of using that `json-bigint` library (it does not seem to be well tested either).

Our problem is:

> haskell serializes arbitrary-precision numbers as Integers.

I'm out of context w.r.t. what are we going to parse here. Still, what can be done on haskell side?

adamczykm: > Would still be scary though!
> 
> I'm also not in favor of using that `json-bigint` library (it does not seem to be well tested either).
> 
> Our problem is:
> 
> > haskell serializes arbitrary-precision numbers as Integers.
> 
> I'm out of context w.r.t. what are we going to parse here. Still, what can be done on haskell side?
I'm also not really introduced to the haskell side.
cc @Benjmhart

src/Helpers.js
ngua: Yes, it would be nice if we could use the reviver directly to modify the resulting JSON, but since the values would already be parsed, they would be JS `number`s, no? So we will have already lost precision by that point. My idea was to detect the numbers using the reviver (so we'd know we weren't in string, for example), but then modify the original JSON string

src/Helpers.js
ngua: There are various fields we need to parse as `BigInt`s when we receive query responses from ogmios (e.g. the `coins` field of the `Value` type defined in ogmios). The encoders are in `Ogmios.Data.Json.Prelude`. For example, `Coin` uses `encodeInteger` (just an alias for the normal aeson `Integer` encoder). While we could change this, we would have to encode all `Coin`s as strings, but I believe this would entail more changes to the jsonwsp schema, etc...

ngua: We want our ogmios fork to be as generally useful as possible, so I don't think we should change it there for what is really a Purescript-specific issue

src/Helpers.js
ngua: I guess we could also just use a regex without using the reviver

adamczykm: @ngua @klntsky thank you for being cautious on this.
Rory, I got your idea the second time I've read it, few hours later. I was thinking about both approaches you have mentioned before implementing low-level JS version. 

**1. Using revivers to detect numbers**
Your idea is to locate numbers -> replace them with a unique identifier -> use the identifiers locations to detect numbers locations in the string. There are some cons to this:
- efficiency
- we'd still need to parse the number manually
**2. Using regexes**
I have a feeling that the regexp correctly identifying (and matching) numbers would be at least as hard to write and read as the algorithm I have written. Do you think that it has more bugs? Maybe we could write a quickchen gen for particularly hard json strings and test it on that?

ngua: Yes, that's basically what I meant (sorry if I didn't explain it very well). Good point about the efficiency aspect, so perhaps we should discard that option. For the regex, I'm not sure how bad it would be. Maybe one of us (or both) can write one and then we can decide on that option as well

I think the quickcheck idea is great, that would be a good guarantee regardless of what implementation we end up adopting

ngua: You've spent more time than I have thinking about and solving this problem, so if you think that a low-level implementation is the most appropriate, I think that's what we should go with

src/Helpers.js
klntsky: What if we re-parsed and re-encoded the jsons on haskell side just before sending them to frontend? I mean a thin layer in front of ogmios?

@ngua

klntsky: Apparently we should be able to traverse the jsons without the need to know the datatype schema that was used to generate them. The advantage is that we would be dealing with already-parsed jsons (as a tree structure), which is more reliable than a custom state machine-based tokenizer of string.

PR #42, opened 2022-01-24
 Change some String newtype representations to Uint8Array

klntsky:
The types here are essentially byte arrays, and JS strings are utf.


adamczykm (approved):
A word of description could be useful. The changes look alright.

PR #43, opened 2022-01-24
Make Bech32 a newtype instead of String

klntsky:


adamczykm (commented):
Don't we want to provide some level of correctness by construction for these types or functions accepting these types will not require any guarantees?

ngua (commented):
I generally like making types more specific, so I am in favor of these changes. I agree with @adamczykm that it would be nice to have some guarantee that the bech32 strings are well-formed with a smart constructor, but I think we could split that into a separate issue for now

Should these have `newtype` `Eq` instances?

PR #44, opened 2022-01-24
FFI Bindings for serialization-lib

klntsky:
This allows us to serialize a minimal transaction using serialization-lib. There's some more code which I commented out because it's untested. I wrote it initially because I misunderstood the scope of this task.

Commit history is, admittedly, a mess (I'll squash it if anyone is concerned).

This code partially clashes with #31 (in particular, see `Serialization.Address`, and also `maybeFfiHelper` should be used here).


src/Serialization.purs
```diff
@@ -0,0 +1,187 @@
...
+  outputs <- convertTxOutputs body.outputs
+  fee <- convertBigInt (unwrap body.fee)
+  txBody <- newTransactionBody inputs outputs fee
+  ws <- convertWitnessSet witness_set
+  newTransaction txBody ws
+
+convertBigInt :: BigInt.BigInt -> Effect BigNum
+convertBigInt = newBigNum <<< BigInt.toString
+
+-- Parts that are commented-out are unused & untested, but should compile
```
ngua, line 100: Are you planning on completing and testing these before merging the PR?

src/Serialization.purs
```diff
@@ -0,0 +1,187 @@
...
+
+import Data.BigInt as BigInt
+import Data.Maybe (Maybe(..))
+import Data.Newtype (unwrap)
+import Data.Traversable (traverse_, for_)
+import Data.FoldableWithIndex (forWithIndex_)
+import Data.UInt (UInt)
+import Effect (Effect)
+import Prelude
+import Serialization.Types (Address, AssetName, Assets, AuxiliaryData, BaseAddress, BigNum, DataHash, Ed25519KeyHash, Ed25519Signature, MultiAsset, PlutusScript, PlutusScripts, PublicKey, ScriptHash, StakeCredential, Transaction, TransactionBody, TransactionHash, TransactionInput, TransactionInputs, TransactionOutput, TransactionOutputs, TransactionWitnessSet, Value, Vkey, Vkeywitness, Vkeywitnesses)
```
ngua, line 23: These two lines are a little long

src/Serialization.purs
```diff
@@ -0,0 +1,187 @@
...
+import Serialization.Types (Address, AssetName, Assets, AuxiliaryData, BaseAddress, BigNum, DataHash, Ed25519KeyHash, Ed25519Signature, MultiAsset, PlutusScript, PlutusScripts, PublicKey, ScriptHash, StakeCredential, Transaction, TransactionBody, TransactionHash, TransactionInput, TransactionInputs, TransactionOutput, TransactionOutputs, TransactionWitnessSet, Value, Vkey, Vkeywitness, Vkeywitnesses) as X
+import Types.ByteArray (ByteArray)
+import Types.Transaction as T
+import Untagged.Union (type (|+|))
+
+-- * cardano-serialization-lib types
+
+foreign import newBigNum :: String -> Effect BigNum
```
ngua, line 31: Do we want to follow the convention from other FFI code of using a `_` prefix for `foreign import`s? (Genuine question, I'm not sure how readable or desirable that would be in this case)

ngua (commented):
I haven't had the chance to look closely at all of the implementations, so I've left a few comments on smaller things for the time being

Do you think it would be possible and in-scope for this PR to include some conversion functions/effects in the opposite direction, e.g. `Value -> Effect T.Value`? We will need these as well soon for wallet effects

src/Serialization.purs
ngua: Oh, also are you hiding certain imports from `Serialization.Types` to avoid re-exporting them? If so, I would personally prefer to use `hiding` in that case

src/Serialization.purs
Benjmhart: i think in general this is good practice.

Benjmhart: it's a big alarm bell that 'the type is possibly wrong'

klntsky:
@ngua 

> Do you think it would be possible and in-scope for this PR to include some conversion functions/effects in the opposite direction

It is possible, but I believe we should merge this PR sooner (the goal here was to make a simple ada tx serializable). Merging will allow to move forward, while an overhead of one more PR is negligible.

src/Serialization.purs
klntsky: Not before. I accidentally wrote more code than was needed. This may save someone some time once we need it. There is some commented-out code in `Transaction.purs`, so I thought that the practice of leaving dead code commented-out is OK.

src/Serialization.purs
klntsky: Good catch, I probably just added a new type and forgot to re-import it here. I think we can drop this re-export.

src/Serialization.purs
klntsky: @Benjmhart but defining a new function without underscore that just reuses the FFI one does not add more guarantees on that front.

I thought that the convention is to use `_` for functions that accept extra arguments or are "implementations" in some other sense.

ngua (approved):

PR #51, opened 2022-01-25
Add ByteArray type with Ord/Show/Eq (replacement for Uint8Array)

klntsky:
Since `TokenName` and `CurrencySymbol` are bytes, and we need to store them as keys in a map, we need Ord instance for our byte array type. Hence this wrapper.


adamczykm:
Lgtm

src/Types/ByteArray.js
```diff
@@ -0,0 +1,64 @@
+// From instance for Array in prelude
```
adamczykm, line 1: This should prefix all functions that are unchanged reuse of Prelude functions. For others we do need tests.

src/Types/ByteArray.js
```diff
@@ -0,0 +1,64 @@
...
+exports.ordImpl = function (f) {
```
vvtran, line 2: Minor nitpick, the rest of the codebase prepends FFIs with an underscore, we should probably stick to one convention.

vvtran (commented):
Looks good.

src/Types/ByteArray.purs
```diff
@@ -0,0 +1,59 @@
...
+-- | Input string must consist of hexademical numbers.
+-- | Length of the input string must be even (2 characters per byte).
+hexToByteArray :: String -> Maybe ByteArray
+hexToByteArray = hexToByteArrayImpl Nothing Just
+
+-- | Characters not in range will be converted to zero.
+foreign import hexToByteArrayUnsafe :: String -> ByteArray
+
+-- | Overflowing integers will be silently accepted modulo 256.
+foreign import byteArrayFromIntArray :: Array Int -> ByteArray
```
adamczykm, line 57: Let's maybe add Unsafe suffix here as well

adamczykm: Having safe API available might also be convenient.
I'd rather have only safe (guaranteed to provide valid output for all of their possible inputs) functions than only unsafe ones.

adamczykm (changes requested):

vvtran:
Noticed something else, in `Transaction` but we should try to make our imports explicit (same for exports) so `import Types.ByteArray (LIST THEM)` instead of `import Types.ByteArray`. This also holds for data constructors (unless there's too many), so try to avoid `(..)` e.g. `Maybe(..)`, you can read more about the Haskell (not PS but presumably it holds here) style guide here https://github.com/mlabs-haskell/styleguide Admittedly we don't have any set in stone conventions of purescript but I assume they'll be similar.

src/Types/ByteArray.purs
ngua: I don't know if I'd qualify this as unsafe in the same way `hexToByteArrayUnsafe` is -- that one can throw an exception from use of `parseInt` (it isn't caught anywhere), where this one will just convert overflowing ints

src/Types/ByteArray.js
```diff
@@ -0,0 +1,64 @@
...
+    return function (xs) {
+        return function (ys) {
```
ngua, line 4: A few stylistic nitpicks (if we feel these don't matter please feel free to disregard :)): 

- In the rest of the codebase, we use arrow functions for currying with FFI stuff 
- Sometimes `const` is used (and `let` where appropriate) in this module, but there's some `var`s here and below

src/Types/ByteArray.purs
adamczykm: Maybe we could add 'modulo' somewhere in its name given its type it'd make it clear that there is some modulo conversion of its input?

ngua (approved):
LGTM

src/Types/ByteArray.purs
klntsky: Silent data corruption is more "unsafe" than an exception. I doubt anyone will use it incorrectly, ever, though.

Note that main use of this function is to be in tests for fixtures and to make `Show` instance that gives valid code.

So I added 'unsafe' suffix and made a new "safe" function with overflow checks.

src/Types/Transaction.purs
```diff
@@ -1,22 +1,24 @@
 module Types.Transaction where
 
 import Prelude
-import Data.ArrayBuffer.Types(Uint8Array)
 import Data.BigInt as BigInt
+import Data.Maybe (Maybe)
```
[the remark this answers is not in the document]
klntsky: Oops.

vvtran (approved):
Looks good!

adamczykm (approved):
Lgtm

Issue #28, opened 2022-01-16
add `purty` autoformat script


Benjmhart:

paluh:
Maybe it would be better to just use `purs-tidy` from the beginning - we can say that it is a de facto standard formatting tool for PS right now ( `purty` has not been updated to the latest PS version. ). Here is an example config from [purescript-contrib/purescript-parsing](https://github.com/purescript-contrib/purescript-parsing/blob/main/.tidyrc.json).

adamczykm:
Cześć @paluh

ngua:
This is taken care of by #52

PR #52, opened 2022-01-25
Add purs-tidy code formatter.

adamczykm:
Addresses #28


ngua:
Very nice. If you can fix the merge conflicts when you have the change we should merge this. Maybe make another issue for checking formatting on CI (I think @Benjmhart will have to configure some things)?

ngua:
Oh and perhaps you could expand the `format` makefile target to include all PS files, including in `test`

adamczykm:
Ok will do the above within couple of hours

Benjmhart (approved):

PR #54, opened 2022-01-26
Viet/ogmios to transaction

vvtran:
- Write helpers to go from Ogmios to Transaction types to run `utxosAt`
- Small bug fix in `OgmiosTxOut` parsing, when looking at `"datum"`.
- Write `utxosAt'` to get utxos in for Transaction types.
- Add error handling for this function call in the balancer.
- Add explicit exports/imports for `Ogmios` and `Types.JsonWsp` since I am editing them.
- Change Ogmios `TxOutRef` to be indexed by unsigned integers.


vvtran:
Just made a small refinement on `convertUtxos`, keeping it in `Effect` until `getUtxos` where it's converted to `Aff`

src/Types/JsonWsp.purs
```diff
@@ -141,7 +158,7 @@ type UtxoQueryResult = Map.Map TxOutRef OgmiosTxOut
 -- TxOutRef
 type TxOutRef =
   { txId :: String
-  , index :: BigInt.BigInt
+  , index :: UInt.UInt
```
klntsky, line 161: TxOutRef index is long integer in https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger/html/Ledger.html#t:TxOutRef

klntsky: In serialization-lib it's just js `Number` though.

src/Ogmios.purs
```diff
@@ -238,3 +261,70 @@ messageFoldF
...
+addressToOgmiosAddress :: Address -> Effect (Maybe JsonWsp.Address)
+addressToOgmiosAddress addr = do
+  addr' <- Serialization.convertAddress addr
+  pure $ newBaseAddressFromAddress addr' >>= addressPubKeyHash <#> unwrap
+
+-- If required, we can change to Either with more granular error handling.
+-- | Gets utxos at an (internal) Address in terms of (internal) Transaction.Types
+utxosAt' :: Address -> QueryM (Maybe UtxoM)
+utxosAt' addr = do
+  addr' :: Maybe JsonWsp.Address <-
```
ngua, line 284: This is a stylistic thing and my personal preference, and I am of course open to debating it: I understand type signatures on `where` clauses and why our Haskell style guide mandates them (`let` bindings less so, but it can sometimes make things clearer and I wouldn't recommend removing them if they're already there). I would prefer, however, to only use type signatures on local bindings if the types cannot be inferred (this also applies to lines 284, 293, 308, in this module), mostly for the sake of readability

ngua (changes requested):
I think it would be good to have add some tests for the conversion. Alternately, it might be better to just update the current `AffInterface` tests with the new `utxosAt'` (although I would really like to switch the names of `utxosAt` and `utxosAt'`)

src/Ogmios.purs
```diff
@@ -238,3 +261,70 @@ messageFoldF
...
+
+-- | Converts an (internal) Address to JsonWsp.Address
+addressToOgmiosAddress :: Address -> Effect (Maybe JsonWsp.Address)
+addressToOgmiosAddress addr = do
+  addr' <- Serialization.convertAddress addr
+  pure $ newBaseAddressFromAddress addr' >>= addressPubKeyHash <#> unwrap
+
+-- If required, we can change to Either with more granular error handling.
+-- | Gets utxos at an (internal) Address in terms of (internal) Transaction.Types
+utxosAt' :: Address -> QueryM (Maybe UtxoM)
```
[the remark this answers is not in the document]
vvtran: I did consider this and agree.

src/Ogmios.purs
vvtran: So I tend to do it to be explicit - it assists with readability and shouldn't make it worse, although it makes things less terse. I think 284, 308 (and 322 for that matter) are fair, as they're easy to infer. Although 293 is a bit more complex and its type is somewhat useful IMO. I would even say the `let` type expression below is helpful too. Potentially one for the style guide, but I would argue to leave it to the reader's judgement, depending on the binding's complexity?

I've removed 284, 308 and 322 but left the others for now.

src/Ogmios.purs
```diff
@@ -238,3 +261,70 @@ messageFoldF
...
+-- | Converts an Ogmios TxOutRef to (internal) TransactionInput
+txOutRefToTransactionInput :: TxOutRef -> Maybe TransactionInput
+txOutRefToTransactionInput { txId, index } = do
+  transaction_id :: TransactionHash <- hexToByteArray txId <#> wrap
+  pure $ wrap
+    { transaction_id
+    , index
+    }
+
+-- https://ogmios.dev/ogmios.wsp.json see "datum", potential FIX ME: it says
```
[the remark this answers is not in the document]
vvtran: https://github.com/Plutonomicon/cardano-browser-tx/issues/78

src/Ogmios.purs
ngua: Fair enough, this is a matter of personal preference. I would add though that the effect on readability is subjective as well, as I find it makes things less readable

vvtran:
@ngua  just removed a redundant import and a `do`.

vvtran:
I think the recent PR into `master` will break my tests due to a refactor so I will try to fix this now.

vvtran:
Should be okay now.

ngua (approved):
Looking good now, let's merge it

Issue #38, opened 2022-01-20
Set up infra for Haskell server


ngua:
Set up the infrastructure for the Haskell server (`applyCode`, fee estimation, etc...) in this repository

PR #55, opened 2022-01-27
Fee estimation (Haskell server)

### LINKED ISSUE #36: Fee Estimation function
we need a function to estimate fees, this will be easiest to use a haskell server for now

ngua:
Closes #36 

This adds an endpoint `/fees` to the Haskell server I started with #38. The endpoint accepts an obligatory query parameter `tx`, which should be a CBOR-encoded `Tx AlonzoEra`. It then uses `evaluateTransactionFee` to calculate a fee estimation. The fee is returned as a JSON string in order to prevent the same issue that led to a need for #29 

TODO: verify the obtained fee against another tool (`cardano-cli`?)


vvtran:
Hey, this looks good on the whole, just a few comments (I haven't looked through Servant yet but will soon).
- Just a preference but I would avoid qualifying imports as the same name personally to avoid confusion (I forget if this is part of the style guide or not)
```
import Cardano.Api qualified as C
import Cardano.Api.Shelley qualified as C
```
There's a few more of these lurking around (`Aeson` related)
- Related, more explicit imports/exports would be better although probably something to do at the end. Also related `Env {..} <- ask`, it's obvious what's going on ofc.
- Have you mixed up the order of `evaluateTransactionFee` for the witnesses? I think Shelley comes first? https://input-output-hk.github.io/cardano-node/cardano-api/src/Cardano.Api.Fees.html#evaluateTransactionFee

vvtran:
Another comment regards to the style guide, there's a few parametrically polymorphic parameters lurking around that will need Kind signatures and forall. quantification. 

A minor nitpick, I don't know if this is a Servant convention but is the type application necessary?
```
api :: Proxy Api
api = Proxy @Api
```
If you prefer this then it's fine of course :)

ngua:
@vvtran 
- Yikes! I did mess up the order of the params in `evaluateTransactionFee`, good catch! 
- You're right, the visible type application is not necessary there, `Proxy` is fine since there's a top-level type signature
- The style guide does recommend avoiding importing several modules under a single qualified name, although IIRC this is one of the points where there's latitude given. I'll change that regardless
Thanks for the feedback!

server/src/Utils.hs
```diff
@@ -0,0 +1,15 @@
...
+import Data.Text (Text)
+import Data.Text qualified as Text
+
+bshow :: Show a => a -> C8.ByteString
+bshow = C8.pack . show
+
+lbshow :: Show a => a -> LC8.ByteString
+lbshow = LC8.fromStrict . bshow
+
+tshow :: Show a => a -> Text
```
vvtran, line 14:
```
bshow :: forall (a :: Type). Show a => a -> C8.ByteString
bshow = C8.pack . show

lbshow :: forall (a :: Type). Show a => a -> LC8.ByteString
lbshow = LC8.fromStrict . bshow

tshow :: forall (a :: Type). Show a => a -> Text
tshow = Text.pack . show
``` 
Do we want to quantify these too?

server/src/Types.hs
```diff
@@ -0,0 +1,86 @@
...
+
+import Cardano.Api.Shelley qualified as Shelley
+import Cardano.Binary qualified as Cbor
+import Control.Exception (Exception)
+import Control.Monad.Catch (MonadThrow)
+import Control.Monad.IO.Class (MonadIO)
+import Control.Monad.Reader (MonadReader, ReaderT)
+import Data.Aeson (FromJSON, ToJSON (..))
+import Data.Aeson qualified as Aeson
+import Data.Aeson.Encoding qualified as Aeson
```
vvtran, line 19: Entirely optional but there's a "duplicate" import here too.

vvtran (commented):
Few comments, otherwise, LGTM.

server/src/Utils.hs
ngua: Ah, yes, good catch

klntsky (commented):
Works for me. A small README in `server/` would be totally great.

ngua:
@klntsky Good idea, I'll add one

server/src/Types.hs
ngua: Will change this too

Benjmhart:
on the topic of docs, would love to see the server added to the makefile. LGTM otherwise

ngua:
@Benjmhart Good idea, I'll add a target to the Makefile

server/README.md
```diff
@@ -0,0 +1,44 @@
+# cardano-browser-tx-server
+
+This is a small Haskell project to provide services to the `cardano-browser-tx` frontend that cannot be achieved using Purescript
```
adamczykm, line 3: Why can't they be achieved using Purescript? Is it actually the goal of this project to provide only such services?

adamczykm (commented):
Looks nice!

adamczykm (approved):

server/README.md
ngua: There are some things that we can't do very easily with PS. The fee estimation we could probably do by replicating `evaluateTransactionFee` and including the necessary information from Alonzo `ProtocolParameters`. There are others that we can't do easily though, like Plutus' `applyCode`

Issue #64, opened 2022-02-01
Implement core of Nami wallet interface


ngua:
While we don't need to implement all of Nami's functionality, we do need FFI wrappers around at least the following to build an `UnbalancedTx` as part of our first goal:

**NOTE**: Some of the links below are for the deprecated `window.cardano` injected API. This is because there are no docs for the new injected API; the parameters and return values are the same for the examples below, however, so I linked to them for convenience

- [x] `enable` (https://github.com/Berry-Pool/nami-wallet#cardanoenable) -- **WORKING** 
- [x] `getUsedAddresses` (https://github.com/Berry-Pool/nami-wallet#cardanogetusedaddresses; note that Nami only has a single address per wallet but returns an array to follow the CIP) -- **WORKING**
- [x] `getCollateral` (https://github.com/Berry-Pool/nami-wallet#apiexperimentalgetcollateral) -- **WORKING** 
- [x] `signTx`. Nami actually returns a `TransactionWitnessSet`, but it probably makes more sense to have the action as follows for ease of use
  ```purescript
  signTx :: Transaction -> Aff Transaction
  ``` 
  In this case `signTx` will add the new, returned witness set to the existing transaction

All of these return an `Effect (Promise ...)` which we can then lift into `Aff` (i.e. `Control.Promise.toAffE`. Most of them return a hexadecimal string which needs to be decoded

Related issues/PRs: #60, #31, #44, #49

klntsky:
So for `getCollateral` we need to work with `TransactionUnspentOutput` (i.e. being able to de-serialize it). Should I create a task to add this type?

klntsky:
As for de-serialization, we only need `BaseAddress` and `TransactionUnspentOutput` for now, right?

ngua:
Yes, that's correct. I can create an issue, although I think we can just add the type and the deserialization in a single PR

PR #66, opened 2022-02-01
Address deserialization

klntsky:
This is needed for `getUsedAddresses` (see #64).

### DIFF
--- a/src/Deserialization.purs
+++ b/src/Deserialization.purs
@@ -1,0 +1,65 @@
+module Deserialization
+  ( convertAddress
+  ) where
+
+import Control.Alt ((<|>))
+import Data.Maybe (Maybe(..))
+import Prelude
+import Serialization.Types (Address, BaseAddress, Ed25519KeyHash, ScriptHash, StakeCredential)
+import Types.ByteArray (ByteArray)
+import Types.Transaction as T
+import Untagged.Union (type (|+|), asOneOf)
+import Data.UInt as UInt
+import Serialization (toBytes)
+
+foreign import _baseAddressFromAddress :: (forall a. Maybe a) -> (forall a. a -> Maybe a) -> Address -> Maybe BaseAddress
+foreign import addressNetworkId :: Address -> Int
+foreign import baseAddressStakeCredential :: BaseAddress -> StakeCredential
+foreign import baseAddressPaymentCredential :: BaseAddress -> StakeCredential
+foreign import _stakeCredentialToScriptHash :: (forall a. Maybe a) -> (forall a. a -> Maybe a) -> StakeCredential -> Maybe ScriptHash
+foreign import _stakeCredentialToKeyHash :: (forall a. Maybe a) -> (forall a. a -> Maybe a) -> StakeCredential -> Maybe Ed25519KeyHash
+
+stakeCredentialToKeyHash :: StakeCredential -> Maybe Ed25519KeyHash
+stakeCredentialToKeyHash = _stakeCredentialToKeyHash Nothing Just
+
+stakeCredentialToScriptHash :: StakeCredential -> Maybe ScriptHash
+stakeCredentialToScriptHash = _stakeCredentialToScriptHash Nothing Just
+
+baseAddressFromAddress :: Address -> Maybe BaseAddress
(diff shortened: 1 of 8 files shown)


src/Deserialization.purs
```diff
@@ -0,0 +1,65 @@
+module Deserialization
+  ( convertAddress
+  ) where
+
+import Control.Alt ((<|>))
+import Data.Maybe (Maybe(..))
+import Prelude
+import Serialization.Types (Address, BaseAddress, Ed25519KeyHash, ScriptHash, StakeCredential)
+import Types.ByteArray (ByteArray)
```
vvtran, line 9: Redundant import?

src/Serialization.purs
```diff
@@ -163,11 +171,7 @@ convertTxOutput (T.TransactionOutput { address, amount, data_hash }) = do
     T.StakeCredentialScript (T.ScriptHash scriptHash) -> do
```
vvtran, line 171: Not really deserialisation but minor stylistic nitpick - the `do` is unnecessary for these oneliner binds. Entirely optional to remove of course :)

src/Serialization.purs
```diff
@@ -45,7 +44,6 @@ foreign import newBaseAddress :: UInt -> StakeCredential -> StakeCredential -> E
 foreign import _newBaseAddressFromAddress :: (forall a. Maybe a) -> (forall a. a -> Maybe a) -> Address -> Maybe BaseAddress
 foreign import baseAddressPaymentCredential :: BaseAddress -> StakeCredential
 foreign import baseAddressToAddress :: BaseAddress -> Effect Address
-foreign import baseAddressFromAddress :: Address -> Effect BaseAddress
```
vvtran, line 48: Is it worth removing the js counterpart if we're not using anymore?

vvtran (commented):
Looks good with a few optional comments. I still need to get around to learning JS properly so didn't get into too much detail here.

test/Deserialization.purs
```diff
@@ -0,0 +1,35 @@
...
+import Types.Transaction as T
+import Effect.Exception (throw)
+
+suite :: TestPlanM Unit
+suite = do
+  group "cardano-serialization-lib bindings (deserialization)" $ do
+    group "Address" $ do
+      test "deserialization works" do
+        let
+          addressString = "addr1qyc0kwu98x23ufhsxjgs5k3h7gktn8v5682qna5amwh2juguztcrc8hjay66es67ctn0jmr9plfmlw37je2s2px4xdssgvxerq"
```
ngua, line 21: Maybe we could have this as a constant somewhere in this module, since the two `addressString`s are identical

ngua (approved):
Very nice work. I merged your branch into the Nami integration one and it works perfectly for `getWalletAddress`

@vvtran and I have suggested some small changes if you'd like to consider them. But I wanted to approve this so we can merge it as soon as possible

adamczykm (approved):
Lots of those will be strongly affected when #31 is merged.

src/Serialization.purs
klntsky: I prefer to leave `do` in case a new statement gets added in the future - then the diff is shorter. Plus, it's an easy to spot indicator that the result is monadic.

src/Deserialization.purs
```diff
@@ -0,0 +1,65 @@
+module Deserialization
+  ( convertAddress
+  ) where
+
+import Control.Alt ((<|>))
+import Data.Maybe (Maybe(..))
```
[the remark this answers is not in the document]
klntsky: Why for `Maybe`?

Also, psc-ide-mode automatically fixes implicit imports by inserting a list of imported things with `(..)` for constructors, so fixing these would be a lot of manual labor for me ^^'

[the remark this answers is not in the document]
vvtran: Mainly because it's mandated by the **Haskell** style guide: https://github.com/mlabs-haskell/styleguide
I don't really mind if you ignore it as someone will likely adhere to the guide at the end :D I tend to do it when I can so it saves on work later on.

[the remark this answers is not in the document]
klntsky: No I think that's important. Just forgot it's in the guide.

vvtran (approved):

PR #68, opened 2022-02-02
Add parcel setup

### LINKED ISSUE #62: Pick a JS bundler
We need to decide on which bundler we want to use, then
- add it to the shell to use it manually when needed
- extend the derivations (e.g. `bundlePursProject`) in `nix/lib.nix`

We've discussed either using Parcel or Webpack

ngua:
Closes #62 

Adds parcel to the project as a `devDependency`. There are two more scripts now:

- `dev`: runs `spago build --watch` and `parcel` for hot reloading
- `build`: runs `spago build && parcel` to build and bundle a project

I'm intentionally using `parcel` without an argument; if you pass `npm run {dev|build} <some-path>` you can point it to a specific `index.html` to use. This is kind of a hack but quite useful to me to test the Nami integration. Parcel's default is `./index.html`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ result-*
 /server/dist-newstyle
 /server/dist-newstyle/
 .projectile
+/**/index.js
+dist/
+.parcel-cache/
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,7 @@ SHELL := bash
 
 autogen-deps:
 	spago2nix generate \
-		&& node2nix -l package-lock.json \
-		&& mv default.nix node2nix.nix \
-		&& git restore default.nix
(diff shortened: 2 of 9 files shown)


nixinator (changes requested):
That's certainly complicated flake.  I've not seen pattern like that a flake , imports even more nix from /nix/. directory, via the flake.nix in the root.

I'm not sure that helps readability, or the level of indirection is needed, but perhaps it is. 

interesting indeed.

ngua:
Unfortunately I wasn't aware that parcel doesn't support loading WASM modules by default. I've gotten it working with webpack, though, so I'm going to close this and open another PR with a webpack setup

ngua:
Opened #70 instead

Issue #69, opened 2022-02-03
Load different CSL packages depending on environment


ngua:
We need to figure out a way to either load the Node or browser version of CSL

klntsky:
Looks like this thing is what we need: https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin

If we want to build with nix, we can pass info to webpack config through environment variables.

ngua:
Closed by #80

ngua:
Closed by https://github.com/Plutonomicon/cardano-browser-tx/pull/80

PR #80, opened 2022-02-07
Add conditional loading (browser/node versions of serialization-lib)

klntsky:
Partially addresses #69.

Nix builds are still TODO.

Note that other PRs should change `serialization-lib` imports when this is merged (by copy-pasting import code from here)


MatthewCroughan (approved):

nixinator:
Having to to commit the nodepackages.nix , it's bit and ugly, but no way around it at this time.

ngua (changes requested):
I'm having a build error from `purs-loader` when trying to build with webpack. I'll take a look tomorrow to figure out what the issue could be

ngua (changes requested):
I got everything working despite webpack's immensely confusing error messages:

- It looks like you haven't merged with your newest deserialization branch? At any rate, I merged that PR already so if you merge with master it should be OK
- Of course we need to adjust the CSL import in the `Deserialization.UnspentOutput` FFI module
- We also need to add `convertInput` to `Deserialization.UnspentOutput`'s exports
- And the comment below about `npm run dev`

klntsky:
I made some more changes to ensure websocket wrapper works in the browser.

@ngua what was the problem with purs-loader?

ngua:
@klntsky The `purs-loader` thing was a misleading error from webpack. The issue was just that the PS wasn't compiling. I didn't have the time that evening to investigate further and found the real issue the next day

ngua (approved):

PR #84, opened 2022-02-08
Viet/value smart constructors

vvtran:
https://github.com/Plutonomicon/cardano-browser-tx/issues/75


src/BalanceTx.purs
```diff
@@ -227,25 +218,28 @@ newtype Expected = Expected Value
 derive instance newtypeExpected :: Newtype Expected _
 
 instance showExpected :: Show Expected where
-  show = show <<< flattenValue <<< unwrap
+  show = show <<< unwrap
```
klntsky, line 221: Let's make Show instances more copy-paste-able. I.e. not omit constructor name entirely. Also, it can be expressed as `derive newtype instance Show Expected`. But in this case I'd go with genericShow.

src/Deserialization/UnspentOutput.purs
```diff
@@ -54,24 +55,27 @@ convertValue :: Value -> Maybe T.Value
 convertValue value = do
   coin <- BigInt.fromString $ bigNumToString $ getCoin value
   -- multiasset is optional
-  multiasset <- for (getMultiAsset maybeFfiHelper value) \multiasset -> do
+  multiasset''' <- for (getMultiAsset maybeFfiHelper value) \multiasset -> do
```
klntsky, line 58: I think we don't need triple `'` here?

src/BalanceTx.purs
ngua: I would also suggest `genericShow` here (and in general unless there is a very compelling reason not to use it)

src/Types/Value.purs
```diff
@@ -269,44 +386,32 @@ flattenValue (Value coin@(Coin lovelaces) nonAdaAsset) =
...
-      Value mempty
-        <<< wrap
-        <<< Map.singleton curSymbol
-        <<<
-          Map.singleton tokenName $ amount
-    true -> Value (wrap amount) mempty
+  mkSingletonValue' curSymbol tokenName amount
 
 -- | Predicate on whether some Value contains Ada only.
 isAdaOnly :: Value -> Boolean
```
ngua, line 398: I noticed this is exported as well, but I think this might be confusing for users. I would expect it more to check that the `Value` contains a non-zero `Coin` amount _and_ the `NonAdaAsset` is equivalent to `mempty`

If this function ever returned `true` in user code, it would mean that we failed to uphold the invariants of our `Value` (that is, ensuring it's impossible to construct a spurious "ada" from a `TokenName` and `CurrencySymbol`)

ngua: I'm fine with leaving the current implementation for internal use if necessary, but I think it would be best to rename it to `isUnsafeAdaOnly` or something to clearly communicate the purpose (and not to export it)

src/Types/Value.purs
```diff
@@ -373,56 +478,63 @@ eq :: Value -> Value -> Boolean
 -- If both are zero then checkBinRel will be vacuously true, but this is fine.
 eq = checkBinRel (==)
 
+isAda :: CurrencySymbol -> TokenName -> Boolean
```
ngua, line 481: As above, I think it would be clearer to add `unsafe` somewhere in here. I mean that we shouldn't refer to something that should never exist as "ada" (and I like the `unsafeAda...` convention you use elsewhere)

ngua (changes requested):
Things generally look good, I have a few small suggestions

src/Deserialization/UnspentOutput.purs
vvtran: @klntsky you're correct.

src/Types/Value.purs
vvtran: I *think* this is fine to export. So it's not quite equivalent to ```Value contains a non-zero Coin amount and the NonAdaAsset is equivalent to mempty``` because it calls `FlattenValue`/`flattenNonAdaValue` first which guards against zero values. In particular, you could have a non-Ada Value with `zero` `BigInt` which isn't `mempty` (because it's not an empty `Map`) but `FlattenValue` will discard such a value for verifying. In other words, `FlattenValue` gets rid of degenerate cases, whether we need this in practice I don't know (maybe the deserialisation deals with this)? It's just following the Plutus logic.

I would say `flattenValue` is unsafe to export because it returns the `unsafeAdaSymbol` etc. in the list - note that it places the unsafe Ada stuff at the *head* of the list and `isAdaOnly` checks for a singleton list.

Whilst these functions don't necessarily protect us from a bad deserialisation due to exploits from an external actor (i.e. if someone managed to send in AdaSymbol/Token into `NonAdaAsset` and we processed badly). It returning `true` should be safe assuming we are constructing `Value` *internally* using our helpers? Because it calls `FlattenValue` to create the unsafe flattened list, and just checks this (it doesn't return the list). We just need a case of degenerate non-Ada values in `NonAdaAsset` (not necessarily an empty map but all zeros) and `Coin` to be non-zero, and it will return `true`.

vvtran: Let me know your thoughts, I could change `flattenValue` to `unsafeFlattenValue` to discourage (provided you are happy with the above ofc).

vvtran: Another of saying this is, we are using validating, not parsing and using Boolean blindness, so it *should* be safe.

src/Types/Value.purs
vvtran: This one I would agree with because we shouldn't be able to construct the adaSymbol etc.

PR #87, opened 2022-02-09
Viet/balancer placeholder integration

vvtran:
https://github.com/Plutonomicon/cardano-browser-tx/issues/86 and https://github.com/Plutonomicon/cardano-browser-tx/issues/71

### DIFF
--- a/src/BalanceTx.purs
+++ b/src/BalanceTx.purs
@@ -264,48 +264,59 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) = do
             -- Combines utxos at the user address and those from any scripts
             -- involved with the contract in the unbalanced transaction. Don't
             -- add collateral to this for now.
-            utxoIndex''' :: Utxo
-            utxoIndex''' = utxos `Map.union` utxoIndex''
+            allUtxos :: Utxo
+            allUtxos = utxos `Map.union` utxoIndex''
           -- Add hardcoded Nami 5 Ada and sign before instead of recursively
-          -- signing in loop.
+          -- signing in loop. Could refactor this part into a separate function.
           signedUnbalancedTx <-
             signTransaction (addTxCollateral unbalancedTx collateral)
               <#> note (SignTxError' $ CouldNotSignTx ownAddr)
           case signedUnbalancedTx of
             Left err -> pure $ Left err
-            Right (Transaction ubTx@{ body }) -> do
+            Right sUbTx ->
               -- After adding collateral, we need to balance the inputs and
               -- non-Ada outputs before looping, i.e. we need to add input fees
               -- for the Ada only collateral. No MinUtxos required:
               -- Alternatively add this functionality back to preBalanceTxBody
               -- and only call collateral adder initially.
-              signedUnbalancedTx' <- pure do
-                -- Balance tx without fees:
-                body' <- balanceTxIns utxoIndex''' zero body
-                body'' <- balanceNonAdaOuts ownAddr utxoIndex''' body'
-                pure $ wrap ubTx { body = body'' }
-              case signedUnbalancedTx' of
+              -- Balance tx without fees:
+              case prebalanceCollateral zero allUtxos ownAddr sUbTx of
                 Left err -> pure $ Left err
-                Right signedUnbalancedTx''@(Transaction ubTx'@{ body: body' }) -> do
-                  fees' <-
-                    lmap CalculateMinFeeError' <$>
-                      calculateMinFee' signedUnbalancedTx''
-                  signedCollUnbalancedTx' <- pure do
-                    -- Balance tx with fees:
-                    fees <- fees'
-                    body'' <- balanceTxIns utxoIndex''' (fees + feeBuffer) body'
-                    body''' <- balanceNonAdaOuts ownAddr utxoIndex''' body''
-                    pure $ wrap ubTx' { body = body''' }
+                Right sUbTx' -> do
+                  fees' <- lmap CalculateMinFeeError' <$> calculateMinFee' sUbTx'
+                  -- Balance tx with fees:
+                  let
+                    signedCollUnbalancedTx' = fees' >>= \fees ->
+                      prebalanceCollateral
+                        (fees + feeBuffer)
+                        allUtxos
+                        ownAddr
+                        sUbTx'
                   case signedCollUnbalancedTx' of
                     Left err -> pure $ Left err
                     Right signedCollUnbalancedTx ->
-                      loop utxoIndex''' ownAddr [] signedCollUnbalancedTx >>=
+                      loop allUtxos ownAddr [] signedCollUnbalancedTx >>=
                         either
                           (Left >>> pure)
-                          ( returnAdaChange ownAddr utxoIndex'''
+                          ( returnAdaChange ownAddr allUtxos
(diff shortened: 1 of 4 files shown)


vvtran:
Please ignore @nixinator @MatthewCroughan. Created from a branch of @ngua and it automatically requested review.

src/BalanceTx.purs
```diff
@@ -109,22 +103,46 @@ import Undefined (undefined)
 -- These derivations may need tweaking when testing to make sure they are easy
 -- to read, especially with generic show vs newtype show derivations.
 data BalanceTxFailure
```
ngua, line 105: I think the number of type/data constructors for different error conditions might be getting a bit out of hand. Some of these can just have `String`s giving more context instead of two additional levels of data constructors (e.g. most of the `*Failure` -> `*Reason` newtypes)

src/BalanceTx.purs
```diff
@@ -109,22 +103,46 @@ import Undefined (undefined)
...
 
+newtype GetWalletAddressFailure = GetWalletAddressFailure GetWalletAddressFailureReason
+
+derive instance newtypeGetWalletAddressFailure :: Newtype GetWalletAddressFailure _
+derive newtype instance showGetWalletAddressFailure :: Show GetWalletAddressFailure
+
+data GetWalletAddressFailureReason = CouldNotGetNamiWalletAddress
+
+instance showGetWalletAddressFailureReason :: Show GetWalletAddressFailureReason where
+  show CouldNotGetNamiWalletAddress = "Couldn't get Nami wallet address."
```
ngua, line 134: I think it's best to treat `Show` as something that prints PS expressions that can be evaluated and not arbitrarily descriptive `String`s. If you'd like to generate descriptions of errors, I would suggest something like `Affjax.printError`

src/BalanceTx.purs
```diff
@@ -356,18 +385,18 @@ balanceTxM ownAddr (UnbalancedTransaction { unbalancedTx, utxoIndex }) = do
...
+        utxoIndex'
+        ownAddr'
+        tx''
 
     case balancedTxBody' of
       Left err -> pure $ Left err
       Right balancedTxBody'' ->
-        if txBody' == balancedTxBody'' then pure $ Right $ wrap tx' { body = balancedTxBody'' }
+        if txBody' == balancedTxBody'' then
+          pure $ pure $ wrap tx' { body = balancedTxBody'' }
```
ngua, line 399: I might suggest using `Right` explicitly here

src/BalanceTx.purs
```diff
@@ -270,72 +301,70 @@ data BuildTxRawFailureReason = CannotBuildTxRaw
...
+  collateral' <- getWalletCollateral
+  -- Saves on an extra nested case:
+  case ownAddr', collateral' of
+    Nothing, _ ->
+      pure $ Left $ GetWalletAddressFailure' $ wrap CouldNotGetNamiWalletAddress
+    _, Nothing ->
+      pure $ Left $ GetWalletCollateralFailure' $ wrap CouldNotGetNamiCollateral
+    Just ownAddr, Just collateral -> do
+      utxos' <- map unwrap <$> utxosAt ownAddr
+      utxoIndex''' <- liftEffect $ utxoIndexToUtxo utxoIndex
```
ngua, line 340: If you're going to use several primes in a row, I might like to suggest using them in ascending order. Here we have `utxoIndex` to `utxoIndex'''`, to `utxoIndex''`

src/BalanceTx.purs
```diff
@@ -412,120 +454,126 @@ returnAdaChange
...
+                -- then adjust as necessary if we have sufficient Ada in the input.
+                let
+                  utxoCost :: BigInt
+                  utxoCost = getLovelace protocolParamUTxOCostPerWord
+
+                  -- An ada-only UTxO entry is 29 words. More details about min utxo
+                  -- calculation can be found here:
+                  -- https://github.com/cardano-foundation/CIPs/tree/master/CIP-0028#rationale-for-parameter-choices
+                  changeMinUtxo :: BigInt
+                  changeMinUtxo = (fromInt 29) * utxoCost
```
ngua, line 529: Perhaps you could make a constant instead of using a `29` literal here?

src/BalanceTx.purs
```diff
@@ -109,22 +103,46 @@ import Undefined (undefined)
...
 
 instance showBalanceTxFailure :: Show BalanceTxFailure where
   show = genericShow
 
+newtype GetWalletAddressFailure = GetWalletAddressFailure GetWalletAddressFailureReason
+
+derive instance newtypeGetWalletAddressFailure :: Newtype GetWalletAddressFailure _
+derive newtype instance showGetWalletAddressFailure :: Show GetWalletAddressFailure
+
+data GetWalletAddressFailureReason = CouldNotGetNamiWalletAddress
```
ngua, line 131: Just to check -- if the intention here to add more constructors over time for different wallets? (same for `GetWalletCollateralFailure` below

src/BalanceTx.purs
vvtran: Will look to simplify errors.

src/BalanceTx.purs
vvtran: Yes, and potentially any other unforeseen error types.

src/BalanceTx.purs
```diff
@@ -228,117 +198,131 @@ derive instance newtypeActual :: Newtype Actual _
...
+                  let
+                    signedCollUnbalancedTx' = fees' >>= \fees ->
+                      prebalanceCollateral
+                        (fees + feeBuffer)
+                        allUtxos
+                        ownAddr
+                        sUbTx'
+                  case signedCollUnbalancedTx' of
+                    Left err -> pure $ Left err
+                    Right signedCollUnbalancedTx ->
```
klntsky, line 297: I think readability of this function can be improved by using ExceptT.

src/QueryM.purs
```diff
@@ -196,6 +196,15 @@ data FeeEstimateError
   = FeeEstimateHttpError Affjax.Error
   | FeeEstimateDecodeJsonError Json.JsonDecodeError
 
+-- No Show instance of Affjax.Error
+instance Show FeeEstimateError where
```
klntsky, line 200: Let's use Show to output strings that are as close to code used to construct the value as possible.

Let's add a new PrettyPrint typeclass for human-readable strings, if this use is common.

src/BalanceTx.purs
vvtran: Yes, there was a short discussion a while back about whether to use `ExceptT` - in the end, we decided against (I can link you on Slack). Another option is to just throw (JavaScript) errors but that feels unpleasant but easier to read at least.

src/QueryM.purs
klntsky: I see, we don't have a valid Show instance for Affjax Error... Can't be helped then.

src/BalanceTx.purs
ngua: If we go with `ExceptT`, that might be something we want to add to the `QueryM` stack. I'll open another issue for standardizing error-handling

ngua (approved):
Nice work. There are undoubtedly more changes that will have to be made as we uncover more bugs, but this gets us to a nice point where we can actually start using the balancer in our testing

PR #88, opened 2022-02-09
Implement deserialization for WitnessSet

### LINKED ISSUE #77: Deserialization for `TransactionWitnessSet`
The type is already defined in `Types.Transaction`, so we just need the ability to deserialize from CSL. This is required for `signTx`

To speed this up, there are several `Maybe` fields we don't care about at the moment that we can just hard-code to `Nothing` for now (e.g. anything Plutus-related). We can open another issue to revisit that when we start working on Plutus support

This is probably the last deserialization-related thing we need to worry about for stage 1

klntsky:
Closes #77


klntsky:
There are no tests for `native_scripts` decoding, because I haven't yet acquired a transaction with it. I'll construct it manually later.

ngua (approved):
Looks good, one small suggestion below

vvtran (commented):
LGTM, few comments.

klntsky:
I added a test for NativeScript, however, it's just a stub because our type is incomplete as well. See https://github.com/Plutonomicon/cardano-browser-tx/issues/91

PR #89, opened 2022-02-10
Basic Nami integration

### LINKED ISSUE #64: Implement core of Nami wallet interface

ngua:
Closes #64 

This depends on #88 being merged


src/QueryM.purs
```diff
@@ -123,6 +129,28 @@ allowError :: (Either Error UtxoQR -> Effect Unit) -> UtxoQR -> Effect Unit
...
+  \(Nami nami) -> nami.getWalletAddress =<< liftEffect (Ref.read nami.connection)
+
+getWalletCollateral :: QueryM (Maybe TransactionUnspentOutput)
+getWalletCollateral = withMWalletAff $
+  \(Nami nami) -> nami.getCollateral =<< liftEffect (Ref.read nami.connection)
+
+signTransaction
+  :: Transaction.Transaction -> QueryM (Maybe Transaction.Transaction)
+signTransaction tx = withMWalletAff $
+  \(Nami nami) -> flip nami.signTx tx =<< liftEffect (Ref.read nami.connection)
```
vvtran, line 146: Entirely optional but did you want to refactor the right part out since it's being repeated three times? Something like (perhaps with a better name).
```
readConnection :: NamiWallet -> Aff NamiConnection
readConnection = liftEffect <<< Ref.read <<< _.connection
```

src/Wallet.purs
```diff
@@ -0,0 +1,119 @@
...
+    , getCollateral
+    , signTx
+    }
+
+  where
+  enable :: Aff NamiConnection
+  enable = Promise.toAffE $ _enableNami
+
+  getWalletAddress :: NamiConnection -> Aff (Maybe Address)
+  getWalletAddress nami = fromNamiHexString _getNamiAddress nami >>= case _ of
```
klntsky, line 67: `traverse`?

klntsky (approved):
All good, please re-merge #88

src/QueryM.purs
ngua: Sorry, somehow missed this comment. I agree that would be nicer

ngua: I can adjust it when working on the Nami proof-of-concept transfer I suppose

PR #97, opened 2022-02-11
Further improvements on Value safety constructors.

vvtran:
- I think it's safe to make `NonAdaAsset` since `CurrencySymbol` and `TokenName` are safe by construction.
- Therefore, I've added a `Newtype` instance for the `NonAdaAsset`.
- General improvements to some of the helpers, especially `mkSingletonValue` to cover more cases.


ngua (approved):
LGTM. Small note about documentation for one function below

Issue #100, opened 2022-02-14
Enable multiple entry points with webpack


ngua:
We have two example modules that can be run in the browser (soon there will be three). It would be nice if we could have multiple entry points configured through webpack to run different examples. Defining multiple entry points itself is not particularly hard,  but we need to get it to work nicely with the bundler and dev server

ngua:
We've changed the way we use webpack, so we don't really need to worry about this anymore

Issue #95, opened 2022-02-11
Add websocket configuration options


ngua:
We can just re-use the existing `ServerConfig` to add configuration options for the Ogmios websocket server to `QueryConfig`. Then we can adjust `mkOgmiosWebSocketAff` to take a `ServerConfig` (instead of `Url`) and construct a URL accordingly

- [x] Export a `defaultWsConfig` with `{secure: false, host: "localhost", port: UInt.fromInt 1337}` as values (these are the defaults for Ogmios)
- [x] Adjust `mkOgmiosWebSocketAff` accordingly

ngua:
Will address this before adding `ogmios-datum-cache` connection to `QueryConfig`

PR #106, opened 2022-02-15
Example `Transaction` built with `QueryM`

### LINKED ISSUE #63: Build a `Transaction` through the `QueryM` interface
This is essentially the core of the stage-1 goal (building a transaction in the browser that works with a light wallet). Since we don't have a constraints API, we need to fill in most of the relevant information by hand 

The transaction will be very simple -- sending two ada to one's own wallet

Here are the rough steps to accomplish this:
- get the wallet's address through Nami interface embedded in `QueryM` (e.g. `getWalletAddress`)
- query the chain with `utxosAt` to get the inputs
- send two ada back to the aforementioned wallet address in the outputs
- fill in the rest of the `UnbalancedTransaction` -- this steps depends on how #37 is resolved (see #61)

ngua:
Closes #63. This just builds a `Transaction` through the `QueryM` interface. Running it locally gets as far as balancing

I've changed `examples/index.js` to run this example instead of the older one. We should figure out #100 in order to enable different entry points to be run

There's a placeholder fee of 0 Ada. In the Plutus definition of `Tx`, the fee is a `Value` and this gets set to `mempty` when the transaction first gets built. I'll open an issue about this so we can address it later

### DIFF
--- a/examples/nami/Pkh2Pkh.purs
+++ b/examples/nami/Pkh2Pkh.purs
@@ -41,9 +41,10 @@ import Wallet (mkNamiWalletAff)
 main :: Effect Unit
 main = launchAff_ $ do
   wallet <- Just <$> mkNamiWalletAff
+  ws <- mkOgmiosWebSocketAff "ws:127.0.0.1:1337"
   tx <- runReaderT
     buildTransaction
-    { ws: {-TODO-}  undefined
+    { ws
     , wallet
     , serverConfig: defaultServerConfig
     }
--- a/examples/nami/Pkh2Pkh.purs
+++ b/examples/nami/Pkh2Pkh.purs
@@ -24,7 +24,7 @@ main = launchAff_ $ do
     }
   undefined
 
-buildTransaction :: QueryM UnbalancedTransaction
+buildTransaction :: QueryM UnbalancedTx
 buildTransaction = do
   ownAddress <-
     maybe
--- a/examples/nami/Pkh2Pkh.purs
+++ b/examples/nami/Pkh2Pkh.purs
@@ -2,14 +2,33 @@ module Examples.Nami.Pkh2Pkh (main) where
 
 import Prelude
 
-import BalanceTx (UnbalancedTransaction)
 import Control.Monad.Error.Class (throwError)
 import Control.Monad.Reader (runReaderT)
+import Data.Array as Array
+import Data.BigInt as BigInt
+import Data.Map as Map
 import Data.Maybe (Maybe(..), maybe)
+import Data.Newtype (unwrap)
+import Data.Tuple (fst)
 import Effect (Effect)
 import Effect.Aff (error, launchAff_)
 import Effect.Aff.Class (liftAff)
-import QueryM (QueryM, defaultServerConfig, getWalletAddress)
+import QueryM (QueryM, defaultServerConfig, getWalletAddress, utxosAt)
+import Types.POSIXTimeRange
+  ( Extended(..)
+  , Interval(..)
+  , LowerBound(..)
+  , UpperBound(..)
+  )
+import Types.Transaction
+  ( NetworkId(..)
+  , Transaction(..)
+  , TransactionOutput(..)
+  , TxBody(..)
+  , emptyTransactionWitnessSet
+  )
+import Types.UnbalancedTransaction (UnbalancedTx(..))
+import Types.Value as Value
 import Undefined (undefined)
 import Wallet (mkNamiWalletAff)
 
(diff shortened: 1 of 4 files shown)


examples/nami/Pkh2Pkh.purs
```diff
@@ -0,0 +1,149 @@
...
+--
+--   - A Nami wallet funded with test Ada ("tAda") and collateral set, If you need
+--     tAda, visit https://testnets.cardano.org/en/testnets/cardano/tools/faucet/
+--
+-- * How to run
+--
+--   The `QueryM` interface requires several external services to be running. From
+--   the repository root, run the following commands:
+--
+--   - `make run-tesnet-node`
```
vvtran, line 18: tesnet typo

examples/nami/Pkh2Pkh.purs
```diff
@@ -0,0 +1,149 @@
...
+--      Starts a testnet Cardano node. May take some time to sync fully
+--
+--   - `make run-tesnet-ogmios`
```
vvtran, line 21: same as above

src/Types/Transaction.purs
```diff
@@ -214,6 +214,16 @@ derive newtype instance Eq TransactionWitnessSet
...
+emptyTransactionWitnessSet :: TransactionWitnessSet
+emptyTransactionWitnessSet = TransactionWitnessSet
+  { vkeys : Nothing
+  , native_scripts : Nothing
+  , bootstraps : Nothing
+  , plutus_scripts : Nothing
+  , plutus_data : Nothing
+  , redeemers : Nothing
+  }
+
```
vvtran, line 226: `make format` will remove the spaces before the colon.

examples/nami/Pkh2Pkh.purs
```diff
@@ -0,0 +1,149 @@
...
+  , TxBody(TxBody)
+  , emptyTransactionWitnessSet
+  )
+import Types.UnbalancedTransaction (UnbalancedTx(UnbalancedTx))
+import Types.Value as Value
+import Untagged.Union (asOneOf)
+import Wallet (mkNamiWalletAff)
+
+main :: Effect Unit
+main = launchAff_ $ do
```
vvtran, line 81: Block Arguments are automatic in PS (unlike Haskell) so the `$` is unnecessary. Although you can keep if you prefer of course.

examples/nami/Pkh2Pkh.purs
```diff
@@ -0,0 +1,149 @@
...
+  wallet <- Just <$> mkNamiWalletAff
+  ws <- mkOgmiosWebSocketAff "ws:127.0.0.1:1337"
```
vvtran, line 83: Is it worth centralising this websocket to a string somewhere, it's also used in tests. Not actually sure how often it will appear in the future admittedly.

examples/nami/Pkh2Pkh.purs
```diff
@@ -0,0 +1,149 @@
...
+        }
+    , requiredSignatories: Map.empty
+    , utxoIndex: Map.empty
+    , validityTimeRange: Interval
+        { from: LowerBound NegInf true
+        , to: UpperBound PosInf true
+        }
+    }
+
+throw :: forall (a :: Type). String -> QueryM a
```
vvtran, line 145: Entirely optional, you could make this more polymorphic via
`throw :: forall (a :: Type) (m :: Type -> Type). MonadAff m => String -> m a`
I suspect some prefer this flexibility and others prefer concrete types for readability?

examples/nami/Pkh2Pkh.purs
```diff
@@ -0,0 +1,149 @@
...
+throw = liftAff <<< throwError <<< error
+
+mthrow :: forall (a :: Type). String -> QueryM (Maybe a) -> QueryM a
```
vvtran, line 148: Same here
`mthrow :: forall (a :: Type)  (m :: Type -> Type). MonadAff m => String -> m (Maybe a) -> m a` (keeping under 80 characters if possible)

vvtran (approved):
LGTM, going to approve but there are a few comments to consider.

examples/nami/Pkh2Pkh.purs
ngua: The `mkOgmiosWebSocketAff` will be overhauled in the future (see #95)

examples/nami/Pkh2Pkh.purs
ngua: If this were to be exported or used with any other monad stack, yes, I would make it more polymorphic. I don't think there would be enough of a benefit here though

examples/nami/Pkh2Pkh.purs
ngua: Honestly I find `BlockArguments` a bit difficult to read, so I'll keep the `$` (if we settle on a consistent style in the future though I won't raise any objections)

PR #110, opened 2022-02-15
Fix `Value` helper and balancer to build transaction

vvtran:
- Balancer had issues which meant we weren't able to build a transaction - this is fixed due to a bug in `Value`.
- Balancer now signs at the very end.
- A few fixes in the post balancer to return Ada change.
- Refining `Show` instances in `Value`


ngua (approved):
Great work getting this done so quickly. I ran the example tx builder from this branch and balancing completes. There may be more issues in the balancer but I think we need to wait until we've tried to submit the built transaction to find them

PR #111, opened 2022-02-15
Provide new Address API (revisit)

adamczykm:
I have simplified address types. They now follow Cardano Serialization Libarary quite directly.

For some reason my editor broke imports formatting. I will run formatter in another PR and add a config to it


ngua (changes requested):
Are you using the PS LSP by any chance? I noticed that also breaks import formatting when auto-completing imports

At any rate, could you run the formatter from the CLI or find another way to revert some of these formatting changes, maybe by merging in master? The diffs are really noisy

src/Serialization.purs
```diff
@@ -193,45 +142,30 @@ convertTxOutputs arrOutputs = do
...
   let
     lovelace = Value.valueToCoin' val
     m = Value.getNonAdaAsset' val
   multiasset <- newMultiAsset
   forWithIndex_ m \scriptHashBytes' values -> do
-    let scriptHashBytes = Value.getCurrencySymbol scriptHashBytes'
+    let mScripthash = scriptHashFromBytes $ Value.getCurrencySymbol scriptHashBytes'
     assets <- newAssets
+
+    scripthash <- errorOnNothing mScripthash
```
klntsky, line 162: Let's use `explain` in cases like this, if we ever get a bug report, a message can save us a lot of time.

src/Types/Aliases.purs
```diff
@@ -4,3 +4,8 @@ module Types.Aliases where
 -- | case Bech32String is not in bech32 format.
 -- | The type alias is here to make the intention of code more readable.
 type Bech32String = String
+
+-- | IMPORTANT: All functions accepting Bech32String should behave correctly in
+-- | case Bech32String is not in bech32 format.
```
klntsky, line 9: s/Bech32String/Base58String

src/Serialization/Address.purs
```diff
@@ -1,376 +1,255 @@
...
+
+byronAddressFromBase58 :: Base58String -> Maybe ByronAddress
+byronAddressFromBase58 = _byronAddressFromBase58 maybeFfiHelper
+
+foreign import _byronAddressFromBytes :: MaybeFfiHelper -> ByteArray -> Maybe ByronAddress
+
+byronAddressFromBytes :: ByteArray -> Maybe ByronAddress
+byronAddressFromBytes = _byronAddressFromBytes maybeFfiHelper
+
+foreign import byronAddressBytes :: ByronAddress -> ByteArray
```
klntsky, line 170: Use `toBytes` from `Serialization` here and below. No need to multiply functions like this.

test/Serialization/Address.purs
```diff
@@ -48,117 +64,95 @@ mPkh = ed25519KeyHashFromBech32 pkhBech32
 mScriptHash :: Maybe ScriptHash
 mScriptHash = scriptHashFromBytes $ hexToByteArrayUnsafe scriptHashHex
 
-suite :: TestPlanM Unit
-suite = do
+addressFunctionsTest :: TestPlanM Unit
```
klntsky, line 67: Why don't you use `test` and `group`? This suite will exit on first fail, I find this not very convenient. Plus, it not possible to use `only` and `skip` helpers to focus on some part of tests when developing.

adamczykm:
> Are you using the PS LSP by any chance? I noticed that also breaks import formatting when auto-completing imports
> 
> At any rate, could you run the formatter from the CLI or find another way to revert some of these formatting changes, maybe by merging in master? The diffs are really noisy

Formatter would make it more noisy, I've fixed some of imports changes but had to turn off LSP so it is cumbersome, I hope that it good enough now. Ill prepare formatting config and PR after this one gets merged

test/Serialization/Address.purs
```diff
@@ -125,7 +91,7 @@ enterpriseAddressFunctionsTest = do
   assertTrue_ $ enterpriseAddressPaymentCred eaddr == keyHashCredential pkh
 
 pointerAddressFunctionsTest :: TestPlanM Unit
-pointerAddressFunctionsTest = do
+pointerAddressFunctionsTest = group "PointerAddress tests" $ do
```
klntsky, line 94: This does not produce any output when the test suite is run, right? I believe it's required to use `test` at least once in a group in order for output to appear.

adamczykm:
Just got burried in merging-in master. Can we resolve this:
| Use toBytes from Serialization here and below. No need to multiply functions like this.
in another PR? I want to discuss it a little bit, but this PR should merge asap. It's really painful to keep merging.

src/Helpers.purs
```diff
@@ -1,10 +1,13 @@
...
-
-import Data.Either (Either)
-import Data.Argonaut (JsonDecodeError, Json, parseJson)
+module Helpers (explain, explainM, parseJsonStringifyNumbers, jsonTurnNumbersToStrings) where
 
 import Prelude
 
+import Data.Argonaut (JsonDecodeError, Json, parseJson)
+import Data.Either (Either(..))
+import Data.Maybe (Maybe(..))
```
vvtran, line 7: explicit imports here.

src/Helpers.purs
```diff
@@ -16,3 +19,13 @@ parseJsonStringifyNumbers s = do
   _ <- parseJson s
   -- valid json ensured at this point
   parseJson $ jsonTurnNumbersToStrings s
+
+explain :: forall a e. e -> Maybe a -> Either e a
```
vvtran, line 23: Kind signature for polymorphic parameters if possible. Also, isn't this the same as `note`? https://pursuit.purescript.org/packages/purescript-either/5.0.0/docs/Data.Either#v:note

src/Helpers.purs
```diff
@@ -16,3 +19,13 @@ parseJsonStringifyNumbers s = do
...
+explain e = case _ of
+  Nothing -> Left e
+  Just x -> pure x
+
+explainM :: forall a. String -> Maybe a -> Effect a
```
vvtran, line 28: Kind signature for polymorphic parameters if possible.

src/QueryM.purs
```diff
@@ -386,39 +385,37 @@ messageFoldF msg acc' func = do
 --------------------------------------------------------------------------------
 -- JsonWsp.Address is a bech32 string, so wrap to Transaction.Types.Bech32
 -- | Converts an JsonWsp.Address to (internal) Address
-ogmiosAddressToAddress :: Address -> Effect (Maybe Transaction.Address)
-ogmiosAddressToAddress ogAddr =
-  newAddressFromBech32 (wrap ogAddr) <#> Deserialization.Address.convertAddress
+ogmiosAddressToAddress :: JsonWsp.Address -> (Maybe Address)
```
vvtran, line 388: Just aesthetic, could remove parenthesis here.

src/QueryM.purs
```diff
@@ -431,15 +428,12 @@ txOutRefToTransactionInput { txId, index } = do
...
-      pure $ wrap
-        { address
-        , amount: value
-        , data_hash: datum >>= hexToByteArray <#> wrap
-        }
+  pure $ wrap
+    { address: address'
+    , amount: value
+    , data_hash: datum >>= hexToByteArray <#> wrap
+    }
```
vvtran, line 439: This is a relief. We could probably remove the `address''` completely with the new functionality and start with `{ address, value, datum }`, then
` address' <- ogmiosAddressToAddress address`

src/BalanceTx.purs
```diff
@@ -633,18 +632,14 @@ getPublicKeyTransactionInput
...
-      pure txOutRef
-    PaymentCredentialScript _ ->
-      Left CannotConvertScriptOutputToTxInput
-
-addressPaymentCredentials :: Address -> PaymentCredential
-addressPaymentCredentials = _.payment <<< unwrap <<< _."AddrType" <<< unwrap
-
-txOutPaymentCredentials :: TransactionOutput -> PaymentCredential
-txOutPaymentCredentials = addressPaymentCredentials <<< _.address <<< unwrap
+  explain CannotConvertScriptOutputToTxInput $ do
```
vvtran, line 635: We could probably remove `explain` and just use `note` from `Data.Either`

src/Serialization/Address.purs
```diff
@@ -1,376 +1,255 @@
...
--- | to use csl functions
-instance (ToCsl (StakeCred d) StakeCredentialCsl, ToCsl (StakeCred p) StakeCredentialCsl) => ToCsl (BaseAddress p d) BaseAddressCsl where
-  toCslRep a = _newBaseAddressCsl
-    { network: networkTagInt a
-    , paymentStakeCred: toCslRep $ StakeCred $ paymentCred a
-    , delegationStakeCred: toCslRep $ StakeCred $ delegationCred a
-    }
+foreign import keyHashCredential :: Ed25519KeyHash -> StakeCredential
+foreign import scriptHashCredential :: ScriptHash -> StakeCredential
+foreign import withStakeCredential :: forall a. { onKeyHash :: Ed25519KeyHash -> a, onScriptHash :: ScriptHash -> a } -> StakeCredential -> a
```
vvtran, line 85: Kind signature if possible

src/Lib.purs
```diff
@@ -1,10 +1,12 @@
...
 
 import Prelude
-import Types.Transaction as Types
-import Data.Maybe (Maybe(..))
-import Data.BigInt as BigInt
 
+import Data.BigInt as BigInt
+import Data.Maybe (Maybe(..))
+import Data.UInt as UInt
+import Serialization.Address (NetworkId, Slot(..))
```
vvtran, line 8: Let's add imports here and above too if possible.

vvtran (commented):
Looks good on the whole, I will try to take another look first thing tomorrorw.. BTW, I think the entire thing needs a `make format` . Also some merge conflicts to deal with.

vvtran:
hey @adamczykm, I just noticed, it looks like the `balanceTx` has been changed to an older version. In particular the `spy` comments have been removed (these aren't so important but they are part of the newer balancer) and now signing should be done at the end.
````
                    Right nonAdaBalancedCollTx ->
                      lmap ReturnAdaChangeError' <$>
                        returnAdaChange ownAddr allUtxos nonAdaBalancedCollTx
                        >>= case _ of
                          Left err -> pure $ Left err
                          Right unsignedTx ->
                            signTransaction unsignedTx
                              <#> note (SignTxError' $ CouldNotSignTx ownAddr)

  -- Some useful spies for debugging.
  -- case signedTx' of
  --   Left err -> pure $ Left err
  --   Right signedTx -> do
  --     let signedTxBody = (unwrap signedTx).body
  --         signedTxOuputs = (unwrap signedTxBody).outputs
  --         x = spy "signedTxInputVal" (getInputValue allUtxos signedTxBody)
  --         y = spy "signedTxOutputs" signedTxOuputs
  --         z = spy "signedTxOutputsVal" (Array.foldMap getAmount signedTxOuputs)
  --         a = spy "signedTxFees" $ (unwrap signedTxBody).fee
  --     pure $ pure $ signedTx
```
should be the last thing in `balanceTxM`. I think there's a few things above it that may be wrong too.  I think `returnAdaChange` is also wrong. I need to head off now but can be more precise about it tomorrow morning.

adamczykm:
@vvtran lol. I don't know how these changes ended up in the PR. Sorry for that. It's fixed now.

src/Serialization/Address.purs
```diff
@@ -1,376 +1,283 @@
...
-  , scriptAddress
-  , scriptHashCredType
-  , networkTag
-  , networkTagInt
-  , paymentCred
-  , delegationCred
-  , toAddressCsl
-  , intToNetTag
-  ) where
+module Serialization.Address where
```
ngua, line 1: Why remove the explicit exports?

src/Serialization/Address.purs
```diff
@@ -1,376 +1,283 @@
...
 import FfiHelpers (MaybeFfiHelper, maybeFfiHelper)
-import Serialization.Csl (class ToCsl, CslType, toCslRep, toCslType)
 import Serialization.Hash (Ed25519KeyHash, ScriptHash)
-import Types.Aliases (Bech32String)
+import Types.Aliases (Bech32String, Base58String)
 import Types.ByteArray (ByteArray)
 
--- | Denotes one of two possible values of Shelley address' network tag.
-data NetworkTag = TestnetTag | MainnetTag
+newtype Slot = Slot UInt
```
ngua, line 16: Why change from `BigInt` to `UInt`? The Plutus `Slot` is a wrapped `Integer`

src/Serialization/Address.purs
```diff
@@ -1,376 +1,283 @@
...
-    TestnetTag -> "TestnetTag"
-    MainnetTag -> "MainnetTag"
+newtype TransactionIndex = TransactionIndex UInt
 
--- | Simple safety wrapper to ensure that only valid types are provided
--- | as address' stake credentials.
-newtype StakeCred a = StakeCred a
+derive instance Eq TransactionIndex
+derive instance Newtype TransactionIndex _
+derive newtype instance Show TransactionIndex
```
ngua, line 26: I'd prefer not to use `newtype` instances for `Show`

src/Serialization/Address.purs
```diff
@@ -1,376 +1,283 @@
...
-netTagToInt = case _ of
-  TestnetTag -> 0
-  MainnetTag -> 1
-
-intToNetTag :: Int -> Maybe NetworkTag
-intToNetTag = case _ of
-  0 -> Just TestnetTag
-  1 -> Just MainnetTag
-  _ -> Nothing
+newtype NetworkId = NetworkId Int
```
ngua, line 129: I liked the previous representation of `NetworkTag`, what was the reason for this change? In the functions that return a `BaseAddress` below, couldn't `netTagToInt` be used?

src/Serialization/Address.js
```diff
@@ -1,146 +1,112 @@
-var CardanoWasm;
-if (typeof BROWSER_RUNTIME != 'undefined' && BROWSER_RUNTIME) {
-    CardanoWasm = require('@ngua/cardano-serialization-lib-browser');
-} else {
-    CardanoWasm = require('@ngua/cardano-serialization-lib-nodejs');
-}
+const CardanoWasm = require("@ngua/cardano-serialization-lib-nodejs");
```
ngua, line 1: This breaks the browser integration, so this change needs to be reverted

ngua (changes requested):
I've left a few comments below. There's a change in how the CSL is imported in `Serialization/Address.js` that breaks browser integration

vvtran (commented):
The changes to the balancer and other things I've mentioned look good. @ngua mentioned a few things that should be addressed, otherwise I think this should be okay to merge in. I would agree with using `genericShow` where possible in `Serialization` although I'm definitely guilty of using newtype Show for `UtxoM` by taking code from `UtxoQR`.

src/Serialization/Address.purs
adamczykm: It's based on cardano serialization library which uses UInt32. I think it is safer to respect this 'lower bound'

src/Serialization/Address.purs
adamczykm: Forgot about those

src/Serialization/Address.purs
adamczykm: I decided to go for a thinner ffi that follows the library types, were kind of force to use it anyway and having only a thin wrapper means less code and types. I can change it back.

src/Serialization/Address.js
adamczykm: Strange, I don't remember dealing with this as a conflict during merges. This branch is definitely too long-lived

src/Serialization/Address.purs
adamczykm: Created an issue for this.

src/Serialization/Address.purs
adamczykm: How it is worse than `genericShow`?

src/Serialization/Address.purs
ngua: OK, I think we should keep your changes for now then. If the Plutus representation and CSL diverge, we should favor the latter, I think. But we can discuss that more and change this later if we decide otherwise

src/Serialization/Address.purs
ngua: Unless there is a compelling reason otherwise, I think a type's `Show` instance should produce a valid PS expression that would produce the shown value if evaluated

src/Serialization/Address.purs
ngua: I personally prefer the representation you had before

adamczykm: Created issue that addresses this

src/Serialization/Address.purs
adamczykm: Addressed in an issue

test/Utils.purs
```diff
@@ -1,27 +1,26 @@
...
 import Effect.Class (class MonadEffect, liftEffect)
 import Effect.Exception (throwException, throw)
-import TestM (TestPlanM)
 import Type.Prelude (Proxy)
 
 foreign import unsafeCall :: forall a b. Proxy b -> String -> a -> b
 
 -- | Make boolean a test
-assertTrue :: String -> Boolean -> TestPlanM Unit
+assertTrue :: forall (m :: Type -> Type). Applicative m => MonadEffect m => String -> Boolean -> m Unit
```
klntsky, line 15: Why a kind annotation here? I'm so used to reading types like this without it I doubt it provides value. Plus it's obvious from `Applicative` constraint.

adamczykm: I was asked to add kind signatures by @vvtran , I think it follows haskells style guide, I don't like it either, unless necessary.

ngua (approved):

PR #114, opened 2022-02-16
Implement proper NativeScript (de)serialization 

### LINKED ISSUE #91: Reconsider NativeScript definition
As for now, our enum type does not include payloads of various NativeScript kinds. The change will affect both serialization and deserialization.

klntsky:
Closes #91


src/Serialization/NativeScript.purs
```diff
@@ -0,0 +1,76 @@
...
+
+import Prelude
+
+import Data.BigInt as BigInt
+import Data.Maybe (Maybe)
+import Data.Traversable (for, traverse)
+
+import Deserialization.FromBytes (fromBytes)
+import FfiHelpers (ContainerHelper, containerHelper)
+import Serialization.Types (Ed25519KeyHash, NativeScript, NativeScripts, ScriptAll, ScriptAny, ScriptNOfK, ScriptPubkey, TimelockExpiry, TimelockStart)
```
adamczykm, line 14: how about breaking long import lines?

test/Deserialization.purs
```diff
@@ -2,27 +2,31 @@ module Test.Deserialization (suite) where
...
+import Deserialization.NativeScript as NSD
 import Deserialization.UnspentOutput (convertUnspentOutput, mkTransactionUnspentOutput, newTransactionUnspentOutputFromBytes)
 import Deserialization.WitnessSet (deserializeWitnessSet, convertWitnessSet)
 import Serialization as Serialization
 import Serialization.BigNum (bigNumFromBigInt)
+import Serialization.NativeScript (convertNativeScript) as NSS
 import Serialization.Types (TransactionUnspentOutput)
 import Serialization.WitnessSet as SW
-import Test.Fixtures (addressString1, txInputFixture1, txOutputFixture1, utxoFixture1, utxoFixture1', witnessSetFixture1, witnessSetFixture2, witnessSetFixture2Value, witnessSetFixture3, witnessSetFixture3Value, witnessSetFixture4)
+import Test.Fixtures (addressString1, nativeScriptFixture1, nativeScriptFixture2, nativeScriptFixture3, nativeScriptFixture4, nativeScriptFixture5, nativeScriptFixture6, nativeScriptFixture7, txInputFixture1, txOutputFixture1, utxoFixture1, utxoFixture1', witnessSetFixture1, witnessSetFixture2, witnessSetFixture2Value, witnessSetFixture3, witnessSetFixture3Value, witnessSetFixture4)
```
adamczykm, line 27: ditto

adamczykm (commented):
The majority of code is related to having cardano serialisation lib's NativeScript(s) types converted to `Types.Transaction.NativeScript`, but is it worth it? I mean do we intend to use this algebraic data type in the project's codebase or we will just be passing it forward. In case of the latter we could just stick to using the serialisation lib's foreign type? Or I'm missing something?

src/FfiHelpers.js
```diff
@@ -1,10 +1,22 @@
 /* global exports */
 
+const unpack = obj => {
```
ngua, line 3: Maybe you could add a comment here, similar to the one for `pack` below

ngua (approved):

src/Deserialization/Address.js
```diff
@@ -1,4 +1,4 @@
-/* global require exports */
+/* global require exports BROWSER_RUNTIME */
```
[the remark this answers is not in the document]
klntsky: `/* global ... */` directives are for some editors to silence "unused variable" warnings. In particular, it works with emacs js2-mode.

Apart from serving their direct purpose, I find them useful because they say something about expectations from runtime.

klntsky:
> I mean, do we intend to use this algebraic data type in the project's codebase or we will just be passing it forward.

That's a good question. I am not sure.

The intent of this PR is to fix something that was already broken (incomplete NativeScripts prevented us from having `serialize . deserialize = id` property). To achieve that, we could go the other route in other places as well - for example, we could just pass through serialized `WitnessSet`.

adamczykm:
I spent some time thinking about the approach with `Address` stuff and decided to go close to the actual implementation. Our project seems to work as a messenger between different libraries such as the CSL and wallets, and backend services. I don't think I got a clear vision on its entirety and final shape but given the amount of code needed for creating native PS data types and required conversion stuff, this might be not worth it.

adamczykm (commented):
Looks okay, apart from too long import lines. I'm thinking about improving `make format` behaviour so that it automatically splits import lines when it is longer than X.

PR #115, opened 2022-02-16
`Monoid` for  `Types.Transaction` 

vvtran:
- Monoid instance for `Transaction`


src/Types/Transaction.purs
```diff
@@ -29,6 +35,31 @@ newtype Transaction = Transaction
...
+-- if both are false, take the rightmost. I'm not sure if this is the behaviour
+-- we actually want. What determines if something "is_valid" or not?
+instance semigroupTransaction :: Semigroup Transaction where
+  append (Transaction tx) (Transaction tx') =
+    Transaction
+      if tx.is_valid then
+        if tx'.is_valid then
+          { body: tx.body <> tx'.body
+          , witness_set: tx.witness_set <> tx'.witness_set
+          , is_valid: true
```
klntsky, line 48: I'm curious what `is_valid` is used for and why it's different in the instance below.

vvtran: Admittedly I'm not sure what determines a "valid" transaction or "invalid" for that matter. I've added a lot of comments in this PR as I'm unsure on a few things, this being one of them. I've made the logic here essentially ignore invalid transactions for the `Monoid` instance - hence why `false` is like the identity. Not sure if this is 100% what we want.

vvtran (commented):
Oops, please ignore. I got confused between tabs.

vvtran:
I've added a bunch of comments in this PR which will probably need addressing, any feedback on them would be much appreciated.

ngua (commented):
I tried to address most of your comments. In general, we need to make arbitrary choices for several of the `Transaction` fields, which I think is fine as long as there is some consistency. For `Maybe` fields, I think using `Last` is fine. This is consistent with some of the other right-biased updates here (like `appendRightHashMap`)

vvtran:
@ngua good idea of the `guard`, hopefully I implemented what you intended, notice that the neutral element now has `true` for `is_valid`

ngua (approved):
Looks good now

Issue #112, opened 2022-02-16
Define datum type and instances


ngua:
We need to add datum-querying capabilities to `QueryM` to reach our stage 3 goal. Part of this is defining a `Data` (or `Datum`) type that we can decode from JSON and serialize across the FFI boundary

Here is the definition of Plutus' `Data`:

```haskell
data Data =
      Constr Integer [Data]
    | Map [(Data, Data)]
    | List [Data]
    | I Integer
    | B BS.ByteString
    deriving stock (Show, Eq, Ord, Generic)
```

When we implement datum-querying through `ogmios-datum-cache`'s websocket interface, we will need to decode the `Data` from JSON. [Here](https://github.com/mlabs-haskell/ogmios-datum-cache#getdatumbyhash) is an example response showing how the type is encoded in the jsonwsp response

In order to actually use the datums in transactions, we also need to serialize them to the CSL representations. [Here](https://github.com/ngua/cardano-serialization-lib/blob/8b7579084dd3eb401a14a3493aa2e91778d48b66/rust/pkg/cardano_serialization_lib.js.flow#L2945) is how they are defined in CSL

klntsky:
I can deal with serialization after I get through what's on my plate currently. If anyone wants to work on ogmios part, let's first merge just the datatype, so that there's less merge noise.

@ngua by serialization you mean de-serialization as well, right?

ngua:
Closed by #116 and in favor of more specific issues

PR #116, opened 2022-02-16
Add PlutusData type

klntsky:
Related to #112


adamczykm (approved):
Isn't it a good one to introduce `purescript-bridge` ?

ngua (approved):

ngua:
@adamczykm Hopefully we can use `purescript-bridge` in the future. We need datum support for stage 3, though, so we can use this for now

vvtran (approved):

PR #121, opened 2022-02-17
Add `submitTx` to Nami interface & update example module

ngua:
This basically closes stage 2. There are still some issues, but transaction submission is working with Nami

- Adds `submitTx` to `NamiWallet` with a `QueryM` wrapper
- Updates the example module to submit the built transaction
- A hack: adds some lovelace to the estimated fees until #123 is resolved, otherwise the tx submission fails

An exception is generated somewhere in `QueryM` after the transaction is submitted. It appears that it's coming from the WS interface. I'm going to open a separate issue for that


ngua:
I'll probably need to make some adjustments to this after #111 is merged and potentially after the balancer is adjusted, so I'm converting to a draft for now

src/Wallet.js
```diff
@@ -15,3 +15,12 @@ exports._getNamiCollateral = (nami) => () =>
 
 // _signTxNami :: String -> NamiConnection -> Effect (Promise String)
 exports._signTxNami = (txHex) => (nami) => () => nami.signTx(txHex);
+
+// _submitTxNami :: String -> NamiConnection -> Effect (Promise String)
+exports._submitTxNami = (txHex) => (nami) => () => {
```
vvtran, line 20: An aside: still learning JS but is there a reason for parameter parenthesis convention as we seem to have both? Not a request to change, just curious.

vvtran (approved):
LGTM, just small changes.

src/Wallet.js
ngua: No reason, I just have autoformat enabled. It's not necessary

PR #128, opened 2022-02-17
Factor Wallet sum type out of `utxosAt`, tx logging and clean up `balanceTx` nesting

vvtran:
- The pkh2pkh example is still showing ~ +4.83 Ada for the user when trying to submit with their UI collateral option.
- To submit a tx via Nami, it appears to be a requirement that the collateral is not used in the transaction input. Although theoretically, this is allowed: if a tx fails to validate, the collateral is the taken and the inputs (which may include collateral) aren't spent, if the tx validates then the collateral is left alone and the input (which again may include the collateral) is spent. In both scenarios, the collateral is used exactly once. For Nami specifically, if collateral is part of the input, then +5 Ada is automatically added to the transaction for the user despite the transaction balancing internally.
- Therefore, we want `utxosAt` for a Nami wallet to specifically filter out the collateral (since we're using Ogmios to get the utxo set).
- Added some logging for transactions.
- Should we decide to submit a tx via Ogmios, we can add the collateral back in, although the modularity of `utxosAt` may be useful if we need specific behaviour on this function for different wallets, although this should be unlikely and hopefully not required.
- Clean up balanceTxM with `ExceptT`

### DIFF
--- a/src/BalanceTx.purs
+++ b/src/BalanceTx.purs
@@ -380,7 +380,7 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) =
   -- 1) A new utxo doesn't need to be created (no extra fees).
   -- 2) A new utxo needs to be created but the buffer provides enough Ada to
   -- create the utxo, cover any extra fees without a further need for looping.
-  -- This buffer could potentially be exactly calculated using (the Haskell server)
+  -- This buffer could potentially be calculated exactly using (the Haskell server)
   -- https://github.com/input-output-hk/plutus/blob/8abffd78abd48094cfc72f1ad7b81b61e760c4a0/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs#L723
   -- The trade off for this set up is any transaction requires this buffer on
   -- top of their transaction as input.
--- a/src/BalanceTx.purs
+++ b/src/BalanceTx.purs
@@ -404,11 +404,12 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) = do
   feeBuffer :: BigInt
   feeBuffer = fromInt 500000
 
+-- Logging for Transaction type
 logTx'
   :: forall m. MonadEffect m => String -> Utxo -> Transaction -> m Unit
 logTx' msg utxos (Transaction { body: body'@(TxBody body) }) =
   traverse_ (log <<< (<>) msg)
-    -- Add to this if you to log more
+    -- Add to log more:
     [ "Input Value: " <> show (getInputValue utxos body')
     , "Output Value: " <> show (Array.foldMap getAmount body.outputs)
     , "Fees: " <> show body.fee
--- a/src/BalanceTx.purs
+++ b/src/BalanceTx.purs
@@ -290,11 +290,11 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) =
       returnAdaChange ownAddr allUtxos nonAdaBalancedCollTx <#>
         lmap ReturnAdaChangeError'
     -- Sign transaction:
-    signedTx <- ExceptT $
+    balancedTx <- ExceptT $
       signTransaction unsignedTx <#> note (SignTxError' $ CouldNotSignTx ownAddr)
 
     -- Logs final balanced tx and returns it
-    ExceptT $ logTx "Post-balancing Tx " allUtxos signedTx <#> Right
+    ExceptT $ logTx "Post-balancing Tx " allUtxos balancedTx <#> Right
   where
   prebalanceCollateral
     :: BigInt
--- a/src/BalanceTx.purs
+++ b/src/BalanceTx.purs
@@ -278,8 +278,8 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) =
     -- Prebalance collaterised tx without fees:
     ubcTx <- except $
       prebalanceCollateral zero allUtxos ownAddr unbalancedCollTx
-    fees <- ExceptT $ calculateMinFee' ubcTx <#> lmap CalculateMinFeeError'
     -- Prebalance collaterised tx with fees:
+    fees <- ExceptT $ calculateMinFee' ubcTx <#> lmap CalculateMinFeeError'
     ubcTx' <- except $
       prebalanceCollateral (fees + feeBuffer) allUtxos ownAddr ubcTx
 
--- a/src/BalanceTx.purs
+++ b/src/BalanceTx.purs
@@ -358,29 +358,21 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) =
     utxoIndex'
     ownAddr'
     (Transaction tx'@{ body: txBody' }) =
-    do
-      pure $ preBalanceTxBody
+    runExceptT do
+      txBodyWithoutFees' <- except $ preBalanceTxBody
         minUtxos'
         zero
         utxoIndex'
         ownAddr'
         txBody'
-      >>= case _ of
-        Left err -> pure $ Left err
-        Right txBodyWithoutFees' -> do
-          let
-            tx'' :: Transaction
-            tx'' = wrap tx' { body = txBodyWithoutFees' }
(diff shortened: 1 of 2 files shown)


src/BalanceTx.purs
```diff
@@ -262,8 +266,9 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) = do
         Just utxos, Just utxoIndex'' -> do
           let
             -- Combines utxos at the user address and those from any scripts
-            -- involved with the contract in the unbalanced transaction. Don't
-            -- add collateral to this for now.
+            -- involved with the contract in the unbalanced transaction.
+            -- For submitting a tx via Nami, this shouldn't include the collateral.
+            -- This is vital for balancing with Nami tx submission.
```
adamczykm, line 271: If this is vital how about moving it into the function's docstring, that'd help with visibility.

src/BalanceTx.purs
```diff
@@ -295,21 +302,15 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) = do
...
-  --     let signedTxBody = (unwrap signedTx).body
-  --         signedTxOuputs = (unwrap signedTxBody).outputs
-  --         x = spy "signedTxInputVal" (getInputValue allUtxos signedTxBody)
-  --         y = spy "signedTxOutputs" signedTxOuputs
-  --         z = spy "signedTxOutputsVal" (Array.foldMap getAmount signedTxOuputs)
-  --         a = spy "signedTxFees" $ (unwrap signedTxBody).fee
-  --     pure $ pure $ signedTx
+                            -- Some post balancer logging:
+                            case signedTx' of
+                              Left err -> pure $ Left err
```
adamczykm, line 310: This is the 3rd level of Either case split where the Left branch just passes up the error. Can we de-nest it somehow, for example by introducing a local helper function?

adamczykm: or the 5th.

src/BalanceTx.purs
```diff
@@ -403,6 +404,21 @@ balanceTxM (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) = do
   feeBuffer :: BigInt
   feeBuffer = fromInt 500000
```
adamczykm, line 405: unrelated: is this an arbitrary number, what is the logic behind it?

src/BalanceTx.purs
vvtran: Good shout.

adamczykm (commented):
otherwise LGTM

src/BalanceTx.purs
vvtran: We expect that the user has a minimum amount of Ada (this buffer) on top of their transaction, so by the time the prebalancer finishes, should it succeed, there will be some excess Ada (since all NonAda is balanced). In particular, this excess Ada will be at least this buffer.

Hopefully at this point, there will be enough excess Ada to just simply return the Ada to the user, whether by the same utxo (shouldn't cost) or a new utxo (incurring more cost but our buffer should "just" cover it).

I think this buffer can be calculated exactly as an upperbound using this https://github.com/input-output-hk/plutus/blob/8abffd78abd48094cfc72f1ad7b81b61e760c4a0/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Evaluation/Machine/Cek/Internal.hs#L723 I should probably make an enhancement issue for this as a use case for our Haskell server.

vvtran: The downside is the input requires more Ada of course.

src/QueryM.purs
```diff
@@ -414,6 +427,18 @@ utxosAt = addressToOgmiosAddress >>> getUtxos
     in
       (wrap <<< Map.fromFoldable) <$> out
 
+-- Nami appear to remove collateral from the utxo set, so we shall do the same.
+-- This is crucial if we are submitting via Nami. If we decide to submit with
+-- Ogmios, we can remove this.
+utxosAtNami :: Address -> QueryM (Maybe Transaction.UtxoM)
```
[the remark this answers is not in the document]
vvtran: Will look into it.

src/BalanceTx.purs
vvtran: Oh, and yes it is arbitrary at the moment.

adamczykm: Maybe you can put a part of this explanation as a comment there?

vvtran: Will do.

adamczykm (commented):
:+1:

src/BalanceTx.purs
ngua: @vvtran If you wanted to, I don't see why you couldn't use `ExceptT`/`runExceptT` locally in situations where the `Either`s get deeply nested

src/QueryM.purs
```diff
@@ -393,9 +393,22 @@ addressToOgmiosAddress :: Address -> JsonWsp.Address
...
+-- Add more wallet types here:
+utxosAtByWallet
+  :: Address -> Wallet -> QueryM (Maybe Transaction.UtxoM)
+utxosAtByWallet addr (Nami _) = utxosAtNami addr
+-- Unreachable but helps build when we add wallets, most of them shouldn't
+-- require any specific behaviour.
+utxosAtByWallet addr _ = utxosAt'' addr
+
+-- Gets utxos at an (internal) Address in terms of (internal) Transaction.Types.
+utxosAt'' :: Address -> QueryM (Maybe Transaction.UtxoM)
```
ngua, line 410: I'm not a big fan of multiple primes, even for local bindings, but especially for top-level definitions I find it a little confusing. Could you think of a more descriptive name, or perhaps move some of the top-level `utxosAt...` effects to a `where` clause?

ngua (approved):
The logic looks good. Left some comments below that you may want to address, but nothing critical

src/BalanceTx.purs
vvtran: @ngua ah, this could work, perhaps that's what other people have meant and I've misunderstood.

src/QueryM.purs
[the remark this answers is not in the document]
vvtran: I will probably leave it then, just to keep this PR tidy.

adamczykm:
👌

src/BalanceTx.purs
```diff
@@ -239,78 +244,58 @@ calculateMinFee' = calculateMinFee >>> map (rmap unwrap)
...
-  --         x = spy "signedTxInputVal" (getInputValue allUtxos signedTxBody)
-  --         y = spy "signedTxOutputs" signedTxOuputs
-  --         z = spy "signedTxOutputsVal" (Array.foldMap getAmount signedTxOuputs)
-  --         a = spy "signedTxFees" $ (unwrap signedTxBody).fee
-  --     pure $ pure $ signedTx
 
+-- | Balances an unbalanced transaction. For submitting a tx via Nami, the
+-- utxo set shouldn't include the collateral which is vital for balancing.
+-- In particular, the transaction inputs must not include the collateral.
+balanceTxM :: UnbalancedTx -> QueryM (Either BalanceTxError Transaction)
```
klntsky, line 251: Just a note: wouldn't it be a good idea to make transaction type wallet-specific?

Totally not a change request.

src/BalanceTx.purs
```diff
@@ -421,130 +433,124 @@ returnAdaChange
...
+                          }
+                          `Array.cons` txBody.outputs
+                    }
+
+              tx' :: Transaction
+              tx' = wrap tx { body = txBody' }
+
+            fees'' <- lmap ReturnAdaChangeCalculateMinFee <$> calculateMinFee' tx'
+            -- fees should increase.
+            pure case fees'' of
```
klntsky, line 526: `pure $ fees'' >>= \fees' -> ...` - less indentation.

klntsky (approved):

ngua (approved):
Using `ExceptT` internally really cleaned things up, looks good

src/BalanceTx.purs
vvtran: This is a possibility and it depends how the project progresses and what wallets we allow. Eventually, we'll want multisig (which requires signing from different entities from potentially different wallets(?)) and this would still need to be able to balance a transaction for those various signatures.

PR #130, opened 2022-02-17
Add native assets parsing.

adamczykm:
- [x]  parsing stuff
- [x]  tests!
- [x] what about the temp function


src/Types/JsonWsp.purs
```diff
@@ -153,6 +196,33 @@ parseMirror = caseJsonObject (Left (TypeMismatch "expected object")) $
       pure { step, id }
   )
 
+newtype Assets = Assets (Map CurrencySymbol (Map TokenName BigInt))
+
+instance DecodeJson Assets where
+  decodeJson j = do
+    assets :: Array (String /\ String) <- toUnfoldable <$> decodeJson j
```
ngua, line 203: I only took a brief look, so this might be completely off, but:

I think the parse issue you're having might be coming from the `Map` decoder ([here](https://github.com/purescript-contrib/purescript-argonaut-codecs/blob/b0a041d92bfd548e2cd793cc7c02363464325a13/src/Data/Argonaut/Decode/Decoders.purs#L190)) which expects the JSON to be a list of "tuples"

adamczykm: I thought so too, but then I directly tried `decodeMap`

src/Types/Value.purs
```diff
@@ -156,7 +156,7 @@ unsafeAdaToken = TokenName mempty
 -- | exported
 mkTokenName :: ByteArray -> Maybe TokenName
 mkTokenName byteArr =
-  if byteLength byteArr > 0 then pure $ TokenName byteArr else Nothing
+  if byteLength byteArr <= 32 then pure $ TokenName byteArr else Nothing
```
ngua, line 159: I might be completely mistaken about this, but are we going to end up with two different encodings of token names? CSL uses CBOR, but I don't think Ogmios does

ngua (commented):
The parsing itself looks good, but I'm not sure if our token representation is consistent

src/Types/JsonWsp.purs
```diff
@@ -153,6 +188,37 @@ parseMirror = caseJsonObject (Left (TypeMismatch "expected object")) $
       pure { step, id }
   )
 
+newtype Assets = Assets (Map CurrencySymbol (Map TokenName BigInt))
+
+instance DecodeJson Assets where
+  decodeJson j = do
+    wspAssets :: Array (String /\ String) <- FO.toUnfoldable <$> decodeJson j
+    Assets <<< Map.fromFoldableWith Map.union <$> sequence (uncurry decodeAsset <$> wspAssets)
```
vvtran, line 196: Could I just confirm we're relying on duplicate keys to be impossible for JSON here? Because `Map.union` has a left bias on duplicate keys and wanted to make sure we aren't "throwing away" `BigInt` (for a given CurrencySymbol and duplicate TokenNames) which is fine if `"assets"` fields are indeed unique.

vvtran (commented):
LGTM,.

fixtures/test/parsing/json_stringify_numbers/expected/t4.json
```diff
@@ -0,0 +1 @@
+[{"type":"jsonwsp/response","version":"1.0","servicename":"ogmios","methodname":"Query","result":[[{"txId":"1ae7bbe9ed3d0205b3704446678c3a9e777ba41a65a3115d066dd4eabe295839","index":"0"},{"address":"addr_test1qr7g8nrv76fc7k4ueqwecljxx9jfwvsgawhl55hck3n8uwaz26mpcwu58zdkhpdnc6nuq3fa8vylc8ak9qvns7r2dsysp7ll4d","value":{"coins":"1000000000","assets":{"0d94e174732ef9aae73f395ab44507bfa983d65023c11a951f0c32e4.815d":"2","0d94e174732ef9aae73f395ab44507bfa983d65023c11a951f0c32e4.a56e":"9","4acf2773917c7b547c576a7ff110d2ba5733c1f1ca9cdc659aea3a56":"15","b16b56f5ec064be6ac3cab6035efae86b366cc3dc4a0d571603d70e5.f7c5951ec5":"9","bd039f956f4b302f3ab6fc7c4bac3350a540f44af81a8492194dd2c2":"12","bd039f956f4b302f3ab6fc7c4bac3350a540f44af81a8492194dd2c2.bbe6":"13"}},"datum":null}]],"reflection":{"step":"INIT","id":"Hello123"}}]
```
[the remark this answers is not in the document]
adamczykm: There are two ways to do it. One would be to change the algorithm that pits quotes on numbers ( it also removes white spaces). The other is to remove white spaces after loading the fixture file. The second one is more difficult. I can go with the first one if you want it.

adamczykm:
> The parsing itself looks good, but I'm not sure if our token representation is consistent

@ngua can you elaborate?

src/Types/JsonWsp.purs
adamczykm: Duplicated keys - a.k.a. more than one entry for the same pair of currencysymbol and TokenName will cause data loss you mentioned above.
Should I replace it with `unionWith (+)` ?

vvtran: I think so yes, just to be safe.

fixtures/test/parsing/json_stringify_numbers/expected/t4.json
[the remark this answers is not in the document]
vvtran: I meant just running `format` in something like http://jsonviewer.stack.hu/ then pasting in manually which felt quick (or if you have it in your text editor). Have I misunderstood something? Were you intending on automating format on `expected` json files?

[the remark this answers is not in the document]
adamczykm: nvm, done

src/Types/Value.purs
adamczykm: I'm not sure why would it be looking at that change. Both CSL and Ogmios limits the size of TokenName to 32 bytes. I'm not sure what do you mean.

ngua:
Sorry, I shouldn't have left my previous comment on `mkTokenName` as its actually more general. To elaborate more, I was wondering if we might end up with two inconsistent ways of representing a `TokenName`, but I think I was mistaken. At first I thought we were deserializing `AssetName`s using `to_bytes()` and then creating a `TokenName` from the resulting `ByteArray` (which would be in CBOR format), but I just double-checked and it appears we are actually using `name()`. So there should be no inconsistency

ngua (approved):

vvtran (approved):

PR #136, opened 2022-02-21
Configuration options for WS URLs

### LINKED ISSUE #95: Add websocket configuration options

ngua:
Closes #95. Will make it more convenient as we add more websocket connections to `QueryConfig` (e.g. `ogmios-datum-cache`)

### DIFF
--- a/examples/nami/Pkh2Pkh.purs
+++ b/examples/nami/Pkh2Pkh.purs
@@ -52,6 +52,7 @@ import Effect.Aff.Class (liftAff)
 import Effect.Console as Console
 import QueryM
   ( QueryM
+  , defaultOgmiosWsConfig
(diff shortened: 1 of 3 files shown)


vvtran (approved):
LGTM.

PR #138, opened 2022-02-22
`TxConstraints` and `Interval` API

vvtran:
- https://github.com/Plutonomicon/cardano-browser-tx/issues/101 (Constraints API)
- Interval/POSIXTime API recreated since some functionality is needed for Constraints - tried to recreate everything for completion.
- Renamed a few Data types to match `Plutus` to make things more ergonomical when copying Haskell code across.
- A handful of typeclass instances.


ngua (commented):
This is coming along very nicely. When it comes time to implement the constraints resolution, don't hesitate to drop constraints that are not critical to support. I think we only need to focus on supporting a subset of the most commonly used ones

src/Types/PlutusData.purs
```diff
@@ -23,3 +36,40 @@ derive instance Generic PlutusData _
...
+derive newtype instance Eq Datum
+derive newtype instance Ord Datum
+derive instance Newtype Datum _
+derive instance Generic Datum _
+
+instance showDatum :: Show Datum where
+  show = genericShow
+
+-- Do we want a newtype?
+type DatumHash = DataHash
```
ngua, line 75: I don't think we even need the type synonym, unless I'm missing something

src/Types/UnbalancedTransaction.purs
```diff
@@ -64,7 +84,7 @@ newtype UnbalancedTx = UnbalancedTx
   { transaction :: Transaction
   , requiredSignatories :: Map PaymentPubKeyHash (Maybe PaymentPubKey)
   , utxoIndex :: Map TxOutputRef ScriptOutput
-  , validityTimeRange :: POSIXTimeRange
+  , validityTimeRange :: POSIXTimeRange -- Should we remove this? I think we have this info inside Transaction?
```
ngua, line 87: We probably can remove this. We should also do something about `requiredSignatories` and `required_signers` which I believe is similarly duplicated

src/Types/PlutusData.purs
```diff
@@ -23,3 +36,40 @@ derive instance Generic PlutusData _
 
 instance Show PlutusData where
   show x = genericShow x
+
+-- On-chain PlutusData. Do we need this representation to separate when writing
+-- the Contract Monad?
+newtype BuiltinPlutusData = BuiltinPlutusData PlutusData
```
ngua, line 42: Do we really need to maintain a distinction here like Plutus does?

src/Types/PlutusData.purs
vvtran: Yeah, it's probably fine without, I was following `type TxOutputRef = TransactionInput` to make it closer to Plutus when people "copy and paste" code over but can remove.

src/Types/PlutusData.purs
vvtran: I would be very glad to see the back of this. My only concern is, do we want a distinction when writing the contract Monad? In particular, do we need to separate on-chain and off chain data so we can't off-chain functions on chain and vice versa? I feel like I don't have a good understanding of the bigger picture ATM (I guess with regards to data serialisation/deserialisation, Ogmios and related helpers).

If so, we still need to keep `ToPlutusData` right? Although have it map to `PlutusData` instead of `BuiltinPlutusData`?

src/Types/PlutusData.purs
ngua: Hmmmm. That's a good point

src/Types/PlutusData.purs
ngua: Yes, we should still have the two typeclasses. For `BuiltinData`, the issue with Plutus is that it is quite possible to maximize code reuse between on-chain and off-chain Plutus code. Our `Contract` interface will be in PS, though. So I don't think we need to worry about the distinction

vvtran:
There's one stub function and potential fixes required in this but `Interval` and `TxConstraints` are essentially replicated. I haven't implemented `processConstraints`, but will do as part of `ScriptLookups` https://github.com/Plutonomicon/cardano-browser-tx/issues/122

I also renamed a few things, so they match the Plutus name. In some cases, I prefer our own names but it will make things more ergonomic when copying & pasting Haskell code over.

src/Types/TxConstraints.purs
```diff
@@ -0,0 +1,370 @@
...
+  | MustBeSignedBy PubKeyHash
+  | MustSpendAtLeast Value
+  | MustProduceAtLeast Value
+  | MustSpendPubKeyOutput TxOutRef
+  | MustSpendScriptOutput TxOutRef Redeemer
+  | MustMintValue MintingPolicyHash Redeemer TokenName BigInt
+  | MustPayToPubKey PubKeyHash Value
+  | MustPayToOtherScript ValidatorHash Datum Value
+  | MustHashDatum DatumHash Datum
+  | MustSatisfyAnyOf (Array TxConstraint)
```
klntsky, line 81: I'm curious why is it a 2D-list in plutus, but 1D here?

src/Types/TxConstraints.purs
```diff
@@ -0,0 +1,370 @@
...
+
+instance showTxConstraint :: Show TxConstraint where
+  show x = genericShow x
+
+newtype InputConstraint (i :: Type) = InputConstraint
+  { icRedeemer :: i
+  , icTxOutRef :: TxOutRef
+  }
+
+derive instance genericInputConstraint :: Generic (InputConstraint i) _
```
adamczykm, line 94: do we need names for instances? they are noisy and usually bring no value :shrug:

src/Types/TxConstraints.purs
```diff
@@ -0,0 +1,370 @@
...
+  -> TxConstraints i o
+mustPayToOtherScript vh dt vl =
+  singleton (MustPayToOtherScript vh dt vl)
+    <> singleton (MustIncludeDatum dt)
+
+-- | Create the given value. FIX ME: Broken until unitRedeemer properly defined.
+mustMintValue :: forall (i :: Type) (o :: Type). Value -> TxConstraints i o
+mustMintValue = mustMintValueWithRedeemer (unitRedeemer Mint)
+
+-- Datatype doesn't match Plutus because of currencyMPSHash in Maybe context.
```
adamczykm, line 211: This comment is unclear to me

src/Types/TxConstraints.purs
```diff
@@ -0,0 +1,370 @@
...
+import Types.RedeemerTag (RedeemerTag(Mint))
+import Types.ScriptHash (MintingPolicyHash, ValidatorHash)
+import Types.Transaction (Redeemer)
+import Types.UnbalancedTransaction (PubKeyHash, TxOutRef)
+import Types.Value (TokenName, Value, currencyMPSHash, getNonAdaAsset, isZero)
+
+--------------------------------------------------------------------------------
+-- TxConstraints Type and related
+--------------------------------------------------------------------------------
+-- Taken from https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-constraints/html/Ledger-Constraints.html
```
adamczykm, line 66: I think it may be of value to note the version of the code that was translated.

adamczykm (commented):
Some minor comments to address, otherwise lgtm

src/Types/TxConstraints.purs
vvtran: I can remove these tomorrow.

src/Types/TxConstraints.purs
vvtran: Ah, forgot to delete, thanks.

src/Types/POSIXTimeRange.purs
```diff
@@ -48,19 +100,177 @@ instance Show a => Show (UpperBound a) where
 -- | that the endpoints may or may not be included in the interval.
 -- |
 -- | The interval can also be unbounded on either side.
-newtype Interval a = Interval { from :: LowerBound a, to :: UpperBound a }
+newtype Interval a = Interval { ivFrom :: LowerBound a, ivTo :: UpperBound a }
```
ngua, line 103: Why the change here?

src/Types/POSIXTimeRange.purs
```diff
@@ -48,19 +100,177 @@ instance Show a => Show (UpperBound a) where
...
+mkInterval f t = Interval { ivFrom: f, ivTo: t }
+
+strictUpperBound :: forall (a :: Type). a -> UpperBound a
+strictUpperBound a = UpperBound (Finite a) false
+
+strictLowerBound :: forall (a :: Type). a -> LowerBound a
+strictLowerBound a = LowerBound (Finite a) false
+
+lowerBound :: forall (a :: Type). a -> LowerBound a
+lowerBound a = LowerBound (Finite a) false
```
ngua, line 156: This should be `true`, no?

src/Types/TxConstraints.purs
```diff
@@ -0,0 +1,370 @@
...
+  | MustSatisfyAnyOf (Array (Array TxConstraint))
+
+derive instance eqTxConstraint :: Eq TxConstraint
+derive instance genericTxConstraint :: Generic TxConstraint _
+
+instance showTxConstraint :: Show TxConstraint where
+  show x = genericShow x
+
+newtype InputConstraint (i :: Type) = InputConstraint
+  { icRedeemer :: i
```
ngua, line 90: I'd prefer not to follow this record field naming convention (also applies to several records below)

src/Types/POSIXTimeRange.purs
vvtran: I do prefer `from` and `to` but it's to help people when they copy & paste - less changes. Happy to change back if you want.

src/Types/POSIXTimeRange.purs
vvtran: Good spot. Thanks.

src/Types/TxConstraints.purs
vvtran: I guess users won't be digging into these so it makes sense to name them better.

src/Types/TxConstraints.purs
vvtran: Hang on, it looks like it's 1D on the latest version on `master` which what I was following, but `staging` is 2D (or whatever they use for https://playground.plutus.iohkdev.io/doc/haddock/index.html), maybe that is what confused me. https://github.com/input-output-hk/plutus/tree/ead3960ad6f70a7b3f88bcc1bc53eca756746f2a I am using this and running haddock on localhost.

vvtran: Screenshot from 2022-02-24 08-15-35

vvtran: Oh wait, I think I may be using an out of date build. Perhaps it is 2d after all.

ngua (commented):
Forgot to mention this earlier: there are several places where doc comments use Haddock syntax (e.g. `@...@`) or have missing continuation pipes

src/Types/TxConstraints.purs
vvtran: Okay, turns out I need to update `MustPayToPubKey` to `MustPayToPubKeyAddress` too.

vvtran:
Applied a bunch of changes suggested above. Also used a more up-to-date `Plutus` repo so there were a few functions I was missing. The Plutus/Plutus-App rev has been commented. Tried to clean up most of the comments in the repo by adding backticks where I felt suitable (and removing remnants of Haskell syntax).

adamczykm (approved):
:+1:

ngua (approved):
LGTM

PR #142, opened 2022-02-23
Fix to `mkSingletonValue`, `mkSingletonValue'` and `adaToken` now exportable

vvtran:
With the new `mkTokenName`, I think it's now safe to export `adaToken`. In particular, we can relax some conditions on `mkSingletonValue'` and `mkSingletonValue`.

### DIFF
--- a/src/Types/Value.purs
+++ b/src/Types/Value.purs
@@ -4,6 +4,7 @@ module Types.Value
   , NonAdaAsset(..)
   , TokenName
   , Value(..)
+  , adaToken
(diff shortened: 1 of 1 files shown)


klntsky (approved):

ngua (approved):
I still don't think we should export `adaToken` -- not because it's unsafe, but just because its name might be confusing for users who will already need to adjust to a new representation of `Value`. I don't want to hold up this PR though, especially if everyone thinks it's OK to export it

PR #146, opened 2022-02-25
Attach datums to transactions

### LINKED ISSUE #126: Attach datums to transaction
This might be trickier than it seems: https://github.com/Emurgo/cardano-serialization-lib/issues/281

Bear in mind that Emurgo's CSL and Sundae's fork (now our fork) might have diverged w.r.t this

ngua:
Closes #126 

A few other changes as well:

- ~The `Semigroup` instance for `TransactionWitnessSet` now uses `Last` for each of its fields. This is because the previous implementation (`lift2 union`) would lead to losing witnesses when appending in common conditions. For example, combining a witness set containing only vkeys with one containing only datums would produce an empty witness set under the previous implementation~. This has been updated, see discussions below
- Using Nami's `signTx` will now combine the new witness set with the existing witness set in the transaction to avoid losing any datums which may be present


ngua:
Here is another possible `Semigroup` instance for the fields of `TransactionWitnessSet`:

```purescript
appendOrLast :: forall (a :: Type). Maybe (Array a) -> Maybe (Array a) -> Maybe (Array a)
appendOrLast (Just xs) (Just ys) = Just $ xs <> ys
appendOrLast xs ys = let Last z = ((<>) `on` Last) xs ys in z

```

Thoughts?

src/Serialization/WitnessSet.purs
```diff
@@ -2,23 +2,42 @@ module Serialization.WitnessSet where
...
+  )
 import Types.Aliases (Bech32String)
-
+import Types.ByteArray (ByteArray)
 import Types.RedeemerTag as Tag
+import Types.Transaction as T
+
+setPlutusData :: PlutusData -> TransactionWitnessSet -> Effect TransactionWitnessSet
+setPlutusData pd ws = _wsSetPlutusData containerHelper ws (Array.singleton pd)
+  *> pure ws
```
klntsky, line 40: Why an argument is returned here?

klntsky: I think that not returning mutable values out of functions mutating them highlights that the values are mutable references of values, not pure data. If the caller made this call, they must have the witness set in scope, so no need to return.

src/Transaction.purs
```diff
@@ -0,0 +1,44 @@
...
+-- | Fails if either the datum or updated witness set cannot be converted during
+-- | (de-)serialization
+attachDatum :: Datum -> Transaction -> Effect (Either ModifyTxError Transaction)
+attachDatum (Datum pd) (Transaction tx@{ witness_set: ws }) = runExceptT $ do
+  pd' <- liftEither
+    $ note ConvertDatumError
+    $ Serialization.PlutusData.convertPlutusData pd
+  newWits <- ExceptT $
+    map
+      ( note ConvertDatumError
```
klntsky, line 39: `ConvertWitsError`?

src/Serialization/WitnessSet.purs
ngua: Fair enough, I'll change that

src/Transaction.purs
ngua: Good catch

adamczykm:
> Here is another possible `Semigroup` instance for the fields of `TransactionWitnessSet`:
> 
> ```purescript
> appendOrLast :: forall (a :: Type). Maybe (Array a) -> Maybe (Array a) -> Maybe (Array a)
> appendOrLast (Just xs) (Just ys) = Just $ xs <> ys
> appendOrLast xs ys = let Last z = ((<>) `on` Last) xs ys in z
> ```
> 
> Thoughts?

Why not:
```purescript
maybeArrayMerge :: forall a. Eq a => Maybe (Array a) -> Maybe (Array a) -> Maybe (Array a)
maybeArrayMerge Nothing y = y
maybeArrayMerge x Nothing = x
maybeArrayMerge (Just x) (Just y) = Just $ union x y
```

src/Transaction.purs
```diff
@@ -0,0 +1,44 @@
...
+import Effect (Effect)
+import Helpers (liftEither)
+import Serialization.WitnessSet as Serialization.WitnessSet
+import Deserialization.WitnessSet as Deserialization.WitnessSet
+import Serialization.PlutusData as Serialization.PlutusData
+import Types.PlutusData (Datum(Datum))
+import Types.Transaction (Transaction(Transaction))
+
+data ModifyTxError
+  = ConvertWitsError
```
adamczykm, line 21: Wits -> Witns? It ConvertWitsError somehow doesn't ring "Witness(es)" in my head. If it suits the others I'll be fine.

src/Types/Transaction.purs
```diff
@@ -273,12 +273,12 @@ instance Show TransactionWitnessSet where
...
-      , bootstraps: lift2 union tws.bootstraps tws'.bootstraps
-      , plutus_scripts: lift2 union tws.plutus_scripts tws'.plutus_scripts
-      , plutus_data: lift2 union tws.plutus_data tws'.plutus_data
-      , redeemers: lift2 union tws.redeemers tws'.redeemers
+      { vkeys: tws.vkeys <<>> tws'.vkeys
+      , native_scripts: tws.native_scripts <<>> tws'.native_scripts
+      , bootstraps: tws.bootstraps <<>> tws'.bootstraps
+      , plutus_scripts: tws.plutus_scripts <<>> tws'.plutus_scripts
+      , plutus_data: tws.plutus_data <<>> tws'.plutus_data
+      , redeemers: tws.redeemers <<>> tws'.redeemers
```
adamczykm, line 281: Current `<<>>` implementation loses data. See my other comment

test/Transaction.purs
```diff
@@ -0,0 +1,45 @@
...
+  , TransactionWitnessSet(TransactionWitnessSet)
+  )
+import Types.PlutusData (Datum(Datum), PlutusData(Integer))
+
+suite :: TestPlanM Unit
+suite = group "attach datums to tx"
+  $ test "datum should be correctly attached" testAttachDatum
+
+testAttachDatum :: Aff Unit
+testAttachDatum = liftEffect $
```
adamczykm, line 29: We could use a test which shows that all witnesses are preserved

adamczykm (changes requested):
.

src/Transaction.purs
ngua: I think I've seen "wits" as an abbreviation for witnesses elsewhere. I'll just put `ConvertWitnessesError` though, it's clearer

test/Transaction.purs
ngua: Good idea

vvtran (changes requested):
Some comments, otherwise LGTM.

src/Transaction.purs
```diff
@@ -0,0 +1,42 @@
...
+-- | Attach a `Datum` to a transaction by modifying its existing witness set.
+-- | Fails if either the datum or updated witness set cannot be converted during
+-- | (de-)serialization
+attachDatum :: Datum -> Transaction -> Effect (Either ModifyTxError Transaction)
+attachDatum (Datum pd) (Transaction tx@{ witness_set: ws }) = runExceptT $ do
+  pd' <- liftEither
+    $ note ConvertDatumError
+    $ Serialization.PlutusData.convertPlutusData pd
+  ws' <- liftEffect $ Serialization.WitnessSet.convertWitnessSet ws
+  void $ liftEffect $ Serialization.WitnessSet.setPlutusData pd' ws'
```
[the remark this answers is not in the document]
ngua: Hmmm. Could've sworn I still got a warning, which surprised me. Maybe I confused myself, not getting it now

src/Helpers.purs
```diff
@@ -33,3 +36,10 @@ fromJustEff :: forall (a :: Type). String -> Maybe a -> Effect a
 fromJustEff e = case _ of
   Nothing -> throw e
   Just x -> pure x
+
+liftEither
+  :: forall (a :: Type) (e :: Type) (m :: Type -> Type)
+   . MonadError e m
+  => Either e a
+  -> m a
+liftEither = either throwError pure
```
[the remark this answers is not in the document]
ngua: It's based off the same effect from `mtl`. We could just use `except`, but I see no need to remove it now especially if we switch to a `MonadError`-based approach to errors eventually

klntsky (approved):

adamczykm (approved):
Lgtm

vvtran (approved):
lgtm

PR #152, opened 2022-02-27
Make ogmios datum cache available in nix shell

adamczykm:


ngua (changes requested):
`ogmios-datum-cache` is already available in the shell (see #137). I didn't add a Makefile target for it since the command is so short already. I'm OK with keeping the Makefile changes here, but the changes to the `shell.nix` aren't necessary (and also break reproducibility)

adamczykm:
I wonder how I missed that :sweat_smile:

Issue #135, opened 2022-02-21
Add `ogmios-datum-cache` WS connection to `QueryConfig`


ngua:

ngua:
Closed by #164

Issue #113, opened 2022-02-16
Add `getDatumByHash` to `QueryM` interface


ngua:
`getDatumByHash :: DataHash -> QueryM (Maybe Data)` or something similar

Currently blocked on #112, #39

ngua:
Also blocked by #135

ngua:
Closed by #164

ngua:
Closed by #164

PR #159, opened 2022-03-03
Add datum cache websocket, queries and serialization functions.

adamczykm:
- Only some queries are supported at the moment (will implement during the review)
- `ogmios-datum-cache` does not support mapping ids to messages - proposed solution relies on the current server and `ws` library implementation ('requests' are handled in fifo order and 'responses' order is preserved).
- Waiting for PlutusData deserialisation
- No tests


src/DatumCacheWsp.purs
```diff
@@ -0,0 +1,127 @@
...
+
+
+deepLookup :: Array String -> Json -> Either JsonDecodeError Json
+deepLookup keys obj = note (UnexpectedValue obj) $ foldM lookup obj keys
+  where
+    lookup :: Json -> String -> Maybe Json
+    lookup j lbl = caseJsonObject Nothing (FO.lookup lbl) j
+
+
+-- TODO
```
ngua, line 121: Are these TODOs for the future, or are you planning on implementing this before the PR is merged? I opened another issue for the ODC control API

src/DatumCacheWsp.purs
```diff
@@ -0,0 +1,127 @@
...
+  maybe
+  (Left $ maybe invalidResponseError WspFault fault)
+  decodeResponse
+  result
+  where
+    decodeResponse r = case methodname of
+      "GetDatumByHash" -> GetDatumByHashResponse <$>
+        let
+          datumFound =
+            Just <$> liftErr (decodeJson =<< deepLookup ["DatumFound", "value"] r)
```
ngua, line 85: We can't use `decodeJson`, right? We need to use the new decoder from #150, don't we?

src/QueryM.purs
```diff
@@ -114,20 +132,61 @@ utxosAt' addr = do
...
-allowError :: (Either Error JsonWsp.UtxoQR -> Effect Unit) -> JsonWsp.UtxoQR -> Effect Unit
+--------------------------------------------------------------------------------
+-- Datum Cache Queries
+--------------------------------------------------------------------------------
+
+getDatumByHash :: DatumHash -> QueryM (Maybe PlutusData)
+getDatumByHash hash = do
+  queryDatumCache (GetDatumByHashRequest hash) >>= case _ of
+    GetDatumByHashResponse mData -> pure $ mData
+    _ -> liftEffect $ throw "Request-response type mismatch. Should not have happened"
```
ngua, line 156: Since this effect is already returning a `Maybe PlutusData`, why not return a `Nothing` here since the caller already needs to deal with the `Maybe`?

src/Types/PlutusData.purs
```diff
@@ -34,6 +35,9 @@ derive instance Eq PlutusData
 derive instance Ord PlutusData
 derive instance Generic PlutusData _
 
+instance DecodeJson PlutusData where
```
ngua, line 38: I don't think we can use a `DecodeJson` instance for `PlutusData`, can we?

src/QueryM.purs
```diff
@@ -376,11 +469,34 @@ utxoQueryDispatch ref str = do
...
 
+-- we parse out the utxo query result, then check if we're expecting a result
+-- with the provided id, if we are then we dispatch to the effect that is
+-- waiting on this result
+datumCacheQueryDispatch
+  :: Ref.Ref (MultiMap String (DcWsp.JsonWspResponse -> Effect Unit))
+  -> String
+  -> Effect (Either Json.JsonDecodeError (Effect Unit))
+datumCacheQueryDispatch dim str = do
+  let parsed' = decodeJson =<< Helpers.parseJsonStringifyNumbers str
```
ngua, line 485: Why not merge with @klntsky's datum decoding branch and use the new decoding approach?

ngua (commented):
It seems like this is a WIP still, but I've left some comments. I thought you were only going to take care of #135  but I'm glad to see this will complete #113 as well

Here are some more general thoughts: 

- I don't think we can directly use Argonaut for decoding the response
- Are you planning on writing a test similar to the `AffInterface` one?

src/DatumCacheWsp.purs
adamczykm: Yes, once it's merged I will use it.

src/Types/PlutusData.purs
adamczykm: Will use aeson once it's merged

src/QueryM.purs
adamczykm: Will do, since the other pr is very close to being merged I decided to wait couple of hours for it

src/DatumCacheWsp.purs
adamczykm: I'm fine with both approaches. What do you suggest?

src/MultiMap.purs
```diff
@@ -0,0 +1,36 @@
...
+import Control.Bind ((>>=))
+import Control.Category ((>>>))
+import Data.Array as Array
+import Data.Bounded (class Ord)
+import Data.Function (($))
+import Data.Map as Map
+import Data.Maybe (Maybe)
+import Data.Monoid ((<>))
+
+newtype MultiMap k v = MultiMap (Map.Map k (Array v))
```
vvtran, line 21: Bunch of kind signatures in this module.

src/MultiMap.purs
```diff
@@ -0,0 +1,36 @@
...
+singleton k v = MultiMap $ Map.insert k (Array.singleton v) Map.empty
+
+insert :: forall k v. Ord k => k -> v -> MultiMap k v -> MultiMap k v
+insert k v (MultiMap m)= MultiMap $ Map.insertWith (<>) k (Array.singleton v) m
+
+delete :: forall k v. Ord k => k -> MultiMap k v -> MultiMap k v
+delete k (MultiMap m)= MultiMap $ Map.update (Array.uncons >>> map _.tail) k m
+
+lookup :: forall k v. Ord k => k -> MultiMap k v -> Maybe v
+lookup k (MultiMap m) = Map.lookup k m >>= Array.head
```
vvtran, line 36: Is it worth adding a few comments for `insert`, `delete` and `lookup` that this is indeed FIFO where the array head is the front so the reader doesn't need to worry about the details.

vvtran (commented):
Lgtm, few comments.

src/Serialization/Address.purs
```diff
@@ -101,14 +104,32 @@ derive newtype instance Eq Slot
...
+derive instance Newtype BlockId _
+derive instance Generic BlockId _
+
+instance EncodeJson BlockId where
+  encodeJson = undefined -- TODO
+
+instance Show BlockId where
+  show = genericShow
+
+-- instance Semigroup Slot where
```
ngua, line 128: Why remove/comment out the `Semigroup` and `Monoid` instances?

src/DatumCacheWsp.purs
```diff
@@ -0,0 +1,220 @@
...
+  | StartFetchBlocks
+  | CancelFetchBlocks
+  | DatumFilterAddHashes
+  | DatumFilterRemoveHashes
+  | DatumFilterSetHashes
+  | DatumFilterGetHashes
+
+derive instance Eq DatumCacheMethod
+
+datumCacheMethodToString :: DatumCacheMethod -> String
```
ngua, line 76: Why not use a `Show` instance with `genericShow` for this?

src/QueryM.purs
```diff
@@ -71,23 +126,23 @@ import Wallet (Wallet(Nami), NamiWallet, NamiConnection)
...
+foreign import _wsClose :: JsWebSocket -> Effect Unit
 
-foreign import _stringify :: forall a. a -> Effect String
+foreign import _stringify :: forall (a :: Type). a -> Effect String
 
-foreign import _wsWatch :: WebSocket -> Effect Unit -> Effect Unit
+foreign import _wsWatch :: JsWebSocket -> Effect Unit -> Effect Unit
 
-data WebSocket
+data JsWebSocket
```
ngua, line 145: Since you're making a change here already, I think it might make sense to turn this into a `foreign import data` since we use that everywhere else

ngua (commented):
I merged #150 into master so you can use the new decoder

src/MultiMap.purs
```diff
@@ -0,0 +1,43 @@
...
+import Data.Array as Array
+import Data.Bounded (class Ord)
+import Data.Function (($))
+import Data.Map as Map
+import Data.Maybe (Maybe)
+import Data.Monoid ((<>))
+
+-- | A specific-use MultiMap where values for the same key are stored in FIFO container
+--   implemented via Array
+newtype MultiMap (k :: Type) (v :: Type) = MultiMap (Map.Map k (Array v))
```
klntsky, line 22: `List` could be used instead of array for faster inserts. `insert` should append to the front of the list.

klntsky (approved):

src/Serialization/Address.purs
adamczykm: Forgot to uncomment.

src/DatumCacheWsp.purs
adamczykm: I'd have to restructure fromString to match on genericShow result. I can do it, but I find this approach just more straightforward.

src/MultiMap.purs
adamczykm: I've run some benchmarks it has the same performance in `insert` and some operations will be faster with `Array`

ngua:
Superseded by #164

PR #158, opened 2022-03-01
Attach redeemer to transaction

ngua:
Closes #144 

This is very similar to #146

PR #162, opened 2022-03-04
Attach scripts and script data hash

### LINKED ISSUE #145: Attach Plutus scripts and script hash to tx

ngua:
Closes #145 

The script witness part is very similar to #146 and #158. This also takes care of setting the script data hash in the tx body


vvtran (approved):
lgtm

PR #170, opened 2022-03-09
Implement ToData and FromData for basic types

### LINKED ISSUE #147: Define `unitRedeemer` and `unitDatum`
`unitRedeemer :: RedeemerTag -> Redeemer` currently in `Types.PlutusData` should be defined as we may need an equivalent to https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger/html/Ledger.html#v:unitRedeemer in Haskell, defined as 
```
-- | @()@ as a datum.
unitDatum :: Datum
unitDatum = Datum $ toBuiltinData ()

-- | @()@ as a redeemer.
unitRedeemer :: Redeemer
unitRedeemer = Redeemer $ toBuiltinData ()
```
where our `toBuiltinData` is `toData` from `class ToData`. Our `Redeemer` type has a few other fields like `index` and `ex_units` which would somehow need filling. These are used in the `Constraints` API when the user wants to mint tokens with a unit redeemer for example.

`unitDatum` should be easier once we can convert to `unit` to `PlutusData`.

klntsky:
Also add `unitRedeemer` and `unitDatum`.

Closes #147


src/Types/Redeemer.purs
```diff
@@ -0,0 +1,18 @@
...
+  ) where
+
+import Types.PlutusData (PlutusData(..))
+import Types.RedeemerTag (RedeemerTag)
+import Types.Transaction (Redeemer(..))
+import Data.BigInt as BigInt
+
+-- FIX ME (create an issue unless someone notices simple solution to this in PR
+-- review)
+unitRedeemer :: RedeemerTag -> Redeemer
```
ngua, line 12: I don't think this is what we want. `unitRedeemer` from Plutus is just: `Redeemer $ toBuiltinData ()`. The only field we could possibly specify here for a `Types.Transaction.Redeemer` would be `data`. We can't know its index, ex units, and tag without knowing which input the redeemer belongs to

src/Types/Redeemer.purs
```diff
@@ -0,0 +1,18 @@
+module Types.Redeemer
+  ( unitRedeemer
+  ) where
+
+import Types.PlutusData (PlutusData(..))
+import Types.RedeemerTag (RedeemerTag)
+import Types.Transaction (Redeemer(..))
+import Data.BigInt as BigInt
+
+-- FIX ME (create an issue unless someone notices simple solution to this in PR
```
ngua, line 10: This comment can be deleted

src/Types/Redeemer.purs
```diff
@@ -0,0 +1,18 @@
+module Types.Redeemer
```
ngua, line 1: This module's name would suggest (to me at least) that it defines a `Redeemer` type, but it only defines and exports `unitRedeemer`. I think that the `unitReedemer` function (which is only for convenience) can just go in `Helpers` or something and this module can be deleted

ngua (changes requested):

src/Types/Redeemer.purs
vvtran: I think I refactored quite a few things, moving the datatype `Redeemer` under `Types.Datum` as they're closely related, feel free to reconcile with these changes in a manner you see fit.

src/Types/Redeemer.purs
vvtran: Yes, so `Redeemer $ toData unit` should be fine. Just to be clear @klntsky, Plutus has extra features for `BuiltInData` which is a wrapper over `Data` (our `PlutusData`) which we can ignore, we can essentially treat it as `toData`.

src/FromData.purs
```diff
@@ -0,0 +1,57 @@
...
+  fromData (List []) = Just unit
+  fromData _ = Nothing
+
+instance FromData Boolean where
+  fromData (Integer n)
+    | n == BigInt.fromInt 0 = Just false
+    | n == BigInt.fromInt 1 = Just true
+  fromData _ = Nothing
+
+instance Fail (Text "Int is not supported, use BigInt instead") => FromData Int where
```
vvtran, line 32: TIL this, cool stuff!

vvtran (commented):
Looks good other than @ngua's comment about the redeemer that needs addressing. The PR will need to reconcile with `master` which included a few refactors too.

ngua (approved):

PR #172, opened 2022-03-10
Viet/redeemer exunits and attach

vvtran:
https://github.com/Plutonomicon/cardano-browser-tx/issues/167


src/Types/ScriptLookups.purs
```diff
@@ -301,29 +289,23 @@ instance Semigroup ValueSpentBalances where
...
   , valueSpentBalancesOutputs :: ValueSpentBalances
   -- Balance of the values produced and required for the transaction's outputs
   , datums :: Array Datum
-  -- *Set* of accumulating datums so we can use `setScriptDataHash`
-  , redeemers :: Map RedeemerTag (Array Redeemer)
-  -- Map of accumulating redeemers separated by `RedeemerTag`. Note that adding
-  -- to this discards the ordering of redeemers from constraints (should that
-  -- matter).
+  -- Ordered accumulation of datums so we can use to `setScriptDataHash`
+  , redeemers :: Array (RedeemerTag /\ T.Redeemer)
```
ngua, line 298: I can't think of why we might need the `RedeemerTag`. Is there a specific future scenario you're thinking of? These are only being stored for `setScriptDataHash`, right? We always need to include all redeemers when calling that (AFAIK), so I don't think we'll need to distinguish between different redeemer types

src/PlutusData.purs
```diff
@@ -0,0 +1,18 @@
+module PlutusData
```
ngua, line 1: Could this go into `Helpers`? If so I personally think that would be better than a whole new module for this (unless there's some circular dependency issue with that too)

ngua (commented):
Looks good, a few comments below

src/Types/ScriptLookups.purs
vvtran: Sure, I can remove.

src/PlutusData.purs
vvtran: Will try add it in.

vvtran: Hmm, that caused some horrible cyclic deps.

vvtran: Going to leave it for now.

ngua: OK, no problem. There's no way around the extra module then

ngua (approved):
I _think_ this is the correct way to deal with the redeemer pointers. We will find out definitively when we start submitting smart contract transactions

PR #173, opened 2022-03-10
Remove old-style Aeson decoding

### LINKED ISSUE #151: Lossless JSON decoding in `utxoQueryDispatch`
https://github.com/Plutonomicon/cardano-browser-tx/pull/150 added another (better) way of decoding JSON, we need to use it everywhere consistently and drop the old code.

The Ask:

Replace `parseJsonStringifyNumbers` with `Aeson.decodeAesonString`. Remove `Helpers.purs`.

klntsky:
Closes #151


src/Aeson.purs
```diff
@@ -134,31 +159,47 @@ jsonToAeson = stringify >>> decodeJsonString >>> fromRight shouldNotHappen
...
-   . (Aeson -> Either JsonDecodeError a)
-  -> Object Aeson
-  -> String
-  -> Either JsonDecodeError a
-getField' decoder obj str =
-  maybe
-    (Left $ AtKey str MissingValue)
-    (lmap (AtKey str) <<< decoder)
-    (FO.lookup str obj)
+getFieldOptional :: forall a. DecodeAeson a => Object Aeson -> String -> Either JsonDecodeError (Maybe a)
```
ngua, line 189: Stylistic: here and below there are a large number of type variables without kind signatures. I think it's worth adding them for consistency with the rest of the codebase

test/Aeson.purs
```diff
@@ -15,18 +18,30 @@ import Aeson
...
-import Data.Argonaut (encodeJson, parseJson, stringify)
-import Data.Array (head)
+import Control.Apply (lift2)
+import Control.Monad.Cont (lift)
+import Data.Argonaut (encodeJson, parseJson)
+import Data.Argonaut as Json
+import Data.Array (head, length, (!!), zip)
 import Data.BigInt as BigInt
-import Data.Either (Either(Right))
+import Data.Either (Either(..))
```
ngua, line 30: I think we've also settled on using explicit constructor imports (there are a few more implicit imports below)

ngua (commented):
Is there any way we could modify or incorporate the `ArbitraryJson` that we had previously?

klntsky:
> Is there any way we could modify or incorporate the ArbitraryJson that we had previously?

Working on that. Just discovered a bug thanks to that!

src/Aeson.purs
```diff
@@ -134,31 +158,47 @@ jsonToAeson = stringify >>> decodeJsonString >>> fromRight shouldNotHappen
...
-   . (Aeson -> Either JsonDecodeError a)
-  -> Object Aeson
-  -> String
-  -> Either JsonDecodeError a
-getField' decoder obj str =
-  maybe
-    (Left $ AtKey str MissingValue)
-    (lmap (AtKey str) <<< decoder)
-    (FO.lookup str obj)
+getFieldOptional :: forall (a :: Type). DecodeAeson a => Object Aeson -> String -> Either JsonDecodeError (Maybe a)
```
adamczykm, line 188: I think we should provide both versions or at least add a clear docstring about this one just as Argonauts does it.

adamczykm (dismissed):

src/Aeson.js
```diff
@@ -28,15 +28,21 @@ exports.stringifyAeson_ = numberIndex => object => {
...
+            let keys = [];
             for (let key in object) {
                 if (object.hasOwnProperty(key)) {
-                    res += JSON.stringify(key);
-                    res += ':';
-                    go(object[key]);
+                    keys.push(key);
+                }
+            };
+            keys.sort(); // for stability of Eq instance
```
adamczykm, line 37: I wonder why do you prefer to have string rep comparison?

src/Aeson.purs
```diff
@@ -90,6 +101,12 @@ newtype AesonPatchedJson = AesonPatchedJson Json
 -- | A piece of JSON where all numbers are extracted into `NumberIndex`.
 newtype Aeson = Aeson { patchedJson :: AesonPatchedJson, numberIndex :: NumberIndex }
 
+instance Eq Aeson where
+  eq a b = stringifyAeson a == stringifyAeson b
```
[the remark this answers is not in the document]
klntsky: The order of the keys is not guaranteed, and thus the order of elements in the index is not guaranteed too. Actually it costed me a few hours to debug.

But `stringifyAeson` ensures the order of keys by sorting.

ngua (approved):

PR #176, opened 2022-03-11
Add UsedUtxoFilter module implementation.

adamczykm:
- [x] Implement cache
- [x] Add unit tests
- [x] Example usage in QueryM


test/Fixtures.purs
```diff
@@ -125,6 +127,71 @@ txOutputBinaryFixture1 :: String
...
+mkSampleTx startTx changes =
+  applyChanges startTx $ buildChanges startTx changes
+
+  where
+  buildChanges :: Transaction -> (SampleTxConfig -> SampleTxConfig) -> SampleTxConfig
+  buildChanges (Transaction { body: TxBody { inputs } }) mkChanges =
+    mkChanges { inputs }
+
+  applyChanges :: Transaction -> SampleTxConfig -> Transaction
+  applyChanges
```
ngua, line 149: Can't we just use a record update for this? Something like `over Transaction _ { inputs = newInputs }`

src/Helpers.purs
```diff
@@ -130,3 +135,20 @@ appendRightHashMap
   -> HashMap k v
   -> HashMap k v
 appendRightHashMap = HashMap.unionWith (flip const)
+
+-- | Ensures that given predicate does not hold for a value.
+never
```
ngua, line 140: Personally, I'm not sure if the name of this function communicates its meaning very clearly

ngua (commented):
Do we want to integrate these changes with other parts of `QueryM` (e.g. `utxosAt`)? Or do we only want to offer this to users as an optional tool they can use if they want to?

src/Helpers.purs
```diff
@@ -25,16 +27,19 @@ import Data.Argonaut
...
+import Data.List.Lazy as LL
 import Data.Map (Map)
-import Data.Map (unionWith) as Map
+import Data.Map as Map
 import Data.Maybe (Maybe(Just, Nothing), maybe)
 import Data.Maybe.First (First(First))
 import Data.Maybe.Last (Last(Last))
-import Data.Either (Either(Right), either)
+import Data.Tuple (uncurry)
+import Data.Typelevel.Undefined (undefined)
```
vvtran, line 42: Redundant import.

src/Helpers.purs
adamczykm: It feels nice to me. I've seen something similar on pursuit as well. Any propositions?

src/UsedTxOuts.purs
```diff
@@ -0,0 +1,114 @@
...
+
+    updateCache :: TxOutRefCache -> TxOutRefCache
+    updateCache cache = foldr
+      ( \{ transaction_id, index } ->
+          Map.alter (fromMaybe Set.empty >>> Set.insert index >>> Just) transaction_id
+      )
+      cache
+      txOutRefs
+  in
+    void $ ask >>= (unwrap >>> Ref.modify updateCache >>> liftEffect)
```
vvtran, line 67: Could use `modify_` and get rid of `void`.

src/UsedTxOuts.purs
```diff
@@ -0,0 +1,114 @@
...
+unlockTransactionInputs
+  :: forall (m :: Type -> Type)
+   . MonadAsk UsedTxOuts m
+  => MonadEffect m
+  => Transaction
+  -> m Unit
+unlockTransactionInputs tx =
+  let
+    txOutRefs :: Array { transaction_id :: TransactionHash, index :: UInt }
+    txOutRefs = unwrap <$> (unwrap (unwrap tx).body).inputs
```
vvtran, line 79: You could unify this into an unexported function as it's used twice.

src/UsedTxOuts.purs
```diff
@@ -0,0 +1,114 @@
...
+  let
+    updateCache :: TxOutRefCache -> TxOutRefCache
+    updateCache cache = foldr
+      ( \{ transaction_id, index } ->
+          Map.update (\old -> never Set.isEmpty $ Set.delete index old) transaction_id
+      )
+      cache
+      txOutRefs
+  in
+    void $ ask >>= (unwrap >>> Ref.modify updateCache >>> liftEffect)
```
vvtran, line 101: `modify_` here too.

vvtran (commented):
lgtm, few comments

src/UsedTxOuts.purs
```diff
@@ -0,0 +1,114 @@
...
+import Data.Unit (Unit)
+import Effect.Class (class MonadEffect, liftEffect)
+import Effect.Ref (Ref)
+import Effect.Ref as Ref
+import Helpers (never)
+import Types.Transaction (Transaction, TransactionHash)
+
+type TxOutRefCache = Map TransactionHash (Set UInt)
+
+-- | Stores stores TxOutRefs in a compact map.
```
klntsky, line 35: Wording?

test/UsedTxOuts.purs
```diff
@@ -0,0 +1,128 @@
...
+
+            -- unlock
+            unlockTransactionInputs tx
+            log "All TxOuts should be now unlocked"
+            anyTxOutsLocked (usedTxOutRefs <> unusedTxOutRefs) `shouldReturn` false
+        )
+
+    test "UsedTxOuts cache properly locks and unlock selected txouts" $ do
+      newUsedTxOuts >>= runReaderT
+        ( do
```
klntsky, line 88: No need for parentheses here. And above too. `do` behaves similarly to Haskell with `-XBlockArguments`.

test/UsedTxOuts.purs
```diff
@@ -0,0 +1,128 @@
...
+            -- starts empty
+            log "TxOuts should start unlocked"
+            anyTxOutsLocked (usedTxOutRefs <> unusedTxOutRefs) `shouldReturn` false
+
+            -- lock
+            lockTransactionInputs tx
+            log "All usedTxOuts should be now locked"
```
klntsky, line 95: Can we avoid logging to console in tests? It becomes annoying when there's a large suite. Would be ideal if there was a solution that outputs collected logs only in case of a test failure. However, since we already do console logging in a couple of places, feel free not to change anything.

src/UsedTxOuts.purs
```diff
@@ -0,0 +1,114 @@
...
+  => MonadEffect m
+  => Foldable t
+  => t { transaction_id :: TransactionHash, index :: UInt }
+  -> m Unit
+unlockTxOutRefs txOutRefs =
+  let
+    updateCache :: TxOutRefCache -> TxOutRefCache
+    updateCache cache = foldr
+      ( \{ transaction_id, index } ->
+          Map.update (\old -> never Set.isEmpty $ Set.delete index old) transaction_id
```
klntsky, line 96: I suggest removing `never` and rewriting with the use of `guard` of `if` to make it more readable.

src/Helpers.purs
```diff
@@ -130,3 +135,20 @@ appendRightHashMap
...
+never p a = if p a then Nothing else Just a
+
+filterMapWithKeyM
+  :: forall (m :: Type -> Type) (k :: Type) (v :: Type)
+   . Ord k
+  => Monad m
+  => (k -> v -> m Boolean)
+  -> Map k v
+  -> m (Map k v)
+filterMapWithKeyM p = map Map.fromFoldable <<< LL.filterM (uncurry p) <<< Map.toUnfoldable
```
klntsky, line 154: Just curious, why lazy list specifically?

test/UsedTxOuts.purs
adamczykm: because it is being run in a readert, I am not able to use test instead. Should we have an issue for these smart logging capabilities?

src/UsedTxOuts.purs
```diff
@@ -0,0 +1,114 @@
+module UsedTxOuts
```
klntsky, line 1: Please add more comments into the module header, describing its purpose.

test/UsedTxOuts.purs
klntsky: I think It's possible to deal with `ReaderT` in definitions of `anyTxOutsLocked` and `allTxOutsLocked` and pass `UsedTxOuts` to them as an extra argument.

adamczykm:
> Do we want to integrate these changes with other parts of `QueryM` (e.g. `utxosAt`)? Or do we only want to offer this to users as an optional tool they can use if they want to?

I'm not sure about it that's why I didn't implement it. Although I think we probably should because it is nontrivial to keep the valid state at all times. I.e. in case a transaction was rejected it should unlock its inputs. Also it should not be keeping txouts that are already included in the ledger.

test/Fixtures.purs
adamczykm: But then there's the Body newtype.

adamczykm: And it will be more painful while extending SampleTxConfig

src/Helpers.purs
adamczykm: I have followed the implementation of `filterMapWithKey`

test/UsedTxOuts.purs
adamczykm: Do you suggest providing both pure and monadask based API?

adamczykm: I'd be against passing newtyped ref for an update like this: `f :: UsedTxOuts -> Effect Unit`

src/UsedTxOuts.purs
adamczykm: ok

src/UsedTxOuts.purs
adamczykm: ok

test/Fixtures.purs
vvtran:
```
applyChanges :: Transaction -> SampleTxConfig -> Transaction
applyChanges tx@(Transaction { body }) { inputs: newInputs } =
  over Transaction _ { body = over TxBody _ { inputs = newInputs } body } tx
 ```
 or
 ```
 applyChanges :: Transaction -> SampleTxConfig -> Transaction
applyChanges (Transaction tx@{ body: TxBody txBody }) { inputs: newInputs } =
  Transaction tx { body = TxBody txBody { inputs = newInputs } }
  ```
 or I think the nicest, using a bunch of lenses from Types.Transaction

 ```applyChanges tx { inputs: newInputs } = tx # _body <<< _inputs .~ newInputs```

vvtran: If there's reason to keep it for generalising SampleTxConfig, then I reckon it's fine either way.

test/UsedTxOuts.purs
klntsky: > Do you suggest providing both pure and monadask based API?

No, I'm talking about changing just the tests. See line 63 of this file. `ReaderT` can be added there, and both functions can accept a pure value of `UsedTxOuts`.

NB: there's an instance of `MonadAsk` for `Function`, so we get both pure and monadic interface all in one.

adamczykm: Ad NB: At first sight but then we have the MonadEffect constraint -.-'

adamczykm: Also (TIL) moving ReaderT to encompass calls will not work - the changes get reverted when exiting.

ngua (approved):
I opened an issue for the integration with existing stuff (#183) so that can be addressed separately

PR #177, opened 2022-03-14
Typed Constraints

vvtran:
- https://github.com/Plutonomicon/cardano-browser-tx/issues/166
- Added typed constraints to script lookups using functional deps instead of associated types which aren't available in Purescript.
- Added a few easy `ToData` `FromData` instances.
- Updated script lookups to include these type constraints.


src/Types/Any.purs
```diff
@@ -0,0 +1,18 @@
...
+  ) where
+
+import Prelude
+import Data.Generic.Rep (class Generic)
+import Data.Show.Generic (genericShow)
+
+-- Not sure if this is necessary but keep for now.
+-- | For converting a subset of polymorphic types to that which will work with
+-- | `PlutusData`.
+data Any
```
ngua, line 12: Missing a kind signature

src/Types/Transaction.purs
```diff
@@ -70,6 +73,14 @@ instance monoidTransaction :: Monoid Transaction where
     , auxiliary_data: Nothing
     }
 
+-- Do we actually need this?
```
ngua, line 76: I really don't understand this instance or the one below. I can't imagine how it could be used sensibly

src/Types/TypedTxOut.purs
```diff
@@ -0,0 +1,289 @@
...
+  let address = typedValidatorAddress networkId typedVal
+  pure $ mkTypedTxOut' (wrap { address, amount, data_hash: pure dHash }) dt
+  where
+  mkTypedTxOut'
+    :: TransactionOutput
+    -> b -- Data
+    -> TypedTxOut a b
+  mkTypedTxOut' txOut dat = TypedTxOut { txOut, data: dat }
+
+data WrongOutTypeError = ExpectedScriptGotPubkey -- | ExpectedPubkeyGotScript
```
ngua, line 179: What's going on with the commented-out constructor at the end?

src/Types/TypedTxOut.purs
```diff
@@ -0,0 +1,289 @@
...
+
+derive instance Generic WrongOutTypeError _
+derive instance Eq WrongOutTypeError
+derive instance Ord WrongOutTypeError
+
+instance Show WrongOutTypeError where
+  show = genericShow
+
+-- | An error we can get while trying to type an existing transaction part.
+data ConnectionError
```
ngua, line 189: Is this name taken from Plutus? If so, I don't think we should follow it. Especially in the context of CTL, I think the name would be take to mean a failure in some networked code

src/Types/TypedTxOut.purs
```diff
@@ -0,0 +1,289 @@
...
+  -> Address
+  -> m (Either ConnectionError Unit)
+checkValidatorAddress networkId typedVal actualAddr = runExceptT do
+  let expectedAddr = typedValidatorAddress networkId typedVal
+  unless (expectedAddr == actualAddr)
+    $ throwError
+    $ WrongValidatorAddress expectedAddr actualAddr
+
+-- -- | Checks that the given redeemer script has the right type.
+-- checkRedeemer
```
ngua, line 222: Do we need this? Is this related to the note on `addOwnInput`?

src/Types/TypedTxOut.purs
```diff
@@ -0,0 +1,289 @@
+module Types.TypedTxOut
```
ngua, line 1: This whole module is to make the typed validators possible and there is no analog in Plutus since it's not necessary, right? If so, can we add a doc comment to the module explaining this?

src/Types/TypedValidator.purs
```diff
@@ -0,0 +1,149 @@
...
+typedValidatorScript = _.validator <<< unwrap
+
+-- | The minting policy that forwards all checks to the instance's validator.
+forwardingMintingPolicy
+  :: forall (a :: Type). TypedValidator a -> MintingPolicy
+forwardingMintingPolicy = _.forwardingMPS <<< unwrap
+
+-- We have a few functions, I'm not sure if we even need these for off chain
+-- code:
+-- -- Broken, see below (we need some notion of `applyCode`)
```
ngua, line 130: We will have a server call that applies `PlutusData` arguments to a `Script` soon, but that is the only type of `applyCode` that we will be able to support (and it will be in `QueryM`)

src/Types/Transaction.purs
vvtran: It's used in `wrapValidator` constraints, although I'm not convinced we need it ATM. I could comment this function out for now?

src/Types/TypedTxOut.purs
vvtran: There seems to be a separation of checking whether something is a PubKey vs Script which is used in the `TxIn` version of our `typeTxOutRef`. https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger/html/src/Ledger.Typed.Tx.html#PubKeyTxOut

I'm tempted to juse remove this datatype and add `ExpectedScriptGotPubkey` as an option to `ConnectionError`

src/Types/TypedTxOut.purs
vvtran: It is, I will rename.

src/Types/TypedTxOut.purs
vvtran: Yes, it's related to the fact they use `TxIn` in `addOwnInput` which adds a redeemer which we don't. I can remove if you want.

src/Types/TypedTxOut.purs
vvtran: https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger/html/src/Ledger.Typed.Tx.html This is the equivalent module. I can add more comments.

src/Types/Any.purs
vvtran: There's no polymorphic variable here and I don't think we have standalone kinds, do we, if that's what you are referring to?

vvtran: Just implemented some changes if you could check.

ngua (approved):

PR #182, opened 2022-03-15
Add Ord instance for Address

klntsky:


src/Serialization/Address.purs
```diff
@@ -168,6 +168,9 @@ showVia nm toShowable addr = "(" <> nm <> " " <> show (toShowable addr) <> ")"
 instance Eq Address where
   eq = eq `on` addressBytes
 
+instance Ord Address where
+  compare a b = addressBytes a `compare` addressBytes b
```
vvtran, line 172: Just to follow a common convention in our libraries, I think `compare = compare on addressBytes` would be better, and it's point free.

vvtran (commented):
One comment.

vvtran:
I think this PR is redundant now, as this change was included in https://github.com/Plutonomicon/cardano-browser-tx/pull/177 which has since been merged.

ngua:
Made redundant by #177

Issue #107, opened 2022-02-15
Choose a consistent error-handling approach


ngua:
Things have started to get a bit complex w.r.t error handling (e.g. "pyramid-of-doom"-style pattern matching on various `Maybe`s and `Either`s). We should pick a single strategy for dealing with this inside `QueryM`, which would allow us to simplify quite a bit

Two possible options:
- add `ExceptT` to the `QueryM` stack
- make `QueryM` an instance of `MonadThrow` (after newtyping it) and throw/catch impure exceptions as necessary

ngua:
I am personally partial to option two -- but this is probably because I am used to this strategy in Haskell where `ExceptT` prevents one from using a lot of nice concurrency features

I would be fine with either of the two solutions above though. Any other ideas?

klntsky:
@ngua do you mean `MonadError`? 

I'd vote for 2, since that would allow us to handle underlying `Aff` errors and our domain errors uniformly.

adamczykm:
> I am used to this strategy in Haskell where `ExceptT` prevents one from using a lot of nice concurrency features
Can we expand on this subject? Do you have any reference materials? I haven't worked with ExpectT and concurrency yet and would like to avoid any pains in the future :)

adamczykm:
What do you think about this library?
https://pursuit.purescript.org/packages/purescript-checked-exceptions/1.0.0

ngua:
We should take care of this soon. The various `Maybe`s, `Either`s, etc... are inconveniencing library users

PR #185, opened 2022-03-16
`Contract` Monad, custom `Prelude` and  `FromData`/`ToData` instances

vvtran:
- https://github.com/Plutonomicon/cardano-browser-tx/issues/178
- https://github.com/Plutonomicon/cardano-browser-tx/issues/154
- `Address` (and related addresses), `Value`, `TransactionInput`, `TransactionOutput`, `Maybe`, `Either` and others `FromData`/`ToData` instances.
- Added `minus` to `Natural` as it's quite widely used, with infix operator `^-` inline with the `AdditiveHemigroup` which we don't have. Furthermore, `Natural` now has a `Semiring` instance which should be helpful.
- I added some comments on `FromData` and `ToData` of `Ratio`, our current implementation accepts zero denominator, which we may need to provide safety for, either for the instances or construction.
- Created a type called `NatRatio` as a newtype over `Rational` inline with `plutus-extra`, which smart constructors.
- Various helpers in `Contract.Monad` for throwing and running the contract.
- `mconcat` + `Helpers` inside `Contract.Prelude` 
- `ScriptLookups` now has a `Maybe` and non-`Maybe` context version to help with those using `<>` or `foldMap` etc, it means the user doesn't need to write `Just scriptlookup1 <> Just scriptlookup2  <> scriptlookup3`. Furthermore there are some `unsafe` versions.
- Improved some of the helpers in `Scripts`.
- Deleted `Serialization/Bech32` because it's unused.
- I originally made the change for the `Interval` fields  `from` -> `ivFrom` but have since reverted to `from`, because our API is quite divergent from Plutus - there's no point using their name if we have row polymorphism, and it's a minor change when converting off-chain code.


src/Contract/Prelude.purs
```diff
@@ -0,0 +1,302 @@
+-- | A custom Prelude that re-exports Purescript's prelude and further expands.
```
ngua, line 1: Maybe we could add an adaptation of `mconcat` since it's so common in off-chain code

src/Contract/Monad.purs
```diff
@@ -0,0 +1,77 @@
...
+derive newtype instance Monad Contract
+derive newtype instance MonadEffect Contract
+derive newtype instance Semigroup a => Semigroup (Contract a)
+derive newtype instance Monoid a => Monoid (Contract a)
+-- TO DO: can we define MonadError for `ContractError`? `Error` below come's from the `Aff`
+-- Monad, is it possible to derive a `ContractError` error type without
+-- using `Contract (ExceptT ContractError QueryM)`? Or would we rather `Error`
+-- to make use of Javascript native erroring.
+derive newtype instance MonadThrow Error Contract
+derive newtype instance MonadError Error Contract
```
ngua, line 60: I think this is OK for now. If we do something like `ContractError`, it will be really restrictive for users who will want to use their own types. Once we take care of #107, we can revisit this

src/Contract/PlutusData.purs
```diff
@@ -0,0 +1,84 @@
+-- | This module that defines query functionality via Ogmios to get `PlutusData`
```
ngua, line 1: Should we export `To`/`FromData` from here?

vvtran:
Ah, I forgot to newtype the `QueryConfig`, will do this first thing tomorrow.

vvtran:
Also, before I forget, I think I will export `Aff` and `Effect` from our prelude (without the helpers), in the same way Haskell has `IO` automatically.

src/Contract/Monad.purs
```diff
@@ -0,0 +1,72 @@
...
+derive newtype instance Monoid a => Monoid (Contract a)
+-- Utilise JavaScript's native `Error` via underlying `Aff` for flexibility:
+derive newtype instance MonadThrow Error Contract
+derive newtype instance MonadError Error Contract
+derive newtype instance MonadAsk QueryM.QueryConfig Contract
+derive newtype instance MonadReader QueryM.QueryConfig Contract
+derive newtype instance MonadRec Contract
+
+-- | Throws an `Error` for any showable error.
+throwError :: forall (e :: Type). Show e => e -> Contract e
```
ngua, line 63: Should we maybe rename this since it shadows the `MonadThrow` effect? I'm not sure how important that is. We could call it `throwContractError` for clarity perhaps?

src/Contract/Transaction.purs
```diff
@@ -0,0 +1,153 @@
...
+
+import Prelude
+import BalanceTx (balanceTx) as BalanceTx
+import BalanceTx (BalanceTxError) as BalanceTxError
+import Contract.Monad (Contract)
+import Data.Either (Either, hush)
+import Data.Maybe (Maybe)
+import Data.Newtype (wrap)
+import QueryM (signTransaction, submitTransaction) as QueryM
+import Types.JsonWsp (OgmiosTxOut, OgmiosTxOutRef) as JsonWsp --  Perhaps it's best not to expose `JsonWsp`.
```
ngua, line 24: I agree with the comment

src/Contract/Transaction.purs
```diff
@@ -0,0 +1,153 @@
...
+      , StakeDeregistration
+      , StakeDelegation
+      , PoolRegistration
+      , PoolRetirement
+      , GenesisKeyDelegation
+      , MoveInstantaneousRewardsCert
+      )
+  , CostModel(CostModel)
+  , Costmdls(Costmdls)
+  -- , DataHash(DataHash)
```
ngua, line 45: Why is this commented out?

src/Serialization/Address.purs
```diff
@@ -171,6 +174,16 @@ instance Eq Address where
 instance Ord Address where
   compare = compare `on` addressBytes
 
+-- Plutus uses `PlutusTx.makeIsDataIndexed ''Address [('Address,0)]` on their
```
ngua, line 177: You could add the issue URL here

src/Contract/Monad.purs
```diff
@@ -0,0 +1,72 @@
+-- | A module defining the `Contract` monad.
```
ngua, line 1: Can we define a `defaultQueryConfig` here (it needs to be effectful since it initializes websockets, in `Aff` not `Contract`)?

ngua (commented):
I think we also need `Contract.Wallet`

src/Contract/Scripts.purs
```diff
@@ -0,0 +1,61 @@
...
+import Types.Scripts
+  ( MintingPolicy(MintingPolicy)
+  , MintingPolicyHash(MintingPolicyHash)
+  , PlutusScript(PlutusScript)
+  , StakeValidator(StakeValidator)
+  , StakeValidatorHash(StakeValidatorHash)
+  , Validator(Validator)
+  , ValidatorHash(ValidatorHash)
+  ) as TypesScripts
+import Types.TypedValidator
```
ngua, line 53: What about `DatumType` and `RedeemerType`

ngua (commented):
Forgot something from my previous review

src/Contract/ScriptLookups.purs
```diff
@@ -0,0 +1,30 @@
...
+-- | Otherwise, there will be just one version that can fail (because of
+-- | hashing)
+module Contract.ScriptLookups (module ScriptLookups) where
+
+import Types.ScriptLookups
+  ( MkUnbalancedTxError(..) -- A lot errors so will refrain from explicit names.
+  , ScriptLookups(ScriptLookups)
+  , generalise
+  , mintingPolicyM
+  , mkUnbalancedTx
```
ngua, line 18: Don't we need to wrap this in `Contract`

ngua: I mean `mkUnbalancedTx`

src/Contract/Monad.purs
ngua: Also I'm not sure how pleasant this would be to use in practice. The type signature for `MonadThrow`'s `throwError` is `forall a. e -> m a`, so at least we can get an `a` back from it. Here we get the `e`, which I'm not sure is what we want

src/Types/ScriptLookups.purs
```diff
@@ -2,15 +2,19 @@ module Types.ScriptLookups
   ( MkUnbalancedTxError(..)
```
ngua, line 2: I'm OK with the solution you have for this, but I wanted to ask what you think of also exporting some `unsafe...` versions of lookups that need to call `validatorHash`, etc...?

src/Contract/Transaction.purs
vvtran: Hmm, actually, because I followed the chain index convention, `unspentOutputs` is currently using `Map TxOutRef OgmiosTxOut` as input so we'd need to expose this for now. I did think about changing this recently, generally script lookups will need refining as mentioned before.

vvtran: https://github.com/Plutonomicon/cardano-browser-tx/issues/200

src/Contract/Transaction.purs
vvtran: I tried to keep things somewhat modular (although not always admittedly), so the commented out things are address or data related. Since it's part of the transaction, I can add them back in.

src/Contract/ScriptLookups.purs
vvtran: Ah, my bad, will do.

src/Contract/Prelude.purs
```diff
@@ -0,0 +1,324 @@
...
+  ) as Helpers
+import Undefined (undefined) as Undefined
+
+-- Imports for extra functions:
+import Prelude
+import Data.Foldable (class Foldable, foldr)
+
+mconcat
+  :: forall (f :: Type -> Type) (m :: Type). Foldable f => Monoid m => f m -> m
+mconcat = foldr (<>) mempty
```
klntsky, line 324: `mconcat = Data.Foldable.fold`

src/Types/NatRatio.purs
```diff
@@ -0,0 +1,78 @@
...
+import Prelude
+import Data.BigInt (BigInt)
+import Data.Maybe (Maybe(Just, Nothing))
+import Data.Ratio ((%), denominator, numerator) as R
+import Types.Natural (Natural, fromBigInt', toBigInt)
+import Types.Rational (Rational)
+
+-- | `NatRatio` is a newtype over `Rational` with smart constructors to create a
+-- | `NatRatio` safely. Therefore, the data constructor is not exported.
+newtype NatRatio = NatRatio Rational
```
klntsky, line 24: Can we remove Rational and do

```
newtype Rational = Rational (Ratio BigInt)
```

?

Rational isn't going to be used much, right?

src/FromData.purs
```diff
@@ -61,13 +85,20 @@ instance FromData ByteArray where
...
+-- we want more safety:
 instance (Ord a, EuclideanRing a, FromData a) => FromData (Ratio a) where
   fromData (List [ a, b ]) = reduce <$> fromData a <*> fromData b
   fromData _ = Nothing
 
 instance FromData PlutusData where
   fromData = Just
 
+-- | This covers `Bech32` which is just a type alias for `String`
+instance FromData String where
```
klntsky, line 98: I'm against this, the user might expect it to work with any string. We only have bytestring instance in plutus as far as I remember, let's keep it that way.

src/FromData.purs
```diff
@@ -16,22 +16,42 @@ import Data.Ratio (Ratio, reduce)
...
+
+instance FromData a => FromData (Maybe a) where
+  fromData (Constr n [ pd ]) = case fromData pd of
+    Just Nothing | n == one -> Just Nothing
+    Just (Just x) | n == zero -> Just (Just x) -- Just is zero-indexed by Plutus
+    _ -> Nothing
+  fromData _ = Nothing
+
+instance (FromData a, FromData b) => FromData (Either a b) where
+  fromData (Constr n [ pd ]) = case fromData pd of
```
klntsky, line 51: Does it call itself?

src/Contract/Prelude.purs
```diff
@@ -0,0 +1,324 @@
+-- | A custom Prelude that re-exports Purescript's prelude and further expands.
```
ngua, line 1: Maybe we can export `Generic` and `genericShow` as well

src/Types/NatRatio.purs
vvtran: I was told that it's used a lot, I'd need to investigate how the Plutus Rational is implemented and probably replicate that, regardless of zero denoms unfortunately. Also, I believe they have a type alias too.

src/FromData.purs
vvtran: Not sure what you mean? The `pd` is already a `PlutusData` by construction, so we just need to check where it is zero- or one-indexed, and whether it's a `Left` or `Right` inside the `Just` (where the `Just` comes from the `fromData` definition)

vvtran:
I'm going to export `unwrap`, `wrap` and `over` too.

src/Contract/Prelude.purs
vvtran: The output of `Data.Foldable.fold` is very much in Purescript style, although I suspect the Haskellers will a return value as opposed to `Fold { step :: a -> Fold a b, finish :: Unit -> b }`, unless I've misunderstood? @ngua any thoughts on this?

vvtran: Also, this is a left fold on a right associative operator `(<>)`? Why is why I used `foldr`. Hmm.

src/FromData.purs
vvtran: That's fair, I was a bit unsure on this TBH. This will affect many other things including `PaymentPubKey` but not `PaymentPubKeyHash` which *I think* is the important one.

src/Types/ScriptLookups.purs
vvtran: It ultimately boils to how unsafe 
```
exports._scriptHashFromBytesImpl = maybe => bytes => {
    return hashFromImpl(CardanoWasm.ScriptHash.from_bytes)(maybe)(bytes);
};
```
is. Is this related to freeing allocated CSL objects? I guess it would just error out if it doesn't work right so it should be okay?

src/Contract/Prelude.purs
ngua: No, I meant like Haskell's `mconcat`. I think we should return the monoidal value, not a `Fold`

vvtran:
Okay, tried to apply all suggestions
- Added `class Newtype` too to prelude.
- newtype derived `MonadAff` for `Contract` which should be helpful.
- Removed `FromData`/`ToData` for `String` and related, which affects `PublicKey` which I think is okay.

src/FromData.purs
vvtran: Removed BTW.

src/Types/NatRatio.purs
klntsky: Sorry, I meant the old `type Rational = Ratio BigInt` isn't going to be used because the intention is to use the safe newtype.

So let's remove the type synonym in order not to occupy a good name.

src/FromData.purs
klntsky: I believe this code is not going to work. I added a test and it failed.

Could you please take a look at https://github.com/Plutonomicon/cardano-browser-tx/tree/klntsky/either-fromdata-bug ?

Hopefully I'm not misunderstanding something.

klntsky (dismissed):
See https://github.com/Plutonomicon/cardano-browser-tx/pull/185#discussion_r830244712

src/FromData.purs
vvtran: @klntsky really good spot. I just realised I was overnesting the Just/Left, we just need one layer of Just to confirm a "successful" fromData. Where `one` or `zero` just represents Right or Left. Check out my latest push, this should work with your test hopefully.

vvtran: Also note I can't write
```
case fromData pd of
    Just e | n == one -> Just (Left e)
    Just x | n == zero -> Just (Right x)
 ```
 because this concretises the type of `a` to be the same as `b`, hence
 ```

instance (FromData a, FromData b) => FromData (Either a b) where
  fromData (Constr n [ pd ])
    | n == zero = maybe Nothing (Just <<< Left) (fromData pd)
    | n == one = maybe Nothing (Just <<< Right) (fromData pd)
  fromData _ = Nothing
 ```
 I also fixed `Maybe`, there's only one type here, so we don't need to worry about any type confusion

Comment (vvtran) [reply src/FromData.purs]: Here's `Maybe` for your convenience:
```
instance FromData a => FromData (Maybe a) where
  fromData (Constr n [ pd ]) = case fromData pd of
    Just _ | n == one -> Just Nothing
    Just x | n == zero -> Just (Just x) -- Just is zero-indexed by Plutus
    _ -> Nothing
  fromData _ = Nothing
  ```

vvtran: Just to be overly explicit, if your example, toData gives `Constr zero [ Integer one ]` and mapping the `Integer` in the array should give `Just one`, not `Just (Left one)` like I originally had.

src/ToData.purs
```diff
@@ -49,6 +67,8 @@ instance (ToData k, ToData v) => ToData (Map k v) where
     where
     entries = Map.toUnfoldable mp :: Array (k /\ v)
 
+-- Note that nothing prevents the denominator from being zero, we could provide
```
klntsky, line 70: It looks like the property of not having zero denominator is not an invariant of `Ratio`, so we shouldn't do that. (Why though? I don't know)

I think we can remove the comment.

src/FromData.purs
```diff
@@ -41,16 +41,15 @@ instance FromData Boolean where
 
 instance FromData a => FromData (Maybe a) where
   fromData (Constr n [ pd ]) = case fromData pd of
-    Just Nothing | n == one -> Just Nothing
-    Just (Just x) | n == zero -> Just (Just x) -- Just is zero-indexed by Plutus
+    Just _ | n == one -> Just Nothing
```
klntsky, line 44: Wait, you are still pattern match with `[ pd ]`, but in case of `Nothing` it's `[]`:

```
> toData (Nothing :: Maybe Integer)
Constr 1 []
```

klntsky (changes requested):

src/ToData.purs
vvtran: So it does seem like `Ratio` doesn't care about this property, and it's down to the user to do whatever they wish, in which case a `Newtype` as you previously suggested could be best. I think the main thing is to replicate whatever Plutus does

src/Types/NatRatio.purs
vvtran: So `NatRatio` is meant to be different to `Rational`, not a replacement, `NatRatio` is just the "non-negative version" so we do need both. Although, perhaps a `Newtype` does make sense for `Rational`. This issue is the default behaviour of Purescript `Ratio` may different to Haskell's `Ratio`, in which case a newtype would make sense to replicate Haskell's behaviour, and therefore Pltuus.

src/FromData.purs
vvtran: Ah yes, you're correct, my bad, I  just changed it to:
```
instance FromData a => FromData (Maybe a) where
  fromData (Constr n [ pd ])
    | n == zero = maybe Nothing (Just <<< Just) (fromData pd) -- Just is zero-indexed by Plutus
  fromData (Constr n [])
    | n == one = Just Nothing
  fromData _ = Nothing  ```

File: src/ToData.purs (hunk 17 above)
Comment (vvtran) [reply src/ToData.purs:70]: Happy to leave this for a separate PR btw. Just to get this merged in asap.

File: src/Contract/Monad.purs [hunk 19]
```diff
@@ -0,0 +1,137 @@
...
+  pure $ ContractConfig
+    { ogmiosWs
+    , datumCacheWs
+    , wallet
+    , serverConfig: QueryM.defaultServerConfig
+    , usedTxOuts
+    }
+
+-- | Same as `defaultContractConfigAff` but lifted into `Contract`.
+defaultContractConfig :: Contract ContractConfig
```
ngua, line 135: I'm not sure how useful this will be, since we need a `ContractConfig` to run any `Contract` action. But I suppose we can keep it (maybe it would be useful with `local` or something?)

ngua (approved):

src/Contract/Monad.purs
vvtran: It could be useful for local, although I wonder if `defaultContractConfig` is a better name for the `Aff` version since that's what everyone is using. Then call the `Contract` version `defaultContractConfigLifted` perhaps.

ngua: I think that's the best/clearest way to name them

Issue #189, opened 2022-03-17
`applyCode` implementation


ngua:
The partially and fully applied scripts obtained from the `apply-args` endpoint don't match the CBOR hex provided by @t4ccer. The endpoint does produce valid scripts, however, in that they serialize and deserialize correctly. Possible reasons:
- `applyArguments` produces different results than `fromCompiledCode . applyCode` 
- different Plutus versions?

ssorc3:
`applyArguments` does produce different results from `fromCompiledCode . applyCode` as can been seen below with a basic validator that always succeeds:
```
> applyArguments testScript [I 1]
"530100003332222220061200120014c101010001"
> testScriptWithArgs 1
"5001000033322222200612001200148009"
```
The reason for this is that `applyArguments` uses `[Data]` for the list of arguments but `applyCode` uses a `Lift` instance to generate the PLC code.
For example for an `Integer` we generate these:
```
> getPlc (liftCode (BuiltinData (I 1)) :: CompiledCodeIn DefaultUni DefaultFun BuiltinData)
Program () (Version () 1 0 0) (Constant () (Some (ValueOf data (I 1))))
> getPlc (liftCode 1 :: CompiledCodeIn DefaultUni DefaultFun Integer)
Program () (Version () 1 0 0) (Constant () (Some (ValueOf integer 1)))
```

If we can generate the PLC from the `Lift` instance then we can use `applyProgram` to apply it to the partial validator and I think it should be the same as using `applyCode`.

ngua:
Nice work @ssorc3! If we could keep using `applyArguments`, that would be ideal

ssorc3:
I don't think we can easily change `applyArguments` because it's used internally by `applyValidator` for example. The only way we can keep using `applyArguments` is to wrap scripts so that they take `Data` or `BuiltinData` instead of Haskell types. 

Alternatively we can use the code I wrote above, which is similar to `applyArguments` but doesn't take `[Data]`. Or maybe it could unwrap the data before applying it but I think that would limit the types you can use as arguments to `Integer` and `ByteString`.

ngua:
> Alternatively we can use the code I wrote above, which is similar to `applyArguments`

I think this is the better choice

The real issue here would be making sure we could obtain the PLC corresponding to a given `Lift` instance, no? It seems that this would unfortunately entail either
- building the PLC on the client from a given type (which would be absurdly complex I think)
- limiting the types that can be used as script parameters as you point out

[...]

PR #190, opened 2022-03-17
Add endpoint wrapping `applyArguments` (with client)

### LINKED ISSUE #24: applyCode
we're going to need to be able to apply arguments to scripts and get the correct resulting hash

example: if a project uses a script that is parameterized over someone's pubkeyhash, then we will need a different version for each pubkey that might need to interact with our dApp.

ideally, we should be working with the stringified, non-CBOR version of the script, and hopefully we can get away with minimal parsing and string manipulation, rather than building out an AST.

we need examples of un-applied and applied scripts to see how to do this properly.

in a worst case scenario we could make this a backend service.

**EDIT**:

Since we are using the Haskell server for this, there are really two tasks:
- [x] Call `applyArgs` on the server and respond with the applied script
- [x] Add a `QueryM` effect to call this on the client side

ngua:
Closes #24. Includes `apply-args` endpoint for applying Plutus scripts and the `QueryM` effect for calling it

**NOTE**: Two of the server tests are failing but we should merge this to unblock more work while investigating the bug (if there is indeed a bug). The endpoint does produce valid scripts after application but the values are different. All of the script fixtures were obtained from @t4ccer by compiling Seabug minting policies. I've opened another issue for this (#189)


src/Types/Scripts.purs
```diff
@@ -32,6 +36,14 @@ derive newtype instance Ord PlutusScript
 instance Show PlutusScript where
   show = genericShow
 
+-- This instance is needed as the server will return a hex-encoded CBOR string
+-- when `applyArgs` is called
+instance DecodeJson PlutusScript where
+  decodeJson j = Json.caseJsonString
+    (Left (Json.TypeMismatch "expected a hex-encoded CBOR string"))
+    (rmap wrap <<< note (Json.UnexpectedValue j) <<< hexToByteArray)
```
vvtran, line 44: Could `map` instead of `rmap`,
or
 `(maybe (Left $ Json.UnexpectedValue j) (Right <<< wrap) <<< hexToByteArray)`

vvtran (approved):
Going to approve to prevent blocking, just one comment.

Issue #205, opened 2022-03-21
Switch to newer version of `purs`


ngua:
#206 replaces all `HashMap`s with `Map`s due to a bug in bundling the project. Apparently this is fixed on latest master of `purs`. We can upgrade the compiler, then revert the changes to code containing `HashMap`s

PR #206, opened 2022-03-21
Apply workaround to enable bundling

ngua:
This makes two changes that allow us to work around a bug with webpack's bundling of the project (unfortunately we are stuck with webpack for now, AFAICT it's the only tool that supports importing WASM with top-level `await`, needed for CSL):

- enabling `bundle` in the `purs-loader` configuration (which calls `spago`'s bundler)
- replaces all `HashMap`s with `Map`s

This is a temporary workaround fix until we can solve the underlying issue with webpack and/or update `purs`. We can track this in a new issue (#205) 

### The original bug
Importing `Types.ScriptLookups` or any other module that imports it results in a runtime error. In short, webpack is not replacing CJS-style `require`s in the JS that `purs` generates and indeed does not include it as an argument:
```js
/***/ ((module) => {
eval("...");
/***/ }),
```
This should be:
```js
/***/ ((module, __unused_webpack_exports, __webpack_require__) => { ... }
```

### The first workaround
We can use `spago bundle-module` to bundle the project into a single JS module before webpack further bundles everything. Unfortunately, this causes another problem

### The second bug
`spago bundle-module` will fail if anything imports `HashMap`s: see for example https://github.com/fehrenbach/purescript-unordered-collections/issues/30. This should be fixed in the latest master of the PS compiler, but upgrading the compiler now would be more involved than ...

### The second workaround
We can just replace all `HashMap`s with `Map`s for now


src/Helpers.purs
```diff
@@ -113,14 +110,14 @@ appendMap
   -> Map k v
 appendMap = Map.unionWith (<>)
 
--- | Provide an append for `HashMap`s where with right bias
-appendRightHashMap
+-- | Provide an append for `Map`s where with right bias
```
vvtran, line 113: This was probably my fault originally but remove the word "where"

vvtran (approved):
LGTM, one small comment.

PR #207, opened 2022-03-21
Viet/remove hash script lookups

vvtran:
- Closes https://github.com/Plutonomicon/cardano-browser-tx/issues/203
- Started a minimal `QueryM` refactor as part of https://github.com/Plutonomicon/cardano-browser-tx/issues/184 this was necessary due to cyclic dependencies. We'd probably want the refactor to mirror the namespace of `Contract`. E.g. `QueryM.Monad`, `QueryM.Scripts` eventually (or whatever is suitable).
- Wrote a stub function `scriptHash :: PlutusScript -> QueryM (Maybe ScriptHash)` and replaced everything affected. Replace with https://github.com/Plutonomicon/cardano-browser-tx/issues/204
- Script lookups for other scripts and minting policy scripts are now formed by an `Array` instead of `Map` so we don't have to hash inside our `Contract` script lookup helpers: `mintingPolicy` and `otherScript`. The tradeoff is we lose this mapping as we run the constraints to create the unbalanced tx, although it's easy to get this information back.
- Extended `Contract` API to include lifted versions of these hashing/Address functions

### DIFF
--- a/src/Contract/Scripts.purs
+++ b/src/Contract/Scripts.purs
@@ -24,7 +24,13 @@ import Address
   , addressStakeValidatorHash
   , addressValidatorHash
   ) as Address
-import QueryM (ClientError(..)) as ExportQueryM
+import QueryM
+  ( ClientError
+      ( ClientHttpError
+      , ClientDecodeJsonError
+      , ClientEncodingError
+      )
+  ) as ExportQueryM
 import QueryM (applyArgs) as QueryM
 import Scripts
   ( typedValidatorAddress
--- a/src/Contract/Scripts.purs
+++ b/src/Contract/Scripts.purs
@@ -4,10 +4,16 @@
 module Contract.Scripts
   ( applyArgs
   , applyArgsM
+  , mintingPolicyHash
+  , scriptHash
+  , stakeValidatorHash
+  , validatorAddress
+  , validatorBaseAddress
+  , validatorHash
   , module Address
-  , module ExportedQueryM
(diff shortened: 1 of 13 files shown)


src/Contract/Scripts.purs
```diff
@@ -18,19 +24,21 @@ import Address
   , addressStakeValidatorHash
   , addressValidatorHash
   ) as Address
-import QueryM (ClientError(..)) as ExportedQueryM
+import QueryM (ClientError(..)) as ExportQueryM
```
adamczykm, line 27: Explicit imports?

src/Scripts.purs
```diff
@@ -86,17 +86,17 @@ validatorHashAddress networkId =
...
-  -> Maybe n
-plutusScriptHash = map wrap <<< scriptHash <<< unwrap
+  -> QueryM (Maybe n)
+plutusScriptHash = map (map wrap) <<< scriptHash <<< unwrap
 
 -- | Converts a `PlutusScript` to a `ScriptHash`.
-scriptHash :: PlutusScript -> Maybe ScriptHash
-scriptHash = scriptHashFromBytes <<< unwrap
+scriptHash :: PlutusScript -> QueryM (Maybe ScriptHash)
+scriptHash = undefined
```
adamczykm, line 102: Intended?

ngua: Yes, I'm going to fix this in a separate PR

ngua (approved):

PR #227, opened 2022-03-23
Adamczykm/submittxnami error fix

adamczykm:


ngua:
I think this fixes the issue. Can you merge master in please so CI will run?

PR #228, opened 2022-03-24
Add endpoint and client for blake2b hashing

### LINKED ISSUE #226: Write `blake2b_256` hashing function for Seabug `Hashable`
The seabug protocol uses a `Hashable` type class
```
class Hashable a where
  hash :: a -> BuiltinByteString
```
which I've attempted to replicate in Purescript as 
```
class Hashable a where
  hash :: a -> Maybe ByteArray
```
`Maybe` because our `Natural` is an underlying `BigInt` and converting to `Int` is generally unsafe.

The underlying hashing function for `BuiltinByteString` is `blake2b_256` https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/PlutusTx-Builtins-Internal.html#v:blake2b_256 which eventually leads to:
https://input-output-hk.github.io/ouroboros-network/cardano-crypto-class/src/Cardano.Crypto.Hash.Blake2b.html#Blake2b_256
https://hackage.haskell.org/package/cryptonite-0.30/docs/src/Crypto.Hash.Blake2b.html#Blake2b_256

We don't have this functionality, I tried using https://github.com/dcposch/blakejs/blob/master/blake2b.js#L327
but have been unable to replicate results in ghci. You can find code in `viet/seabug-contract`

Here's what I ran in Haskell ghci
```
Prelude > 
toLazyByteString $ byteStringHex $ fromBuiltin $ blake2b_256 $ stringToBuiltinByteString "8fc27ab302f4b4b3f7243ef91611575ee0af531cd3df27c2ae7808ca1d96ae4f"
"e06062ddf88d990fb8642de80f5c4a6f4fd5d4a33e0b76c27207037616755db3"
```
for `byteStringHex` and `toLazyByteString` in `Data.ByteString.Builder`

Running the input `"8fc27ab302f4b4b3f7243ef91611575ee0af531cd3df27c2ae7808ca1d96ae4f"` on our PS equivalent doesn't not give the above with hashsize set to 32. See `Types.ByteArray.purs` and `Types.ByteArray.js`.

The JavaScript implementation has a few parameters like hashsize (this has to be 32), salt and personal where the latter are length-16 Uint8Arrays but it's not clear what they should be.

Last resort is to use the Haskell server so that `hash :: a -> QueryM/Contract (Maybe ByteArray)`

ngua:
Closes #226

### DIFF
--- a/node-packages.nix
+++ b/node-packages.nix
@@ -751,15 +751,6 @@ let
         sha512 = "jDctJ/IVQbZoJykoeHbhXpOlNBqGNcwXJKJog42E5HDPUwQTSdjCHdihjj0DlnheQ7blbT6dHOafNAiS8ooQKA==";
       };
     };
-    "blakejs-1.2.1" = {
-      name = "blakejs";
-      packageName = "blakejs";
-      version = "1.2.1";
-      src = fetchurl {
-        url = "https://registry.npmjs.org/blakejs/-/blakejs-1.2.1.tgz";
-        sha512 = "QXUSXI3QVc/gJME0dBpXrag1kbzOqCjCX8/b54ntNyW6sjtoqxqRk3LTmXzaJoh71zMsDCjM+47jS7XiwN/+fQ==";
-      };
-    };
     "bluebird-3.7.2" = {
       name = "bluebird";
       packageName = "bluebird";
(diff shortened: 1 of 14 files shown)


vvtran (approved):
LGTM

PR #229, opened 2022-03-24
Upgrade Cardano libs for Haskell server project

ngua:
Needed for building the transaction on the backend as only a recent version of `cardano-api` has CBOR deserialization for datums, etc...


klntsky:
Please also update `server/README.md` to suggest using `nix develop .#packages.x86_64-linux.hsDevShell`

PR #231, opened 2022-03-24
Properly re-throw exceptions in submitTxNami

### LINKED ISSUE #218: Adjust error-handling for submitting tx in `Wallet` FFI module
Not sure if there has been a breaking change with Nami, but submission failures apparently don't always throw an exception but instead return an object describing the failure. For example:
```js
{
    "code": 2,
    "info": "Wallet could not send the tx.",
    "message": "..."
}
```
So we need to check for this as well as catching actual exceptions

adamczykm:
Closes #218

### DIFF
--- a/src/Wallet.js
+++ b/src/Wallet.js
@@ -4,27 +4,26 @@
 exports._enableNami = () => window.cardano.nami.enable();
 
 // _getNamiAddress :: NamiConnection -> Effect (Promise String)
-exports._getNamiAddress = (nami) => () =>
+exports._getNamiAddress = nami => () =>
   nami.getUsedAddresses().then((addrs) => addrs[0]);
 
 // _getNamiCollateral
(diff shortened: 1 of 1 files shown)


adamczykm:
Created a new one branched out of the current master.

src/Wallet.js
```diff
@@ -4,27 +4,26 @@
...
-    return nami.submitTx(txHex);
-  } catch (e) {
-      console.log(e);
-      throw(e);
-  }
+exports._submitTxNami = txHex => nami => () => {
+  return nami.submitTx(txHex)
+      .catch(e => {
+          console.log("Error in submitTxNami: ", e);
+          throw (JSON.stringify(e));
```
klntsky, line 27: parentheses are optional after throw.

klntsky (approved):

ngua (approved):

PR #232, opened 2022-03-27
Add CBOR decoding for ValidatedTx, Redeemers and Datums

klntsky:
One step forward in implementing server-side finalization

TODOs:

- pass protocol parameters for `hashScriptIntegrity` 
- Add `instance Docs.ToSample FinalizedTransaction`


server/src/Api/Handlers.hs
```diff
@@ -46,6 +67,57 @@ hashScript :: HashScriptRequest -> AppM HashedScript
 hashScript (HashScriptRequest script) =
   pure . HashedScript $ hashLedgerScript script
 
+finalizeTx :: FinalizeRequest -> AppM FinalizedTransaction
+finalizeTx req@(FinalizeRequest {tx, datums, redeemers}) = do
+  pparams <- asks protocolParams
+  liftIO $ do
```
ngua, line 73: Are these `print`s and `putStrLn`s just for debugging (I mean here and elsewhere)?

src/QueryM.purs
```diff
@@ -407,6 +413,66 @@ calculateMinFee tx = do
...
+      <<< asOneOf
+      <$> Serialization.convertTransaction tx
+  -- datums
+  encodedDatums <- liftEffect do
+    for datums \datum -> do
+      byteArrayToHex <<< Serialization.toBytes <<< asOneOf
+        <$> maybe' (\_ -> throw $ "Failed to convert plutus data: " <> show datum) pure
+          (Serialization.convertPlutusData datum)
+  -- redeemers
+  encodedRedeemers <- liftEffect $
```
ngua, line 450: I don't think this approach to redeemers is going to work, see my comment above as well

server/src/Api/Handlers.hs
```diff
@@ -46,6 +67,57 @@ hashScript :: HashScriptRequest -> AppM HashedScript
...
+  let
+    addIntegrityHash t =
+      t { body = body t &
+          \body -> body { scriptIntegrityHash = mbIntegrityHash } }
+    addDatumsAndRedeemers t =
+      t { wits = wits t &
+          \witness -> witness { txdats = txDatums, txrdmrs = decodedRedeemers } }
+    finalizedTx = addIntegrityHash $ addDatumsAndRedeemers decodedTx
+    response = FinalizedTransaction . encodeCborText . Cbor.serializeEncoding $
+      Tx.toCBORForMempoolSubmission finalizedTx
```
ngua, line 109: Hopefully this method of serializing to CBOR will work with Nami. Just making a note here for later debugging (if it comes up of course)

server/src/Types.hs
```diff
@@ -104,6 +107,19 @@ newtype AppliedScript = AppliedScript Ledger.Script
   deriving stock (Show, Generic)
   deriving newtype (Eq, FromJSON, ToJSON)
 
+data FinalizeRequest = FinalizeRequest
+  { tx :: Cbor
+  , datums :: [Cbor]
+  , redeemers :: Cbor
```
ngua, line 113: I think we'll need a different approach for sending the redeemers to the backend. I'll send you a message to discuss

server/src/Types.hs
```diff
@@ -213,6 +237,27 @@ instance Docs.ToSample HashedScript where
       )
     ]
 
+instance Docs.ToSample FinalizeRequest where
```
ngua, line 240: Can you update the server readme with this please? You can use `markdown apiDocs` and copy-paste the new endpoint docs

server/src/Api/Handlers.hs
klntsky: Yes

klntsky: Should I remove them?

ngua: I would say so, when you're done debugging

ngua (approved):
One of us can fix the merge conflicts and then we can merge this in. We might still need to make some adjustments later but it's great we have the endpoint in a working state

vvtran:
Just dealt with conflicts and formatted @klntsky

server/src/Api/Handlers.hs
klntsky: It is working with `serialization-lib`, so there's a good chance it'll work with Nami.

PR #237, opened 2022-03-28
Viet/seabug deployment/list nft

vvtran:
- Moved/added `filterMapWithKeyM` and `filterMapM` to `Contract` API (will make a separeate PR for this commit into `master`)
- Endpoint to list market place NFTs


ngua (approved):
I believe this should work. We can make adjustments later if necessary

PR #241, opened 2022-03-29
Fix `ownPaymentPubKeyHash`

vvtran:
Closes issue https://github.com/Plutonomicon/cardano-browser-tx/issues/236
- Because the `Address` from our wallet is the full address, using `addressBech32` provides the entire bech32string and not the public key bech32string. Subsequently, the call on `ed25519KeyHashFromBech32` doesn't work because it requires the public key version.
- To resolve I had to go via `StakeCredential` but this means we don't have a `pubKeyHash :: PublicKey -> PubKeyHash` function ATM (since `PublicKey` is a wrapper over the bech32string which is an alias for `String`). I've had to comment out relevant functions and lookups, thankfully I don't believe they affect Seabug.


ngua (approved):
Will work for now, great work!

PR #244, opened 2022-03-29
Viet/seabug deployment test utxos at

vvtran:
Closes https://github.com/Plutonomicon/cardano-browser-tx/issues/243
- Fix for `typedValidatorAddress` by removing it (since it goes via a `BaseAddress`, redefining it to be `typedValidatorBaseAddress` to be clear). What we need is to go via an `EnterpriseAddress`, hence the creation of `typedValidatorEnterpriseAddress` - since an `EnterpriseAddress` doesn't care about stake credentials.
- In fact, Plutus' `validatorAddress` function works on `scriptHashAddress` found here: https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Address.html#scriptHashAddress which also sets the stake credential to `Nothing` for now. So I believe it's safe to replace all instances of `typedValidatorAddress` in the codebase by `typedValidatorEnterpriseAddress`.
- I've separated the commits so we can merge the fix into `master`

### DIFF
--- a/examples/Seabug/Contract/MarketPlaceBuy.purs
+++ b/examples/Seabug/Contract/MarketPlaceBuy.purs
@@ -5,8 +5,7 @@ module Seabug.Contract.MarketPlaceBuy
 
 import Contract.Prelude
 import Contract.Address
-  ( NetworkId(TestnetId)
-  , getNetworkId
+  ( getNetworkId
   , ownPaymentPubKeyHash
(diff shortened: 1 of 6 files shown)


adamczykm (commented):
:+1:

ngua (approved):

PR #245, opened 2022-03-29
Server tx finalization with Ogmios submit

klntsky:


adamczykm:
This does not provide a line of description nor links any issues, which makes it hard to determine if the PR meets its goal or put in a context if getting back after an absence.

server/src/Api/Handlers.hs
```diff
@@ -67,7 +67,7 @@ hashScript (HashScriptRequest script) =
   pure . HashedScript $ hashLedgerScript script
 
 finalizeTx :: FinalizeRequest -> AppM FinalizedTransaction
-finalizeTx req@(FinalizeRequest {tx, datums, redeemers}) = do
+finalizeTx (FinalizeRequest {tx, datums, redeemers}) = do
```
adamczykm, line 70: Can we have a docstring here?

src/Contract/Transaction.purs
```diff
@@ -171,4 +183,11 @@ balanceTx = wrap <<< BalanceTx.balanceTx
 
 -- | Attempts to balance an `UnbalancedTx` hushing the error.
 balanceTxM :: UnbalancedTx -> Contract (Maybe Transaction)
-balanceTxM = map hush <<< balanceTx
\ No newline at end of file
+balanceTxM = map hush <<< balanceTx
+
+finalizeTx
```
adamczykm, line 188: a docstring? also I think that we should put some information on why the finalization takes place on the server, what is it or why do we need it

server/plutus-scripts/Main.hs
```diff
@@ -0,0 +1,16 @@
+module Main (main) where
```
adamczykm, line 1: Raison d'être of the module and the package as a whole could be useful.

ngua:
I don't want to merge in all of the commits from my `submit-with-datums` branch, particularly since the plutus-tx-plugin is making CI break (I don't have the time to investigate this at the moment). I've applied they key changes to another branch, so I'm going to close this and open another PR

PR #258, opened 2022-04-01
Address fixes 

vvtran:
- Various address helper fixes and lookup changes. The addresses *seem* to align with what they should be now.
- Added network id to transaction body inside balancer.
- Outstanding issue is we can't get the collateral added.

### DIFF
--- a/src/Address.purs
+++ b/src/Address.purs
@@ -1,9 +1,9 @@
 module Address
-  ( addressMintingPolicyHash
-  , addressScriptHash
-  , addressStakeValidatorHash
-  , addressToOgmiosAddress
-  , addressValidatorHash
+  ( addressToOgmiosAddress
+  , enterpriseAddressMintingPolicyHash
+  , enterpriseAddressScriptHash
+  , enterpriseAddressStakeValidatorHash
+  , enterpriseAddressValidatorHash
   , getNetworkId
   , ogmiosAddressToAddress
   ) where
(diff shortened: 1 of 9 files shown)


src/Types/ScriptLookups.purs
```diff
@@ -621,19 +622,23 @@ addMissingValueSpent = do
     -- Step 4 of the process described in [Balance of value spent]
     lookups <- use _lookups <#> unwrap
     let
-      -- lookups = unwrap $ cfg.scriptLookups
       pkh' = lookups.ownPaymentPubKeyHash
       skh' = lookups.ownStakePubKeyHash
+    -- Potential fix me: This logic may be suspect:
```
adamczykm, line 627: why do you suspect it, and of what? this info belongs to the comment

src/Address.purs
```diff
@@ -47,20 +48,26 @@ addressToOgmiosAddress = addressBech32
 -- `Address` to `ScriptHash`
 --------------------------------------------------------------------------------
 -- | Get the `ScriptHash` with an internal `Address`
-addressScriptHash :: Address -> Maybe ScriptHash
-addressScriptHash = stakeCredentialToScriptHash <=< addressPaymentCred
```
adamczykm, line 51: We lose a lot of generality in this module's API. I wonder if  using `enterpriseAddressToAddress >=> addressScriptHash` isn't a better option.

ngua (approved):

PR #261, opened 2022-04-04
Extend QueryConfig to store arbitrary data

errfrom:
Context: https://github.com/Plutonomicon/cardano-browser-tx/issues/201

### DIFF
--- a/src/Contract/Monad.purs
+++ b/src/Contract/Monad.purs
@@ -1,9 +1,9 @@
 -- | A module defining the `Contract` monad.
 module Contract.Monad
-  ( Contract'(..)
-  , Contract
-  , ContractConfig'(..)
-  , ContractConfig
+  ( Contract
+  , ContractConfig(..)
+  , DefaultContractConfig
+  , wrapContract
   , defaultContractConfig
   , defaultContractConfigLifted
   , liftContractE
@@ -35,7 +35,7 @@ import Effect.Aff (Aff)
 import Effect.Aff.Class (class MonadAff, liftAff)
 import Effect.Class (class MonadEffect, liftEffect)
 import Effect.Exception (Error, throw)
-import QueryM (QueryMExtended, QueryConfig, liftQueryM)
+import QueryM (QueryM, QueryMExtended, QueryConfig)
 import QueryM
   ( DatumCacheListeners
   , DatumCacheWebSocket
@@ -47,6 +47,7 @@ import QueryM
   , OgmiosWebSocket
   , ServerConfig
   , WebSocket
+  , liftQueryM
   , defaultDatumCacheWsConfig
   , defaultOgmiosWsConfig
   , defaultServerConfig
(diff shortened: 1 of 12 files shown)


src/Contract/Monad.purs
```diff
@@ -77,40 +79,47 @@ import Wallet (mkNamiWalletAff)
 -- | All useful functions written in `QueryM` should be lifted into the
 -- | `Contract` monad and available in the same namespace. If anything is
 -- | missing, please contact us.
-newtype Contract (a :: Type) = Contract (QueryM a)
+newtype Contract' (r :: #Type) (a :: Type) = Contract (QueryMExtended r a)
```
adamczykm, line 82: Isn't `# Type` notation going to be deprecated? If so, we should use `Row Type` instead.

adamczykm (commented):
To use Contract monad with extended config, every call will need to be lifted. If this gets used a lot we may want to make it more convenient.

ngua (commented):
Would you mind formatting this please? You can use `make format` in the Nix environment

I agree with @adamczykm that this might not be convenient if used extensively. We have a huge number of `Contract` actions exported in the various `Contract.*` modules. What do you think about adding another type variable to `Contract` directly?

errfrom:
I made all `Contract` functions use a polymorphic version of `Contract` newtype, thus getting rid of the `liftContract'` function, and replaced all occurrences of deprecated `#Type` with `Row Type`, as @adamczykm pointed out.

src/QueryM.purs
```diff
@@ -179,9 +182,28 @@ type QueryConfig =
...
+type DefaultQueryConfig = QueryConfig ()
+
+type QueryM (a :: Type) = ReaderT DefaultQueryConfig Aff a
+
+type QueryMExtended (r :: Row Type) (a :: Type) = ReaderT (QueryConfig r) Aff a
+
+liftQueryM :: forall (r :: Row Type) (a :: Type). QueryM a -> QueryMExtended r a
+liftQueryM = withReaderT toDefaultQueryConfig
+  where
+  toDefaultQueryConfig :: forall (r :: Row Type). QueryConfig r -> DefaultQueryConfig
```
vvtran, line 197: Let's write this as `toDefaultQueryConfig :: QueryConfig r -> DefaultQueryConfig` to prevent shadowing, or use a different letter to `r`.

src/Contract/Monad.purs
```diff
@@ -1,7 +1,9 @@
 -- | A module defining the `Contract` monad.
 module Contract.Monad
-  ( Contract(..)
```
vvtran, line 3: If we're going `derive instance Newtype (Contract r a) _`, we may as well export the data constructor, as we can just `wrap` and `unwrap`?
You should get a warning like
```
An export for Contract hides data constructors but the type declares an instance of Data.Newtype.Newtype.
    Such instance allows to match and construct values of this type, effectively making the constructors public.
  ```

**Review by vvtran — COMMENTED** [review]
Looks good on the whole, there's going to be quite a few conflicts as we merge `seabug-deployment` back into `master` shortly.

File: src/QueryM.purs [hunk 4]
```diff
@@ -170,7 +173,7 @@ type Url = String
 
 -- when we add multiple query backends or wallets,
 -- we just need to extend this type
-type QueryConfig =
+type QueryConfig (r :: # Type) =
```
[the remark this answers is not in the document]
ngua: I think this might have been missed (other `#` occurrences were changed to `Row`)

src/Contract/Monad.purs
```diff
@@ -77,107 +80,112 @@ import Wallet (mkNamiWalletAff)
...
+
+derive instance Newtype (ContractConfig r) _
+
+wrapContract :: forall (r :: Row Type) (a :: Type). QueryM a -> Contract r a
+wrapContract = wrap <<< QueryM.liftQueryM
 
 -- | Throws an `Error` for any showable error using `Effect.Exception.throw`
 -- | and lifting into the `Contract` monad.
-throwContractError :: forall (e :: Type) (a :: Type). Show e => e -> Contract a
+throwContractError :: forall (e :: Type) (r :: Row Type) (a :: Type). Show e => e -> Contract r a
```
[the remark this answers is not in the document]
ngua: We need to add line length to the formatter with a `.tidyrc`. I'll do that soon

ngua:
I'm going to close this PR, fetch your branch, push to the main repo, fix the merge conflicts, and reopen it

PR #262, opened 2022-04-04
Temporary Seabug's 'packaging' solution.

adamczykm:
I have failed in packaging CTL and using it from another project and nix environment, so I propose a workaround put in this PR.

Temporarily it is assumed that the development/deployment of the frontend takes place inside CTL's nix environment. You can find the information on how to use it in `seabug_contracts/README.md`

- Moved Seabugs stuff from examples to the `seabug_contracts` directory
- Created an api.js containing js facing function wrappers
- Created a new module `CallContract.purs` containing functions calling Seabug's contracts and taking care of data types transformation
- For now, I assumed conversions from simple JS types. Should the API use CSL types instead?

Please let me know, what should be added to close the PR and what are the next steps wrt to packaging setup finalization.


package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "test": "spago run --main Test.Main",
     "dev": "BROWSER_RUNTIME=1 webpack-dev-server --progress --hot",
-    "build": "BROWSER_RUNTIME=1 webpack --mode=production"
+    "build": "BROWSER_RUNTIME=1 webpack --mode=production",
+    "bundle-seabug": "BROWSER_RUNTIME=1 webpack --config ./webpack.config.seabug-bundle.js"
```
ngua, line 13: Do we want `--mode=production` for this as well?

seabug_contracts/CallContract.purs
```diff
@@ -0,0 +1,276 @@
...
+-- | Calls Seabugs marketPlaceListNft and takes care of converting data types.
+--   Returns a JS promise holding nft listings.
+callMarketPlaceListNft :: ContractConfiguration -> Effect (Promise (Array ListNftResultOut))
+callMarketPlaceListNft cfg = Promise.fromAff do
+  contractConfig <- buildContractConfig cfg
+  listnft <- runContract contractConfig marketPlaceListNft
+  pure $ buildNftList <$> listnft
+
+-- | Configuation needed to call contracts from JS.
+type ContractConfiguration =
```
ngua, line 103: I suppose this will work, but why not just take `ServerConfig`s, which we've already defined? It would make it easier to update the `defaultContractConfig` as well

seabug_contracts/CallContract.purs
```diff
@@ -0,0 +1,276 @@
...
+      }
+  , nftIdArgs ::
+      { collectionNftTn :: String -- TokenName
+      , price :: BigInt -- Natural
+      , owner :: String -- PaymentPubKeyHash
+      }
+  }
+
+-- Placeholder for types I'm not sure how should we represent on frontend.
+type ValueOut = Array { currencySymbol :: String, tokenName :: String, amount :: BigInt }
```
ngua, line 135: I'm not sure if this representation will work. Why not use the browser-builtin `Map` for the non-Ada assets?

frontend/index.js
```diff
@@ -0,0 +1,8 @@
+"use strict";
+
+const SeabugContracts = require("../seabug_contracts/api.js");
+
+const args = 'test';
+
+console.log("app starting");
+setTimeout(() => SeabugContracts.callMarketPlaceBuyTest(args).then(r => console.log(r)), 1);
```
ngua, line 8: Why the `setTimeout`?

ngua (commented):
It's a shame you didn't have much luck packaging the project to be imported elsewhere. Hopefully I'll have more time to help out with this soon, since we really need to figure out the packaging for other projects

package.json
adamczykm: it is set in the config, i can move it here if that's better

ngua: Ah OK, sorry I missed that

seabug_contracts/CallContract.purs
klntsky: + for `Map`.

But note that unlike PS `Data.Map` it can only hold keys of primitive types by value (it holds object keys by reference). So this is not a universal solution.

adamczykm: Why do you think this may not work? It basically returns in what is available in Plutus by 
[flattenValue]() :: [Value](https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/Plutus-V1-Ledger-Value.html#t:Value) -> [([CurrencySymbol](https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/Plutus-V1-Ledger-Value.html#t:CurrencySymbol), [TokenName](https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/Plutus-V1-Ledger-Value.html#t:TokenName), [Integer](file:///nix/store/52vknxdwxg37m5wdklm17cfcwj0zxrdd-basement-lib-basement-0.0.12-haddock-doc/share/doc/basement/html/Basement-Compat-Base.html#t:Integer))]

seabug_contracts/CallContract.purs
adamczykm: I can do it, but also this way the user-facing structure of the config is independent of how we want to keep things in Purescript modules. Should I go with ServerConfig instead?

ngua (commented):
We also need a project ID specifically for Seabug. I'll add that and fix the merge conflict. Then we can merge @klntsky's PR as well

seabug_contracts/CallContract.purs
ngua: I think this will be fine for now, I don't want to hold up moving forward with Seabug

PR #263, opened 2022-04-04
fix ToData and strip datum and redeemers

vvtran:
- Strip datums and redeemers after constraint lookups.
- Fix ToData instance of  marketplacedatum


ngua (approved):
The datum and redeemer stripping is redundant since we overwrite it on the server, but it won't negatively affect finalizing the tx anyway I suppose

Let's just merge this for now to get the more important fix in (the `ToData` instance)

PR #265, opened 2022-04-04
Various Seabug fixes

vvtran:
- Fix an append issue in script lookups, we were cons'ing before, and that reverses the order.
- Will need to fix `datumHash` which prepends all bytearrays by `[88,32,` i.e `5820` in hex string
- Changed tx input insertion to be via `insert` instead of append for lexicographical sorting.
- This means we can only reindex `Spend` redeemers at the end, after all tx inputs are inserted.


ngua (changes requested):
I'd prefer not to merge in all of the logging and commented-out stuff. We should merge the lookups fixes; for the datum hash issue, let's wait until we've confirmed that this is indeed the solution (I still haven't tested it)

vvtran:
@ngua of course, this isn't ready!

vvtran:
I've had a lot of issues actually testing this afternoon because of webpack, I only got around to testing it now with `NFT-1-2` but it's having index issues again. So I'll need to investigate this properly.

ngua:
I'd like to merge this in now so we can keep seabug-deployment up to date. The only other fix we need is calculating spend redeemer indices, so please open another PR for that

ngua (approved):

PR #267, opened 2022-04-04
Add second Tx signing

klntsky:
Add local signing. Server-side signing which I was working on earlier is also present here, but it's not used and it does not work.

I was able to buy an NFT with this:

https://explorer.cardano-testnet.iohkdev.io/en/transaction?id=d04657a1ee675eb1bacabb1ab819e50c9a9c249de58cc0eecb4c3907cbb7e857

However, the test is not reproducible, you'll get `NFT not found on marketplace` error.


vvtran:
@klntsky There seems to be another bug when switching to `NFT-1-2`, I have updated my branch with script lookups, but kept the 2-byte hack for now. I won't merge this into yours yet in case it contaminates but I think the bug holds your branch too.

ngua:
Since the server endpoint isn't going to work (I suspect the entire issue here is actually the encoding/decoding required by the server), I'm going to close this. I'm merging in changes taken from here though (I gave your authorship @klntsky) for attaching the vkey witness to the CBOR tx

PR #269, opened 2022-04-05
Add lexicographical ordering on tx inputs

vvtran:
- Wrap `datumHash` to `Contract` API.
- Changed all `Array.Cons` to `Array.Insert` to preserve create a lexiographically ordered tx inputs, in line with Plutus. From testing a small sample, it appears that the transaction hash is tested first (on its hexstring representation), then the index. I've defined custom `Ord` instances for `TransactionHash` and `TransactionInput`.
- Added a function `reindexSpentScriptRedeemers` to reindex the `Spend` redeemers *after* balancing. This is important because tx inputs is altered by balancing, so we can't actually index `Spend` redeemers during lookups, it must be done near the end. This was causing issues before.
- Added minting policy/collection 2 to potential inputs.
- So far, `NFT-1-1` and `NFT-1-2` have been used.


ngua (approved):
Hopefully this is the correct ordering, although the minting one might need to be changed

I think we should automatically re-index the spend redeemers after balancing instead of leaving this to the users to do

vvtran:
I'll write a wrapper for this as discussed.

klntsky (approved):

vvtran:
Created a wrapper here https://github.com/Plutonomicon/cardano-browser-tx/pull/270 so I will merge this PR in.

PR #270, opened 2022-04-05
Viet/seabug deployment refactor, add `balanceAndSignTx` and related newtypes 

vvtran:
- Created from https://github.com/Plutonomicon/cardano-browser-tx/tree/viet/seabug-deployment-fixes
- Made the "temporary" `balanceTx'` the main one (which now returns `UnattachedUnbalancedTx` newtype), because even if we eventually attach datums with CSL, we still need the redeemers (at the very least) to reattach after balancing. I kept this separate from `balanceAndSignTx` (see below).
- ^ So `balancedTx'` and `balanceTx` have swapped names unless it's unclear.
- `balanceAndSignTx`  balances the tx, reindex `Spend` redeemers, attaches datums and redeemers and signs returning the `BalancedSignedTransaction` newtype.
- Removed `submitTransaction :: Transaction -> Contract (Maybe TransactionHash)` from `Contract` API to prevent confusion.
- Finally, the user just needs to submit.
- Any feedback would be great, especially better names.

I was thinking of changing the following names (given the logic in `balanceAndSignTx`):
1) `QueryM.FinalizedTransaction` -> `QueryM.AttachedTransaction` (maybe rename `finalizeTx` -> `attachDatumsRedeemersToTx`?)
2) `BalancedSignedTransaction` -> `FinalizedTransaction` (maybe rename `balanceAndSignTx` -> `finalizeTx`?)
Any thoughts?

### DIFF
--- a/src/Contract/Transaction.purs
+++ b/src/Contract/Transaction.purs
@@ -1,46 +1,51 @@
 -- | A module that defines the different transaction data types, balancing
 -- | functionality, transaction fees, signing and submission.
 module Contract.Transaction
-  ( balanceTx
+  ( BalancedSignedTransaction(..)
+  , balanceAndSignTx
+  , balanceTx
   , balanceTxM
   , calculateMinFee
   , calculateMinFeeM
-  , reindexSpentScriptRedeemers
-  , signTransaction
-  , signTransactionBytes
-  , submitTransaction
   , finalizeTx
   , module BalanceTxError
   , module ExportQueryM
   , module JsonWsp
-  , module ScriptLookups
   , module ReindexRedeemersExport
+  , module ScriptLookups
   , module Transaction
   , module TxOutput
   , module UnbalancedTx
+  , reindexSpentScriptRedeemers
+  , signTransaction
+  , signTransactionBytes
+  , submit
   ) where
 
 import Prelude
 import BalanceTx (balanceTx) as BalanceTx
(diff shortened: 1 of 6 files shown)


adamczykm (commented):
lgtm

src/Contract/Transaction.purs
```diff
@@ -206,12 +214,54 @@ finalizeTx tx datums redeemers = wrap $ QueryM.finalizeTx tx datums redeemers
...
+-- | logging and more importantly, the `ByteArray` to be used with `Submit` to
+-- | submit  the transaction.
+balanceAndSignTx
+  :: UnattachedUnbalancedTx
+  -> Contract (Maybe BalancedSignedTransaction)
+balanceAndSignTx
+  (UnattachedUnbalancedTx { unbalancedTx, datums, redeemersTxIns }) = do
+  -- Balance unbalanced tx:
+  balancedTx <- liftedE' $ balanceTx unbalancedTx
+  log "balanceAndSignTx: Transaction successfully balanced"
```
ngua, line 253: I'm not going to hold up this PR for this, but we're already logging way too much. I'd like to avoid adding any more logging

ngua (approved):
I'm definitely in favor of having newtypes for distinguishing different stages of transaction-building, although I think `UnattachedUnbalancedTx` might not be a very clear name. That said, I can't think of a better name :)

I'd like to remove the logging. I think we are in general logging too much

PR #271, opened 2022-04-05
Implement bundling and add usage example 

klntsky:
Based on top of https://github.com/Plutonomicon/cardano-browser-tx/pull/262


adamczykm:
I've successfully bundled and run the example app.
My only comment is that maybe we should move `bundle-seabug-example` to the seabug-example directory.

And these are my suggestions for the README:
```
diff --git a/npm-packages/seabug-example/README.md b/npm-packages/seabug-example/README.md
index eb20496..9ca3557 100644
--- a/npm-packages/seabug-example/README.md
+++ b/npm-packages/seabug-example/README.md
@@ -2,7 +2,7 @@

 A package that uses `cardano-transaction-lib-seabug` as a dependency.

-1. Run `npm run bundle-seabug` to bundle PS code. This will prepare `cardano-transaction-lib-seabug` package.
+1. From CTL root directory, from within nix-shell, run `npm run bundle-seabug` to bundle PS code. This will prepare `cardano-transaction-lib-seabug` package. After that you may leave nix-shell.
 2. Register it as a local package: `cd npm-packages/cardano-transaction-lib-seabug && npm install && npm link .`
 3. Make `cardano-transaction-link-seabug` visible to `seabug-example`: run `npm link cardano-transaction-lib-seabug` in this directory.
 3. Then, from root directory: `npm run bundle-seabug-example`
 ```

ngua (approved):
Very happy to see we could get this working as an actual package despite the workaround for bundling CTL inside the shell first

klntsky:
@ngua we can perform the build from inside nix and run `npm publish` - and since then we wouldn't have any ties to nix anymore.

PR #277, opened 2022-04-06
Snake case fields to camel case

vvtran:
- Changed native fields to camel case before CTL adoption.
- Tried to keep all foreign JS related stuff as snake case.
- Removed `Lib.purs` as it's legacy and unused.
- Removed a few named type instances.


klntsky (approved):

ngua (approved):
Fix the `spago-packages.nix` problem and we can merge this, I'm very happy to see this change

PR #280, opened 2022-04-06
Provide `PlutusAddress` type with `ToData` and `FromData` instances to match `PlutusTx`

errfrom:
Context: https://github.com/Plutonomicon/cardano-browser-tx/issues/193

I copied the `Address`, `Credential` and `StakingCredential` types from `PlutusTx` and introduced `ToData` and `FromData` instances for them, which are supossed to yield the same results as the native Plutus instances.

At the moment I am trying to provide the possibility of conversion between foreign and native `Address` types (based on `CIP-0019` specification [CIPs/CIP-0019 at master · cardano-foundation/CIPs · GitHub](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0019)), so that we can ensure that our foreign `Address` representation has the correct `ToData` and `FromData` instances as well.

The work is still in progress, but I would like to make sure that my approach makes sense and that I didn't miss anything.


src/Types/PlutusAddress.purs
```diff
@@ -0,0 +1,192 @@
+module Types.PlutusAddress where
+
+import Prelude
+import Data.Maybe (Maybe(..))
```
ngua, line 4: We've been using explicit imports for all constructors (even common types like `Maybe`)

src/Types/PlutusAddress.purs
```diff
@@ -0,0 +1,192 @@
...
+import Serialization.Address (Address) as Serialization
+import Serialization.Hash (ed25519KeyHashFromBytes, scriptHashFromBytes) as Hash
+import ToData (class ToData, toData)
+import FromData (class FromData, fromData)
+
+--------------------------------------------------------------------------------
+-- Credential
+--------------------------------------------------------------------------------
+
+data Credential
```
ngua, line 25: If this type is directly based on a Plutus type, could you include a note on which revision you're working off?

src/Types/PlutusAddress.purs
```diff
@@ -0,0 +1,192 @@
...
+  fromData (Constr zero [ credD, stakingCredD ]) =
+    Address <$> ({ addressCredential: _, addressStakingCredential: _ } <$>
+      fromData credD <*> fromData stakingCredD)
+  fromData _ = Nothing
+
+--------------------------------------------------------------------------------
+-- IsoNativeForeign
+--------------------------------------------------------------------------------
+
+class IsoNativeForeign native foreign' | native -> foreign', foreign' -> native where
```
ngua, line 100: Please add (a) kind signature(s), either standalone (i.e. `class IsoNativeForeign :: ...`) or on the type variables

ngua, line 100: Could you add some documentation to this class, please? We should clarify which type is native and which is foreign (I can guess from below, but it's still nice to be explicit about this)

src/Types/PlutusAddress.purs
```diff
@@ -0,0 +1,192 @@
...
+  toNativeType :: foreign' -> Maybe native
+-- TODO: toForeignType :: native -> foreign'
```
ngua, line 102: Will you be implementing this in this PR?

src/Types/PlutusAddress.purs
```diff
@@ -0,0 +1,192 @@
...
+      -- %b0111 | network tag | script hash
+      7 -> buildAddress scriptCredential Nothing
+
+      _ -> Nothing
+    where
+    addrBytes :: Array Int
+    addrBytes = byteArrayToIntArray $ addressBytes addrForeign
+
+    addrType :: Maybe UInt
+    addrType = head addrBytes >>= (pure <<< flip zshr (fromInt 4) <<< fromInt)
```
ngua, line 141: Could you make some of these numbers constants, or add some documentation explaining their significance? This applies to the literals below as well

src/Types/PlutusAddress.purs
ngua, line 100: Also, if this is meant to be a typeclass that we will use for all of the "problematic" types we have (e.g. `Value`), then we should probably put this into its own module

ngua, line 100: There is a `Serialization.Csl` module with `ToCsl` and `FromCsl` classes. We should probably remove that in this PR as well (it's not really used anywhere and this PR would supersede its intended functionality)

src/Types/PlutusAddress.purs
```diff
@@ -0,0 +1,192 @@
...
+  fromData (Constr one [ slotD, txIxD, certIxD ]) =
+    StakingPtr <$> ({ slot: _, txIx: _, certIx: _} <$>
+      fromData slotD <*> fromData txIxD <*> fromData certIxD)
+  fromData _ = Nothing
+
+--------------------------------------------------------------------------------
+-- Address
+--------------------------------------------------------------------------------
+
+data PlutusAddress = Address
```
ngua, line 76: Is there a reason why the type and data constructors are named differently?

ngua, line 76: I would make this a newtype over a record

ngua, line 76: I think the the type constructor should be `Address`, as this is the familiar representation for our users (compared to our other `Address`). I believe @vvtran was planning on re-organizing modules into `Cardano.*` and `Plutus.*` namespaces, so this could live in `Plutus.Types.Address` and be imported as `Plutus.Address` or something

ngua (commented):
I'll need to check the conversion implementation closer, but I've left some preliminary comments below

Please also add some conversion tests as well

src/Types/PlutusAddress.purs
vvtran: I'd also add `derive instance Newtype PlutusAddress _`

vvtran (commented):
Few comments, I agree with @ngua, more documentation would be helpful.

src/Types/PlutusAddress.purs
errfrom: I planned it that way initially, but I may make a seperate PR depending on the amount of work

src/Types/PlutusAddress.purs
```diff
@@ -0,0 +1,192 @@
...
+instance ToData StakingCredential where
+  toData (StakingHash credential) =
+    Constr zero [ toData credential ]
+  toData (StakingPtr ptr) =
+    Constr one [ toData ptr.slot, toData ptr.txIx, toData ptr.certIx ]
+
+instance FromData StakingCredential where
+  fromData (Constr zero [ pd ]) =
+    StakingHash <$> fromData pd
+  fromData (Constr one [ slotD, txIxD, certIxD ]) =
```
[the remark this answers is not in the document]
errfrom: @vvtran Thanks for pointing that out. I must have gotten tired or something to not notice it :)

ngua:
@vvtran I agree, he knows the address stuff the best

src/Plutus/IsoNativeForeign.purs
```diff
@@ -0,0 +1,151 @@
...
+import Plutus.Types.Address (Address) as Plutus
+import Plutus.Types.Credential
+  ( Credential(PubKeyCredential, ScriptCredential)
+  , StakingCredential(StakingHash, StakingPtr)
+  )
+
+-- | This type class asserts that two types (native Purescript and
+-- | foreign Javascript) are isomorphic, so they can be converted
+-- | to each other without losing information.
+class IsoNativeForeign :: Type -> Type -> Constraint
```
ngua, line 25: I would perhaps recommend changing the name of the class and adjusting the comment, since the types are not isomorphic given the signature `frgn -> Maybe native`. This means that information loss is indeed possible

src/Plutus/IsoNativeForeign.purs
```diff
@@ -0,0 +1,151 @@
...
+
+-- | Extracts the variable-length positive number from a byte array
+-- | following the ABNF grammar below:
+-- |
+-- | variable-length-uint =
+-- |     (%b1 | uint7 | variable-length-uint)
+-- |   / (%b0 | uint7)
+-- | uint7 = 7bit
+_parseChainPtr
+  :: forall t
```
ngua, line 123: Kind signature

ngua (commented):
Very nice. Just a few suggestions below. I'd still like some input from @adamczykm since he knows the address-related stuff pretty well

Please merge master in, there's been quite a number of changes recently. And run `make format` again, the formatting config has been updated

src/Types/PlutusAddress.purs
ngua: Feel free to do it in a separate PR. If so, you should also open a new issue for it and give it the "stage 4" label

adamczykm:
Sorry for taking so long to review. I'm on it now.

adamczykm (commented):
I agree with Viet in that that we should probably try to chase a direct way for converting foreign types to Plutus-compatible ToData/FromData instances if there are no native types already in place in CTL.

That is if there is no other reason other than to provide a structure on which defining the instances feels more direct and clear. If there are, then I’m missing a part of the context here would appreciate an introduction 🙂

Having multiple types that have no semantic differences leads to unproportional mental overhead when devising code that has to use them unless we decide on laws of some kind that’d govern such decisions, but this also complexifies the project.

As for the code, I haven't looked closely at the implementation details, but I think that the PR should contain tests against fixtures got via Plutuses ToData instances.

errfrom:
Implemented the `toForeignType` function for Shelley addresses and added some tests covering the conversion between their two representations (both for `toForeignType` and `toNativeType` functions).

It remains an open question, whether we will provide support for Byron and reward addresses, since it seems impossible to represent them using Plutus `Address` type. This is also part of the reason why we have to make `NativeForeignConvertible` functions return an optional result.

PR #282, opened 2022-04-06
Change name, version and add LICENSE

### LINKED ISSUE #281: Fix name and add a LICENSE file
Currently the package name proudly states "realpab @ 1.0.0". I think we should stick to cardano-transaction-lib at 0.1.0.

Also, the repository lacks a LICENSE file so we should pick one. Will **ISC** be okay? @Benjmhart  @ngua 

> A permissive license lets people do anything with your code with proper attribution and without warranty. The ISC license is functionally equivalent to the [BSD 2-Clause](https://choosealicense.com/licenses/bsd-2-clause/) and [MIT](https://choosealicense.com/licenses/mit/) licenses, removing some language that is no longer necessary.

The github repository name should be corrected as well.

adamczykm:
Closes #281


ngua (commented):
I'm definitely in favor of the name change. I'm not sure if we should go with ISC though, our open-source policy prefers MIT or Apache

adamczykm:
I've changed the license and updated every occurrence of the old name which means that it'll fail to build until we rename the repo.
@ngua

ngua:
@adamczykm The build issue is because you changed the haskell.nix selector without changing the project name. I can fix that, it should build after that. We'll change the actual repo name soon

ngua (approved):

PR #296, opened 2022-04-11
Add tests for ApplyArguments

ssorc3:
Removes compiled examples and replaces them with scripts that have been wrapped to accept `BuiltinData`. See #189.


server/nix/default.nix
```diff
@@ -27,7 +27,7 @@ pkgs.haskell-nix.cabalProject {
       cardano-prelude
       optparse-applicative
       plutus-tx
-      plutus-ledger-api
```
ngua, line 30: This package should be added back to `additional`

ngua (commented):
Looks like this should be formatted again since the formatting check is failing on CI. There's a makefile command in the `server` directory (`make format`)

It would be preferable to have the script fixtures generated in the tests, but it looks like `plutus-tx-plugin` is causing Hercules to freak out. I'll investigate that but if we can't get it to work then it's also OK to just include the scripts as CBOR like we did before

server/nix/default.nix
ssorc3: This was removed accidentally instead of plutus-tx-plugin.

ngua (commented):
Looks good, thanks for taking care of this. I'm going to split the CBOR into multiple lines then we can merge it

ngua (approved):

PR #301, opened 2022-04-12
Provide `PlutusAddress type` with `ToData` and `FromData` instances to match `PlutusTx`

errfrom:
Re-play of #280. 

**Last update:**
Implemented the `toForeignType` function for Shelley addresses and added some tests covering conversion between their two representations (both for `toForeignType` and `toNativeType` functions).

It remains an open question, whether we will provide support for Byron and reward addresses, since it seems impossible to represent them using Plutus `Address` type. This is also part of the reason why we have to make `NativeForeignConvertible` functions return an optional result.


src/Plutus/NativeForeignConvertible.purs
```diff
@@ -0,0 +1,289 @@
...
+  )
+
+-- | This type class asserts that two types (native Purescript and
+-- | foreign Javascript) can be converted to each other.
+class
+  NativeForeignConvertible (native :: Type) (frgn :: Type)
+  | native -> frgn
+  , frgn -> native where
+  toNativeType :: frgn -> Maybe native
+  toForeignType :: native -> Maybe frgn
```
adamczykm, line 54: I think that conversions that produce Maybes will make it tiresome to handle Nothing cases in pure code and correctly transport out the errors, but I don't have an idea on how to 'fix' it ad-hoc.
Using something like `ExceptV` or at least `type EitherV er a = Either (Variant er) a` throughout CTL apis and then providing functions for transporting errors, so  sth like:
`toNativeType :: forall r m. frgn -> ExceptV (frgnConvErr :: FrgnConvErr | r) m native`
along with handlers, "monad Identity cases" and hushed versions:
`toNativeType_ :: forall r m. frgn -> Maybe native`
looks like a solution to that in a long run.
cc @ngua

test/Plutus/Address.purs
```diff
@@ -0,0 +1,143 @@
...
+  )
+
+suite :: TestPlanM Unit
+suite = do
+  group "Plutus.Types.Address" $ do
+    group "NativeForeignConvertible instance" $ do
+      group "Shelley addresses" $ do
+        let indices = 0 .. (length addresses - 1)
+        let testData = zip (zip addressesBech32 addresses) indices
+        flip traverse_ testData $ \(Tuple (Tuple addrBech32 addr) addrType) ->
```
adamczykm, line 32: for_ = flip traverse_

adamczykm (approved):
lgtm

ngua (approved):
Very nice work, thank you for taking care of this @errfrom. I'll address some of the review comments and merge this

src/Plutus/NativeForeignConvertible.purs
ngua: I agree that this will be cumbersome to work with. Let's address this in #107 as well. We probably want a single way to handle such cases across the codebase

Issue #93, opened 2022-02-10
support CIP 25 - style metadata


Benjmhart:

zygomeb:
https://cips.cardano.org/cips/cip25/

PR #307, opened 2022-04-14
Add CIP 25 (NFT Metadata Standard) Support

errfrom:
Context: https://github.com/Plutonomicon/cardano-transaction-lib/issues/93
CIP: https://cips.cardano.org/cips/cip25/


src/Metadata/CIP25.purs
```diff
@@ -0,0 +1,261 @@
...
+              caseJsonObject assets $ \objAssets ->
+                for (objToArray objAssets) $ \(assetName /\ contents) -> do
+                  policyId_ <- decodePolicyId policyId
+                  assetName_ <- decodeAssetName assetName
+                  metadataEntryDecodeJson_ policyId_ assetName_ contents
+    where
+    objToArray :: forall a. FO.Object a -> Array (Tuple String a)
+    objToArray = FO.toUnfoldable
+
+    caseJsonObject
```
ngua, line 209: I might recommend changing the name -- it's not exactly name-shadowing since `Data.Argonaut` has been imported qualified, but it might be confusing nonetheless

src/Metadata/CIP25.purs
```diff
@@ -0,0 +1,261 @@
...
+    decodePolicyId :: String -> Either Json.JsonDecodeError MintingPolicyHash
+    decodePolicyId =
+      note (Json.TypeMismatch "Expected hex-encoded policy id")
+        <<< map wrap
+        <<< (scriptHashFromBytes <=< hexToByteArray)
+
+    decodeAssetName :: String -> Either Json.JsonDecodeError TokenName
+    decodeAssetName =
+      note (Json.TypeMismatch "Expected UTF-8 encoded asset name")
+        <<< (mkTokenName <=< byteArrayFromString)
```
ngua, line 225: I'm not exactly sure how we should handle this, but the `<asset_name>` key is indeed UTF-8 (for now) according to the CIP. But `byteArrayFromString` is only for ASCII strings -- so if anything code points in the asset name fall outside of the ASCII range we'll get a `Nothing` back for the whole thing

test/Fixtures.purs
```diff
@@ -910,6 +926,96 @@ seabugMetadataDeltaFixture1 = SeabugMetadataDelta
...
+  , files: []
+  }
+
+cip25MetadataFixture1 :: CIP25Metadata
+cip25MetadataFixture1 = CIP25Metadata
+  [ cip25MetadataEntryFixture1, cip25MetadataEntryFixture2 ]
+
+cip25MetadataJsonFixture1 :: Json.Json
+cip25MetadataJsonFixture1 = fromRight Json.jsonNull $ Json.parseJson $
+  """
```
ngua, line 972: I think for small JSON strings it's OK to use literals, but for larger fixtures like this I'd rather put it into a file. It would be easier to spot any accidental issues with the JSON since editors would highlight them

src/Metadata/CIP25.purs
```diff
@@ -0,0 +1,261 @@
+module Metadata.CIP25
```
ngua, line 1: Another stylistic thing: we've settled on only capitalizing the first letter of initialisms/acronyms (so it would be more consistent to use `Cip25` than `CIP25`)

ngua (changes requested):
Thanks for taking care of this!

I'd like to change some of these `ByteArray`s and make sure the structure of `Cip25Metadata` itself is correct

src/Metadata/CIP25.purs
klntsky: I guess we should use `TextEncoder` and `TextDecoder` JS APIs here. And now we should rename `byteArrayFromString` to something like ` byteArrayFromASCII`.

In Seabug metadata we didn't have arbitrary utf-8 strings, which is the reason why this problem didn't hit us earlier.

src/Metadata/CIP25.purs
```diff
@@ -0,0 +1,261 @@
...
+    files <- obj .:? "files" >>= \arr ->
+      maybe (pure mempty) (traverse Json.decodeJson) (Json.toArray =<< arr)
+    pure $
+      wrap { policyId, assetName, imageUris, mediaType, description, files }
+
+--------------------------------------------------------------------------------
+-- CIP25Metadata
+--------------------------------------------------------------------------------
+
+newtype CIP25Metadata = CIP25Metadata (Array CIP25MetadataEntry)
```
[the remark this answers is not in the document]
errfrom: @ngua The CIP actually states that the structure allows multiple tokens with different minting policies. The only change I made was to flatten the tree structure and move the `policy_id` and `asset_name` fields directly to the `Cip25MetadataEntry` type, so it should be easier to work with.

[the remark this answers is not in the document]
ngua: Ah OK, thanks for the clarification

ngua:
@errfrom You need to regenerate `spago-packages.nix` since you've added new packages. Run `spago2nix generate` from the development shell and commit the updated file

src/Types/ByteArray.purs
```diff
@@ -100,8 +100,8 @@ instance Arbitrary ByteArray where
 
 -- | Convert characters in range `0-255` into a `ByteArray`.
 -- | Fails with `Nothing` if there are characters out of this range in a string.
-byteArrayFromString :: String -> Maybe ByteArray
-byteArrayFromString str = do
+byteArrayFromAscii :: String -> Maybe ByteArray
```
ngua, line 103: I think this name is clearer than the previous one

ngua (approved):
Thanks for taking care of this!

PR #317, opened 2022-04-20
Add configurable logging

### LINKED ISSUE #20: strip logs
once we have a working prototype, we should remove a bunch of logs that are currently built into the js FFI calls, etc.   

ideally the system should work silently except to report errors.

ngua:
Closes #20. Library users can now specify the verbosity of logging. This has required adding `LoggerT` to the existing monad stack and the addition of extra parameters to FFI actions in order to apply the configured log level to FFI calls as well. Additional logging actions have been exported from `Contract.Monad` as well

In order to help users avoid needing to call `QueryM` code directly (e.g. websocket initialization) in case they are not using the `defaultContractConfig`, I've also added a new `ConfigParams` type that can be used to specify the configuration at a higher level


vvtran (approved):
lgtm

Issue #208, opened 2022-03-21
Investigate Plutus `Rational` and align with our implementation


vvtran:
- Plutus' `Rational` behaves differently to our currently implementation, in particular, they don't allow zero denominators and the sign is controlled by the numerator alone meaning the denominator is always positive.
- We'd want to replicate this behaviour as `NatRatio` (a newtype wrapper over `Rational`) is commonly used along with their helpers.
- We also probably want to match their `FromData`/`ToData` instances.
- Extra care should be taken when using `div`, `mod` etc. due to the differences between Euclidean and truncating division in Purescript and Haskell.

PR #334, opened 2022-04-21
Align `Rational` with Plutus implementation

errfrom:
Context: https://github.com/Plutonomicon/cardano-transaction-lib/issues/208


src/Types/Rational.purs
```diff
@@ -1,11 +1,94 @@
...
+  , (%)
+  , recip
+  , numerator
+  , denominator
+  , denominatorAsNat
   ) where
 
+import Prelude
+
+import Data.BigInt (BigInt, fromInt) as BI
```
ngua, line 14: Very minor stylistic thing: we use `BigInt` elsewhere in the codebase and import it qualified as `BigInt`. It would be nice to keep that consistent

src/Types/Rational.purs
```diff
@@ -1,11 +1,94 @@
...
+import Data.Maybe (Maybe(Just, Nothing), fromJust)
 import Data.Ratio (Ratio)
-import Data.BigInt (BigInt)
+import Data.Ratio ((%), numerator, denominator) as R
```
ngua, line 17: I'm not really a fan of single-letter qualified imports, especially when they differ between modules (e.g. `Types.Rational` is imported as `R` in another module)

src/Types/Rational.purs
```diff
@@ -1,11 +1,94 @@
...
+
+derive newtype instance Show Rational
+derive newtype instance Eq Rational
+derive newtype instance Ord Rational
+derive newtype instance Semiring Rational
+derive newtype instance Ring Rational
+derive newtype instance CommutativeRing Rational
+
+instance EuclideanRing Rational where
+  degree _ = 1
```
ngua, line 36: Why not `one`?

ngua (approved):
LGTM, some small comments below. @vvtran can you also take a look, please?

ngua (commented):
I'm just going to move the new module to `Plutus.Types.Rational` and export it from `Contract.Numeric` as well. Otherwise this looks good

src/Types/Rational.purs
```diff
@@ -1,11 +1,96 @@
...
+  fromData (Constr c [ Integer n, Integer d ])
+    | c == zero = reduce n d
+  fromData _ = Nothing
+
+--------------------------------------------------------------------------------
+-- RationalComponent
+--------------------------------------------------------------------------------
+
+class RationalComponent t where
+  reduce :: t -> t -> Maybe Rational
```
vvtran, line 83: I think it might be worth adding a comment to say that our `reduce` mirrors `ratio` from `PlutusTx.Ratio` https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/src/PlutusTx.Ratio.html#ratio as opposed to `reduce` https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/src/PlutusTx.Ratio.html#reduce which does not do any positive den. normalisation. People may get confused by this.

vvtran (commented):
LGTM, few comments.

PR #335, opened 2022-04-21
Deserializing transaction.

adamczykm:
Apologies for the large PR
The PR contains an implementation of Tx deserialisation from JS CSL type into PS CTL type:

```purescript

-- | Deserializes CBOR encoded transaction to a CTL's native type.
deserializeTransaction
  :: forall (r :: Row Type). { txCbor :: ByteArray } -> Err r T.Transaction

-- | Converts transaction from foreign CSL representation to CTL's one.
convertTransaction
  :: forall (r :: Row Type). CSL.Transaction -> Err r T.Transaction

```
As well as new types for error transporting which can be used in other modules if you like it - the contract/querym monad can be wrapped in `ExceptV` and default handlers implemented. Along with the recent LoggerT addition, it should cover all our needs for error tracing/logging/handling which currently is in a bad shape given the library's role in the ecosystem.

More details and remarks:
- Dealing with errors as Nothings mixed with optionality was awful I've proposed a solution for it in the PR which might be adapted to the rest of the modules iteratively.
- No tests for deserialising transactions, but I need to switch to another project  _ASAP_.
- That's a lot of code in our deserialisation modules and there are probably some bugs lurking around. Additionally, we desperately need some kind of regression tests for CSL ffi which would harden our code against more and less subtle changes in CSL, currently, they might go unnoticed...
- I've implemented conversion for only some kinds of certificates...
- I needed to make adjustments to some existing types implementations.

Exunits computation issue was blocked by this PR.


adamczykm:
I've realised I've forgotten to implement one function. Will do it soon.
```exports._unpackProtocolParamUpdate = maybe => protocolParamUpdate => {throw "not implemented";};```

File: src/Error.purs [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+   . DebugWarning
+  => Monad m
+  => ExceptV v m a
+  -> MaybeT m a
+traceAndHushAll act = MaybeT $ runExceptT act >>= case _ of
+  Left v -> traceM v *> pure Nothing
+  Right a -> pure (pure a)
+
+-- | Lifts underlying monad to MaybeT allowing hushing selected errors.
+hushable
```
klntsky, line 76: I don't see how can it be used, could you add an example? Is it adapted from somewhere?

src/Deserialization/BigNum.purs
```diff
@@ -1,14 +1,34 @@
...
+import Error (E, noteE)
 import Serialization.Types (BigNum)
+import Type.Row (type (+))
 
 foreign import bigNumToString :: BigNum -> String
 
 bigNumToBigInt :: BigNum -> Maybe BigInt
 bigNumToBigInt = bigNumToString >>> fromString
+
+bigNumToBigInt' :: forall r. String -> BigNum -> E (FromCslRepError + r) BigInt
```
klntsky, line 23: I see why it's convenient to use functions like this one, but can we perhaps make a more generic function that prepends a given `String` to error message inside `E (something + r) a`?

Pros: separation of concerns, no need to duplicate each function.

src/Deserialization/Transaction.js
```diff
@@ -0,0 +1,225 @@
...
+exports._unpackMetadatums = containerHelper => containerHelper.unpackKeyIndexed;
+
+// foreign import _unpackMetadataMap :: ContainerHelper -> CSL.MetadataMap -> Array(Tuple CSL.TransactionMetadatum CSL.TransactionMetadatum)
+exports._unpackMetadataMap = containerHelper => containerHelper.unpackKeyIndexed;
+
+// foreign import _unpackMetadataList
+//   :: ContainerHelper -> CSL.MetadataList -> Array CSL.TransactionMetadatum
+exports._unpackMetadataList = containerHelper => containerHelper.unpack;
+
+exports._convertMetadatum = metadataCtors => cslMetadatum => {
```
klntsky, line 178: Too much logic here for a FFI function. I'd prefer it to be written with the use of Alternative on PS side.

klntsky (approved):
The code looks good, but I'm not sure all of it is actually runnable since there are no tests. We can merge it though and I'll add tests for it as part of my work on serialization.

src/Error.purs
adamczykm: I've added an example to the docstring.

src/Deserialization/BigNum.purs
adamczykm: I agree that this is a little bit ugly. 
The upside is that it is more convenient to just put any meaningful string instead of "", than to wrap it in `addErrTrace`-like functions.
I'd change it but I don't have time to do it currently. Feel free to revise these, and take a look at `addErrTrace`.

src/Deserialization/Transaction.js
adamczykm: It's good that it's pretty trivial. And ditto.

src/Deserialization/UnspentOutput.purs
```diff
@@ -95,12 +97,13 @@ convertValue value = do
           :: Array (ScriptHash /\ Array (AssetName /\ BigNum))
     -- convert to domain types, except of BigNum
     multiasset'' :: Map T.CurrencySymbol (Map T.TokenName BigNum) <-
-      Map.fromFoldable <$>
```
ngua, line 98: Why has this been changed to `notImplemented`?

ngua (changes requested):
I'd like to replace all of the `Contract` imports. I don't think it's a good idea to import them internally because of issues with cyclic module dependencies

Why has the `multiasset` parsing been changed to `notImplemented`? Won't this break our existing ability to deserialize UTxOs? Or am I missing something?

src/Deserialization/FromBytes.purs
```diff
@@ -1,81 +1,122 @@
...
+  , _fromBytesError
+  , fromBytesError
+  , fromBytes'
   , fromBytes
   , fromBytesEffect
   ) where
 
 import Prelude
 
+import Contract.Prelude (Either(..), hush)
```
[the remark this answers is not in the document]
adamczykm: It's my IDE that does that. Will fix.

src/Deserialization/UnspentOutput.purs
adamczykm: It stopped compiling, I marked it to come back later and forgot. Sorry. Fixed now.

src/Deserialization/Transaction.purs
```diff
@@ -0,0 +1,642 @@
...
+import Serialization.Hash (ScriptHash)
+import Serialization.Types (NativeScripts, PlutusScripts)
+import Serialization.Types as Csl
+import Type.Row (type (+))
+import Types.Transaction as T
+import Types.Value (scriptHashAsCurrencySymbol, tokenNameFromAssetName)
+
+-- | Deserializes CBOR encoded transaction to a CTL's native type.
+deserializeTransaction
+  :: forall (r :: Row Type). { txCbor :: ByteArray } -> Err r T.Transaction
```
klntsky, line 86: Why not just `ByteArray`?

src/Deserialization/Transaction.purs
```diff
@@ -0,0 +1,642 @@
+module Deserialization.Transaction where
```
klntsky, line 1: Please add explicit exports.

src/Deserialization/Transaction.purs
adamczykm: To make Cbor keyword mandatory, as soon as we have newtype wrappers for Cbor Bytearrays we can switch to direct types

src/Deserialization/Transaction.purs
adamczykm: Ok

adamczykm: done

src/Deserialization/Transaction.purs
ngua: Could you add a note with the issue URL please?

src/Deserialization/Transaction.purs
ngua: Looks like that broke formatting

src/Deserialization/Transaction.purs
```diff
@@ -0,0 +1,707 @@
...
+  , convertUpdate
+  , cslNumberToUInt
+  , cslRatioToRational
+  , deserializeTransaction
+  )
+  where
+
+import Prelude
+
+import Contract.Address
```
ngua, line 70: More `Contract` imports in this module

ngua:
Just going to run the full test suite. If it's all good we can merge

src/Types/Transaction.purs
```diff
@@ -77,6 +77,8 @@ module Types.Transaction
 
 import Prelude
 
+import Contract.Numeric.Rational (Rational)
```
ngua, line 80: More `Contract` imports. I can just remove it

adamczykm: How does it always creep in :o

PR #338, opened 2022-04-22
submitTx returning TransactionHash

tippenein:
- [ ] write tests for parsing TransactionHash and the submit itself
- [x] change types for `submit` (String -> TransactionHash)


tippenein:
I see this got fixed! Closing

PR #341, opened 2022-04-22
Exposed nixpkgsFor

bladyjoker:


ngua:
Thanks @bladyjoker. I'm going to close this as I opened #342, which also adds some helpers to the overlay (like `runPursTest`, etc...). I also didn't want to include some of the overlays we need internally (i.e. `haskell-nix.overlay`), so I didn't include that

Issue #174, opened 2022-03-11
Switch to Ogmios for ex unit calculation


vvtran:
- This arises from https://github.com/Plutonomicon/cardano-browser-tx/issues/167 where we intend to add a buffer for executions units of redeemers as a temporary solution.
- To improve on this, we can call Ogmios to add the ex units on a transaction with the redeemers added. The ordered redeemers are contained in the field `redeemers` `ConstraintProcessingState`
- See https://ogmios.dev/mini-protocols/local-tx-submission/#evaluatetx

ngua:
Since we are going to take care of #273 for stage 4.1, we can avoid deserialization entirely, so this issue is blocked until that's complete

ngua:
Moving this back to 4.1

errfrom:
@klntsky Yes, I think we can close it

PR #343, opened 2022-04-22
Fix MustPayToPubKeyAddress in lookups & increase fees

vvtran:
- Fix MustPayToPubKeyAddress in lookups
- I increased fees and ex units for Cardax testing. However, we need https://github.com/Plutonomicon/cardano-transaction-lib/issues/174 to resolve this. Happy to remove the fee buffer stuff for merging to master but we need the lookup change above.


ngua (approved):
Thanks for fixing this so quickly. I'm fine with adding the padding to ex units and fees for now, since Cardax is likely not the only project that will run into issues with this

Issue #278, opened 2022-04-06
Create Plutus-style `Value`


vvtran:
As part of  https://github.com/Plutonomicon/cardano-browser-tx/issues/193

Issue #193, opened 2022-03-17
Address problematic datatypes for `FromData` and `ToData` to match `PlutusTx`


vvtran:
In Purescript, there is an option to use foreign datatypes from JavaScript. One example is the datatype `Serialization.Address.Address` `foreign import data Address :: Type`. This corresponds to the CSL representation of an address, which is in bytes as opposed to the Rust representation, containing actual fields. `PlutusTx` of course has fields too: https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger/html/Ledger-Address.html#t:Address. This will likely be an issue when implementing parallel Purescript `FromData` and `ToData` instances. Why? Because our foreign Address would convert to the `B`/`Bytes` constructor of `Data`/`PlutusData`, whereas `PlutusTx` would index into `Constr` via `PlutusTx.makeIsDataIndexed ''Address [('Address,0)]`.

Similarly, our `Value` is isomorphic to that of `PlutusTx`, however, we currently do not accept the empty bytearray as a currency symbol due to serialisation issues.  Therefore, instead of a nested `Map`, the Purescript `Value` is split into `Coin` and `NonAdaAssets` (see `Types.Value`).

We would therefore need *native* datatypes to match those of `PlutusTx` with mapping functionality.

Here are a list of problematic datatypes:
- `Serialization.Address.Address` (and related addresses like `BaseAddress`). This is built of `Credential` and `StakingCredential` here https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger/html/Ledger-Address.html
- `Types.Value`

vvtran:
Feel free to add/edit as you see fit, especially the list of datatypes.

klntsky:
Synchronized existing types in #202

Issue #234, opened 2022-03-28
Create `RawBytes` and `CborBytes` as newtypes over `ByteArray`


vvtran:
- Due to the different representations we have to deal with, we should newtype `ByteArray` so we can write valid `DecodeJson` instances for example.

adamczykm:
Since they'll be wrapping Bytearray why the Hex suffix?

ngua:
I was going to take care of this but I haven't had the time recently. I think this would be a good issue for a new contributor to take care of

PR #352, opened 2022-04-26
Add Plutus-style `Value`, Implement conversion between our `Value` and Plutus-style `Value`

errfrom:
Context: https://github.com/Plutonomicon/cardano-transaction-lib/issues/278 (as part of https://github.com/Plutonomicon/cardano-transaction-lib/issues/193)

Some points that may require additional consideration:
1. I replaced the `NativeForeignConvertible` typeclass with two separate `FromPlutusType` and `ToPlutusType` typeclasses, that take an arbitrary type constructor `(f :: Type -> Type)` to allow different wrappers for various types (like `Maybe` and `Identity` for `Plutus.Address` and `Plutus.Value` respectively). 

```PureScript
class FromPlutusType :: (Type -> Type) -> Type -> Type -> Constraint
class FromPlutusType f pt t | pt -> t, t pt -> f where
  fromPlutusType :: pt -> f t

class ToPlutusType :: (Type -> Type) -> Type -> Type -> Constraint
class ToPlutusType f t pt | t -> pt, t pt -> f where
  toPlutusType :: t -> f pt
```

2. The `FromPlutusType \ ToPlutusType` instances for `Value`s can be found in the module `Types.Value`. I decided to move the instances in there because of some issues with cyclic dependencies when trying to implement `FromData \ ToData` instances in terms of those of Plutus `Value` to align with the Plutus implementation (Is this the expected behaviour?).


ngua (commented):
I'll review this more thoroughly soon, but some quick thoughts:

- there's a `.DS_STORE` binary that got checked in. Could you add something to the gitignore for this? Maybe we don't have any macOS-specific lines in there yet
- I'm not sure if we want to export the Plutus-style `Value` from `Types.Value`. My understanding is that we'd have `Plutus.*` and `Cardano.*` namespaces. The current `Value` would go into `Cardano.Types.Value` and then we'd have a conversion between them (which is implemented in this PR). I'm not even sure if we should have a `ToData` instance for `Types.Value` at this point (please correct me if that's not the plan @vvtran)

src/Plutus/Types/Value.purs
```diff
@@ -0,0 +1,227 @@
...
+
+-- https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Value.html#checkPred
+checkPred :: (These BigInt BigInt -> Boolean) -> Value -> Value -> Boolean
+checkPred f l r = all (all f) (unionVal l r)
+
+-- https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Value.html#checkBinRel
+-- Check whether a binary relation holds for value pairs of two `Value` maps,
+-- supplying 0 where a key is only present in one of them.
+checkBinRel :: (BigInt -> BigInt -> Boolean) -> Value -> Value -> Boolean
+checkBinRel f l r = checkPred (these (f zero) (f zero) f) l r
```
klntsky, line 200: I'm curious why are there `zero`s. As a result of them, comparison functions do treat negative values differently when they are present only in one of the values.

I.e. `( A -> B1 -> -1) < (A -> B2 -> 1)` is `false` which is counter-intuitive. Code to reproduce:

```
        let b1 = hexToByteArrayUnsafe ""
        let ada = unsafePartial $ fromJust $ V.mkCurrencySymbol b1
        let tn1 = unsafePartial $ fromJust $ mkTokenName $ hexToByteArrayUnsafe "00"
        let tn2 = unsafePartial $ fromJust $ mkTokenName $ hexToByteArrayUnsafe "11"
        let v1 = V.singleton ada tn1 $ negate one
        let v2 = V.singleton ada tn2 $ one
        liftEffect $ Console.log $ show $ Plutus.Types.Value.lt v1 v2
```

(This is not a change request since it's copied from upstream).

src/Plutus/Types/Value.purs
```diff
@@ -0,0 +1,227 @@
...
+
+newtype CurrencySymbol = CurrencySymbol ByteArray
+
+derive newtype instance Eq CurrencySymbol
+derive newtype instance Ord CurrencySymbol
+derive newtype instance FromData CurrencySymbol
+derive newtype instance ToData CurrencySymbol
+
+instance Show CurrencySymbol where
+  show (CurrencySymbol cs) = "(CurrencySymbol" <> show cs <> ")"
```
klntsky, line 214:
```suggestion
  show (CurrencySymbol cs) = "(CurrencySymbol " <> show cs <> ")"
```

src/Plutus/Types/Value.purs
```diff
@@ -0,0 +1,227 @@
...
+
+adaSymbol :: CurrencySymbol
+adaSymbol = CurrencySymbol mempty
+
+mkCurrencySymbol :: ByteArray -> Maybe CurrencySymbol
+mkCurrencySymbol byteArr
+  | byteArr == mempty =
+      pure adaSymbol
+  | otherwise =
+      scriptHashFromBytes byteArr *> pure (CurrencySymbol byteArr)
```
klntsky, line 227:
```suggestion
      scriptHashFromBytes byteArr $> CurrencySymbol byteArr
```

src/Types/Value.purs
```diff
@@ -388,16 +371,51 @@ instance Split Value where
...
+
 instance FromData Value where
-  fromData (PD.List [ coin, nonAdaAsset ]) =
-    Value <$> fromData coin <*> fromData nonAdaAsset
-  fromData _ = Nothing
+  fromData pd = (unwrap <<< fromPlutusType <<< PlutusValue) <$> fromData pd
 
 instance ToData Value where
-  toData (Value coin nonAdaAsset) = toData (coin /\ nonAdaAsset)
+  toData = toPlutusType >>> unwrap >>> unwrap >>> toData
```
vvtran, line 418: Am I correct in thinking we want to remove `ToData`/`FromData` for the CSL-style `Value`? I think it only makes sense to have it for the Plutus version? We similarly want to remove `ToData/FromData` (Purescript) Map and only have AssocMap to prevent confusion.?
I.e. remove
```
instance (ToData k, ToData v) => ToData (Map k v) where
  toData mp = Map $ entries # map (toData *** toData)
    where
    entries = Map.toUnfoldable mp :: Array (k /\ v)
```
and its FromData counterpart.

src/Types/Value.purs
```diff
@@ -388,16 +371,51 @@ instance Split Value where
...
+      adaTokens <- Tuple.snd <$> head adaTokenMap
+      PlutusMap.lookup adaToken adaTokens
+
+    nonAdaAssets :: NonAdaAsset
+    nonAdaAssets = unsafePartial $ fromJust
+      $ mkNonAdaAssetsFromTokenMap
+      $ nonAdaTokenMap <#> \(cs /\ tokens) ->
+          PlutusValue.getCurrencySymbol cs /\ Map.fromFoldable (unwrap tokens)
+
+instance ToPlutusType Identity Value PlutusValue where
```
vvtran, line 398: Correct me if I'm wrong but we have to be careful with these conversions. E.g. if we go from PlutusValue -> CSLValue -> PlutusVaue. The final PlutusValue won't necessarily equal the first due to the position Ada appears in the final value compare to the first. In particular, the Ada position can easily move around in the AssocMap? If so, should we add a comment here regarding this, in particular, CSL Value should only be used for CSL related (not To/FromData) and `Plutus.Types.Value` should be used in the `Contract` Monad and we shouldn't convert back and forth inside `Contract`?

For example, I swapped the ordering of the components in `compositeValue` and ran into issues, e.g. the below fails the test. Let me know if I misunderstood something of course!
```
  compositeValues :: Plutus.Value /\ Types.Value
  compositeValues =
    ( fst nonAdaValues <> fst adaValues <>
        PlutusValue.singleton currencySymbol1' tokenName2 amount2
    ) /\
      ( snd nonAdaValues <> snd adaValues <>
          Value.mkValue mempty nonAdaAsset2
      )
```

vvtran (commented):
Some comments

vvtran:
Ah, I just read @ngua your comment but repeated the same thing in my PR, yes let's remove `ToData/FromData` for `Types.Value` and `Data.Map`. We want the `Contract` Monad to only be exposed to the Plutus-style Value, then all conversions to do CSL should happen under the hood, although I *think* we need to be careful about converting back and forth (see one of my comments).

src/Types/Value.purs
vvtran: I think this an unfortunate result of `AssocMap` not having an `Ord` on keys which is annoying?

src/Plutus/Types/Value.purs
errfrom: @klntsky Thanks for pointing that out! 
There was actually a pretty obvious bug, which caused this weird behaviour. 

It seems to have been fixed after passing `flip f zero` as the first parameter to `these`. Thus, if some amount is present only in the lhs value, it is compared to `zero` and not the other way around.

```PureScript
checkBinRel :: (BigInt -> BigInt -> Boolean) -> Value -> Value -> Boolean
checkBinRel f l r = checkPred (these (flip f zero) (f zero) f) l r
```

src/Types/TokenName.purs
```diff
@@ -0,0 +1,60 @@
...
+
+-- | The empty token name.
+adaToken :: TokenName
+adaToken = TokenName mempty
+
+-- | Create a `TokenName` from a `ByteArray` since TokenName data constructor is
+-- | not exported
+mkTokenName :: ByteArray -> Maybe TokenName
+mkTokenName byteArr =
+  if byteLength byteArr <= 32 then pure $ TokenName byteArr else Nothing
```
klntsky, line 46: Is this raw bytes or CBOR encoded? (in terms of CSL, is it `AssetName.new` or `AssetName.from_bytes`?) 

I believe CBOR could contribute extra bytes to the length, we should be extra careful here and be sure it's not CBOR.

ngua: It should be raw bytes. At least that's how it's used elsewhere in the codebase. Once we have #234 in place this will be clearer

src/Plutus/Types/Value.purs
```diff
@@ -0,0 +1,225 @@
...
+instance Show Value where
+  show (Value mp) = "(PlutusValue " <> show mp <> ")"
+
+instance Semigroup Value where
+  append = unionWith add
+
+instance Monoid Value where
+  mempty = Value (Plutus.Map [])
+
+--------------------------------------------------------------------------------
```
vvtran, line 54: We could alternatively define an `Eq` instance using our own `eq` for convenience? I'll leave that down to you.

src/Plutus/Types/Value.purs
```diff
@@ -0,0 +1,225 @@
...
+
+instance Show Value where
+  show (Value mp) = "(PlutusValue " <> show mp <> ")"
+
+instance Semigroup Value where
+  append = unionWith add
+
+instance Monoid Value where
+  mempty = Value (Plutus.Map [])
+
```
vvtran, line 53: We could add:
```
import Data.Lattice (class JoinSemilattice, class MeetSemilattice)

instance JoinSemilattice Value where
  join = unionWith max

instance MeetSemilattice Value where
  meet = unionWith min
```
they might be useful to someone.

vvtran (commented):
As far as I can tell, `Contract.Value` still re-exports things from `Types.Value` and it should really be `Plutus.Types.Value`, however, there are functions that are missing.

1) Would it be possible to reproduce the following in `Plutus.Types.Value` (you should be able to copy and paste them from `Types.Value`:)
- `currencyMPSHash`
- `mpsSymbol` (and/or `scriptHashAsCurrencySymbol`)
2) There are a few more things like `minus` and `adaOnly` but I think they're not required.
3) Change `Contract.Value` to re-export `Plutus.Types.Value`

vvtran:
This also may be a good opportunity to move `Types.Value` to `Cardano.Types.Value` if we're looking to do incremental refactors. `Types.TokenName` is shared, not sure if we'd want a `Shared.Types` namespace or just leave it in `Types` for now.

src/Types/PubKeyHash.purs
```diff
@@ -0,0 +1,46 @@
...
+  show = genericShow
+
+-- This is needed for `ApplyArgs`. Plutus has an `getPubKeyHash` field so don't
+-- newtype derive.
+instance DecodeJson PubKeyHash where
+  decodeJson = caseJsonObject
+    (Left $ TypeMismatch "Expected object")
+    (flip getField "getPubKeyHash" >=> decodeJson >>> map PubKeyHash)
+
+-- payPubKeyHash :: PaymentPubKey -> Maybe PaymentPubKeyHash
```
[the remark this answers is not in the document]
vvtran: I think I commented these out a while back, as part of the lack of `PublicKey` issue.

src/Plutus/FromPlutusType.purs
```diff
@@ -0,0 +1,177 @@
...
+        _toVarLengthUInt ptr.certIx
+
+-- | Encodes the variable-length positive number as a byte array
+-- | according to the ABNF grammar below.
+-- |
+-- | variable-length-uint =
+-- |     (%b1 | uint7 | variable-length-uint)
+-- |   / (%b0 | uint7)
+-- | uint7 = 7bit
+_toVarLengthUInt :: forall (t :: Type). Newtype t UInt => t -> ByteArray
```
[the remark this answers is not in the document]
errfrom: @ngua That was indeed my original intention, but now I'm going to remove them to align with the style

src/Deserialization/UnspentOutput.purs
```diff
@@ -46,14 +46,13 @@ import Types.TransactionUnspentOutput
...
   , CurrencySymbol
-  , TokenName
   , Value
   , mkCurrencySymbol
   , mkNonAdaAsset
-  , mkTokenName
   , mkValue
   ) as T
-import Types.Value (assetNameName)
+import Types.TokenName (TokenName, mkTokenName) as T
```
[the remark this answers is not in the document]
errfrom: @ngua It might be justified in this particular case, since we have to distinguish between CSL and our types

src/Types/PubKeyHash.purs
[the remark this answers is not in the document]
ngua: Ah OK

ngua (approved):

Issue #355, opened 2022-04-27
Add tests for tx deserialization


ngua:
Tests were not a part of #335, so we should add some. This could also be a part of #174

ngua:
Moving to stage 5 since that's when we'll need this (see #174)

ngua:
Marking this as "invalid" since we might not even need tx deserialization. Waiting for #273 to confirm

klntsky:
Done long ago, including roundtrip tests

PR #356, opened 2022-04-27
Add AuxiliaryData and `is_valid` serialization

### LINKED ISSUE #354: Add serialization for remaining Transaction fields
- [ ] isValid
- [ ] auxiliaryData

klntsky:
Closes #354

There are no tests for now (see #355)

### DIFF
--- a/src/Deserialization/Transaction.purs
+++ b/src/Deserialization/Transaction.purs
@@ -236,7 +236,9 @@ convertTxBody txBody = do
 
   where
   intToSlot
-    :: forall s. Int -> Either (Variant (fromCslRepError :: String | s)) Slot
+    :: forall (s :: Row Type)
+     . Int
+    -> Either (Variant (fromCslRepError :: String | s)) Slot
   intToSlot x =
     cslErr ("validityStartInterval UInt.fromInt': " <> show x)
       <<< map Slot
@@ -426,7 +428,7 @@ convertMetadatum err = fix \_ -> addErrTrace err <<< _convertMetadatum
 
   where
   stringToInt
-    :: forall s
+    :: forall (s :: Row Type)
      . String
     -> String
     -> Either (Variant (fromCslRepError :: String | s)) Int.Int
--- a/src/Serialization/Types.purs
+++ b/src/Serialization/Types.purs
@@ -70,6 +70,8 @@ module Serialization.Types
   , ProtocolParamUpdate
   , ProtocolVersion
   , ExUnitPrices
+  , GeneralTransactionMetadata
+  , TransactionMetadatum
   ) where
 
 import Prelude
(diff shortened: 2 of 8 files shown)


src/Types/Transaction.purs
```diff
@@ -775,7 +775,7 @@ instance Show TransactionMetadatumLabel where
 data TransactionMetadatum
   = MetadataMap (Map TransactionMetadatum TransactionMetadatum)
   | MetadataList (Array TransactionMetadatum)
-  | Int BigInt
+  | Int Int.Int
```
adamczykm, line 778: Why not use a native type here?

adamczykm (commented):
:+1:

src/Types/Int.purs
```diff
@@ -3,21 +3,38 @@ module Types.Int
...
 foreign import newNegative :: BigNum -> Int
 foreign import _intToStr :: Int -> Prim.String
 
-instance Prelude.Eq Int where
-  eq = Prelude.eq `on` _intToStr
+instance Eq Int where
+  eq = eq `on` _intToStr
+
+instance Ord Int where
+  compare = compare `on` \number ->
```
ngua, line 30: Just a stylistic thing: I think it would be nicer point-free

ngua (approved):

src/Serialization/Types.purs
```diff
@@ -81,6 +81,8 @@ module Serialization.Types
   , ProtocolParamUpdate
   , ProtocolVersion
   , ExUnitPrices
+  , GeneralTransactionMetadata
+  , TransactionMetadatum
```
vvtran, line 85: This module has multiple exports of the same thing:
```
ExUnitPrices
GeneralTransactionMetadata
ProposedProtocolParameterUpdates
ProtocolParamUpdate
TransactionMetadatum
Update
```
hopefully I haven't missed any.

src/Deserialization/Transaction.purs
```diff
@@ -416,21 +417,21 @@ convertMetadatum
...
     , from_text: pure <<< Text
     , from_map: convertMetadataMap convertMetadatum
     , from_list: convertMetadataList convertMetadatum
     }
   )
 
   where
-  stringToBigInt
+  stringToInt
     :: forall s
```
vvtran, line 429: :: forall (s :: Row Type)

vvtran: There's also another one in `Deserialization.Transaction` under `intToSlot` if you wouldn't mind changing that.

src/Types/Int.purs
klntsky: Type inference breaks when using `unsafePartial` with composition.

```
  Could not match constrained type

    Partial => t0

  with type

    BigInt
```

src/Types/Transaction.purs
klntsky: It's just like in CSL.

adamczykm: I just wonder, since we're converting the entire thing to native types why leave such things as ints in their foreign CSL versions.

klntsky: My understanding is that

- it's hard to recreate it in pure PS (although possible - can be modeled as `(Sign, BigNum)`)
- we don't need to perform arithmetic operations on it for now

Same goes for `BigNum`.

CC @ngua what do you think?

ngua: I'm fine with using the foreign `Int`

PR #359, opened 2022-04-29
Arion setup for launching runtime dependencies

### LINKED ISSUE #333: Add docker-compose and Arion setup for adopters
We should expose an easy way to compose the various services required for CTL, since most deployment scenarios will be similar/identical

ngua:
Closes #333. This adds the ability to configure, generate, and run Arion expressions (a wrapper around docker-compose) to launch all of CTL's required runtime dependencies

For local development and testing, all that is now needed is `nix run .#ctl-runtime` and everything will be launched. I've removed the old `shellHook` and Makefile commands since the `ctl-runtime` app supersedes them and is much easier to maintain

For CTL users, the overlay exposes functions for generating the Arion expressions. I've added a new guide to the readme illustrating how to use this

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,11 @@ SHELL := bash
 .PHONY: autogen-deps run-testnet-node run-testnet-ogmios
 .SHELLFLAGS := -eu -o pipefail -c
 
-ps-sources := $$(find ./* -iregex '.*.purs')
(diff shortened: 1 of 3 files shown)


bladyjoker (approved):
Nothing to add, this looks great!

PR #361, opened 2022-04-29
Use reconnecting-websocket

### LINKED ISSUE #233: Error Recovery Mode
currently, if a websocket dies, we just clear all listener sets and throw an error.

we should build a specific output (such as the pending queries in a queue to be resent, transactions that may have been submitted) for the error state, and a function which can accept this output, along with new websockets,  and restart those same queries, verify transactions were submitted/settled, etc.

### LINKED ISSUE #321: Fix websocket connection issues
Websocket connection to ogmios and datum cache fail about 50% of the time and there is nothing communicated back to the caller of the CTL functions (i.e., an error logs to the console but frontend is not notified of any failure and things hang).

Existing intel by @nrutledge and t.m.:
> t:"websocket issue is on frontend/ctl side, I've started local ws server with websocat and only 1/4 requests are getting there"
> n:"I've tested a bit on my end by manually creating websocket connections from the browser dev tools and also using Postman. In both cases, when testing, I'm getting identical behavior as the Seabug frontend. In fact, I've tested simultaneously, and if the frontend websocket connection fails, the connection from dev tools or Postman also fails."
> t:"Ok it is nightmare race condition to debug, I created dummy ogmios (not datum cache) server, and I wasn't received some of the requests because if frontend opens websocket to datum cache before everything breaks due to Unexpected server response: 404 returned by ogmios-datum-cache"

Anyone more involved is welcome to provide a more structured description and hints on the issue.

klntsky:
Closes #233
Closes #321

So I decided to give up on implementing WebSocket resurrection by hand and used a library instead, because it implements significant part of the logic leaving only requests queue for our side.

`reconnecting-websocket` maintains a request queue under the hood. If there's no available WebSocket when a request is made, it is put into the queue to be sent when a new websocket connects. So the idea is to collect request bodies for which we haven't received a response yet and when error happens, just resend the requests (they first get pushed to the internal `reconnecting-websocket` queue).


package.json
```diff
@@ -19,6 +19,7 @@
     "big-integer": "1.6.51",
     "bufferutil": "4.0.5",
     "node-polyfill-webpack-plugin": "1.1.4",
+    "reconnecting-websocket": "^4.4.0",
```
ngua, line 22: I think we should use exact versions. My thought was that any JS dependencies obtained directly via `npm` (in case users wanted to use `npm install` or something) should be exactly the same as those we get through Nix (see #286)

src/JsWebSocket.js
```diff
@@ -1,24 +1,39 @@
...
+  constructor (url, protocols, options) {
+    options = options || {};
+    options.perMessageDeflate = false;
+    super(url, protocols, options);
+  }
+};
+
 // _mkWebsocket :: (String -> Effect Unit) -> String -> Effect WebSocket
-exports._mkWebSocket = logger => url => () => {
+exports.mkWebSocket = logger => url => () => {
```
ngua, line 21: Applies to this whole module and the `foreign import`s in PS: why remove the leading underscore? That was our convention for FFI stuff, right?

src/QueryM.purs
```diff
@@ -813,70 +772,79 @@ listeners :: forall (listeners :: Type). WebSocket listeners -> listeners
...
+        Ref.modify_ (Map.insert id req) pr
   }
 
-removeAllListeners
-  :: forall (a :: Type). LogLevel -> DispatchIdMap a -> Effect Unit
-removeAllListeners lvl dim = do
-  logString lvl Error "error hit, removing all listeners"
-  Ref.write MultiMap.empty dim
-
 -- TODO after ogmios-datum-cache implements reflection this could be generalized to make request for the cache as well
```
ngua, line 818: I know this isn't part of the issues that this PR addresses, but do you think it would be possible to address this TODO as well? (Since you're already changing the internals of the jsonwsp stuff.) AFAIK @adamczykm made a PR against ODC that implemented reflection 

If you don't want to include it in this PR, we should open an issue for this and add the URL

src/QueryM/ServerConfig.purs
```diff
@@ -0,0 +1,53 @@
+module QueryM.ServerConfig where
```
ngua, line 1: Explicit exports

ngua (commented):
Looks like a good solution. ~One thing that I think is missing (unless I've misunderstood something) is the retry logic for the datum cache connection. IIRC this connection was the most unreliable for the Seabug deployment~ Edit: got the different connections messed up, ignore that comment please :)

My thoughts on that: I think we can get rid of the distinction between Ogmios and the datum cache (see comment below) since reflection was implemented for ODC. Then we can use the same retrying for that connection after some adjustments. Otherwise we can just make the same changes to `QueryM.DatumCacheWsp` as well

ngua:
Also, unless I'm mistaken, this should also close #321 since it includes switching to `reconnecting-websocket`

src/JsWebSocket.js
klntsky: I thought that having a public module interface without underscores is prioritized over keeping FFI clearly separate.
I'll restore the underscores.

klntsky:
This should not be merged yet. reconnecting-websocket exposes ES module interface (we use commonjs everywhere) which leads to runtime errors in bundled code.

klntsky:
Also note that I didn't implement retries counting is that `reconnecting-websocket` already has this functionality with sane defaults.

https://www.npmjs.com/package/reconnecting-websocket#default-values

ngua (approved):

PR #368, opened 2022-05-03
Adjust lookups and constraints

### LINKED ISSUE #200: Review `Scriptlookups` and `TxConstraints` logic
- Some of the functionality used OgmiosTxOuts as a replacement for ChainIndexTxOut, but we may want to change this to an internal tx output type to keep ogmios tx outs under the cover e.g
```
unspentOutputs
  :: forall (a :: Type). Map TxOutRef OgmiosTxOut -> ScriptLookups a
```
Furthermore, constraints like `mustPayToTheScript` don't really make sense because we aren't validating a "current" script, there are just many scripts as part of a tx body.

As a consequence, we can probably remove OgmiosTx related types from the `Contract` monad.

ngua:
Closes #200:

- I removed all references to `OgmiosTxOut` from `Contract`. This is an internal type and we don't need to export anything related to it
- I removed `mustPayToTheScript` since the "current script" is a Plutus concept that we can't port over
- I renamed `mustPayToOtherScript` to `mustPayToScript`, since we don't really need the "other" designation
- Similarly, I renamed `otherScript` and `otherData` to `validator` and `datum`, respectively. Those were the best names I could come up with, but I'm open to other suggestions of course
- I had to rewrite some of the internal `processConstraint` logic to avoid shadowing the new names


vvtran (approved):
lgtm

PR #370, opened 2022-05-04
Handle missing Nami collateral

### LINKED ISSUE #305: Handle missing Nami collateral
At the moment, connecting to a Nami wallet that has not set its special collateral can lead to difficult-to-track bugs. We need to handle this explicitly, either in `namiUtxosAt` where we attempt to filter the collateral out (necessary if collateral is set, otherwise Nami will freak out during balancing/submission) or during wallet initialization (perhaps we should throw an exception?)

jy14898:
Closes #305

### DIFF
--- a/src/QueryM/Utxos.purs
+++ b/src/QueryM/Utxos.purs
@@ -14,10 +14,12 @@ import Data.Bifunctor (bimap)
 import Data.Bitraversable (bisequence)
 import Data.Map as Map
 import Data.Maybe (Maybe, maybe)
-import Data.Newtype (unwrap, wrap)
+import Data.Newtype (unwrap, wrap, over)
 import Data.Traversable (sequence)
 import Data.Tuple.Nested (type (/\))
 import Effect.Aff (Aff)
+import Effect.Class (liftEffect)
+import Effect.Exception (throw)
 import Helpers as Helpers
 import QueryM (QueryM, getWalletCollateral, mkOgmiosRequest)
 import Serialization.Address (Address)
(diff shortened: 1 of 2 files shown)


src/QueryM/Utxos.purs
```diff
@@ -89,13 +91,12 @@ utxosAt addr = asks _.wallet >>= maybe (allUtxosAt addr) (utxosAtByWallet addr)
...
   namiUtxosAt :: Address -> QueryM (Maybe Transaction.UtxoM)
-  namiUtxosAt address = do
-    utxos' <- allUtxosAt address
-    collateral' <- getWalletCollateral
-    pure do
-      utxos <- unwrap <$> utxos'
-      collateral <- unwrap <$> collateral'
-      pure $ wrap $ Map.delete collateral.input utxos
+  namiUtxosAt address = getWalletCollateral >>= maybe
+    (liftEffect $ throw "Nami wallet missing collateral")
```
jy14898, line 95: I'm throwing the error in Aff here, but above it suggests we should move to Either for fine-grained error handling. Although, it seems the errors inside the Maybe are only a small subset of the possible errors from calling Ogmios?

jy14898:
I haven't added the check during wallet initialisation as I think it's somewhat superfluous to have (at least without the changes I've made thus far). If a wallet lost their collateral during a previous transaction or removed it themselves, we'd be back to square one with difficult to solve bugs.

It may still be worth adding however just to fail faster

src/QueryM/Utxos.purs
ngua: I think throwing an exception is fine. This isn't an error that can be meaningfully recovered from. Nami doesn't really work without collateral set

ngua:
I think we should also add a check during wallet initialization

src/QueryM/Utxos.purs
ngua: That said, we can also change the return type of `utxosAt` to `Either ...`

jy14898:
@ngua I've added the check at wallet init.

Would you like me to make the changes to `utxosAt` to move to using `Either`?

jy14898:
FYI I've yet to test this PR on a real wallet as I'm waiting for cardano-node to finish syncing. Should be done soon

ngua:
I can try running the example with Nami

jy14898:
> I can try running the example with Nami

If you'd like, but I think it's about to finish in the next couple of minutes. Currently at block 57264210 of 57376972

ngua:
OK, I ran it with my main account with collateral as well as a new account without, seems to be working as intended

ngua (approved):

Issue #369, opened 2022-05-04
Add additional `EncodeJson` instances


peter-mlabs:
`CurrencySymbol` has a [`DecodeJson` instance](https://github.com/Plutonomicon/cardano-transaction-lib/blob/cb0af8f023a7f5f1cadeba8a8f2e02523f661371/src/Cardano/Types/Value.purs#L190-L197), but no `EncodeJson`. This means that types that are exported via the `argonaut` option to purescript-bridge fail when they wrap a currency symbol.

ngua:
We can also add one for `Rational`. Let me know if there's more types that need the instance

PR #371, opened 2022-05-05
Add EncodeJson and DecodeJson for a Rational, TokenName, CurrencySym

adamczykm:
Addresses #369

### DIFF
--- a/src/Cardano/Types/Value.purs
+++ b/src/Cardano/Types/Value.purs
@@ -47,7 +47,6 @@ module Cardano.Types.Value
 
 import Prelude hiding (join)
 
-import Contract.Prim.ByteArray (byteArrayToHex)
 import Control.Alt ((<|>))
 import Control.Alternative (guard)
 import Data.Argonaut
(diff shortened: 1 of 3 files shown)


src/Types/TokenName.purs
```diff
@@ -10,6 +10,16 @@ module Types.TokenName
 
 import Prelude
 
+import Contract.Prelude (Either(Left), note)
```
ngua, line 13: `Contract` import again

src/Cardano/Types/Value.purs
```diff
@@ -47,12 +47,15 @@ module Cardano.Types.Value
 
 import Prelude hiding (join)
 
+import Contract.Prim.ByteArray (byteArrayToHex)
```
ngua, line 50: `Contract` import

src/Types/Rational.purs
```diff
@@ -34,6 +42,30 @@ derive newtype instance Semiring Rational
...
+  encodeJson r = encodeJson
+    { "unRational":
+        { "numerator": encodeJson (StringifiedBigInt (numerator r))
+        , "denominator": encodeJson (StringifiedBigInt (denominator r))
+        }
+    }
+
+newtype StringifiedBigInt = StringifiedBigInt BigInt
+
+derive instance Eq StringifiedBigInt
```
ngua, line 55: This could be a `newtype` instance

src/Types/TokenName.purs
```diff
@@ -10,6 +10,16 @@ module Types.TokenName
...
+import Contract.Prim.ByteArray (byteArrayToHex, hexToByteArray)
```
ngua, line 14: Here too

ngua (changes requested):
We should avoid `Contract` imports in internal modules

ngua (approved):

Issue #353, opened 2022-04-27
Set transaction metadata


ngua:
We currently don't have a way to set transaction metadata. Various projects need this functionality though

Tasks:
- [ ] Allow setting the `Transaction.auxiliaryData` field
- [x] (De)serialization for that field (all implemented now)
- [ ] Conversion between the typed metadata API (i.e. `Metadata.Cip25`/`Seabug`) and `GeneralTransactionMetadata`
- [ ] A way to calculate and set the auxiliary data hash
- [ ] Tests for the above functionality

For the last task, this could be achieved by passing a HOF to a general conversion function (e.g. `Cip25Metadata -> GeneralTransactionMetadata` and its inverse) or through a typeclass (e.g. `ToFromMetadata`)

We might also need to use the server for the third task, depending on what CSL gives us as the hash

ngua:
Related: #323

ngua:
Closed by #375

PR #375, opened 2022-05-05
Set transaction metadata

errfrom:
Context: https://github.com/Plutonomicon/cardano-transaction-lib/issues/353

1. Conversion between metadata types (`Metadata.Cip25`/`Seabug`) and `GeneralTransactionMetadata` can be performed via `fromGeneralTxMetadata` and `toGeneralTxMetadata` functions. They require metadata types to conform to the `MetadataType` typeclass, which has the following definition: 
```PureScript
class (FromMetadata a, ToMetadata a) <= MetadataType (a :: Type) where
  metadataLabel :: Proxy a -> TransactionMetadatumLabel
```
2. To calculate the auxiliary data hash, `hashAuxiliaryData` is used, which is merely a wrapper around the corresponding CSL function.
```PureScript
foreign import auxiliaryDataHashBytes
  :: AuxiliaryData -> ByteArray

hashAuxiliaryData :: T.AuxiliaryData -> Effect T.AuxiliaryDataHash
hashAuxiliaryData =
  map (wrap <<< auxiliaryDataHashBytes) <<< convertAuxiliaryData
```
3. It is now possible to set the `auxiliaryData` and `auxiliaryDataHash` fields of `UnattachedUnbalancedTx` by using the functions defined in the `Contract.AuxiliaryData`, though there might be a better way to define them.


src/Metadata/ToMetadata.purs
```diff
@@ -0,0 +1,85 @@
...
+  toMetadata = Bytes
+
+instance ToMetadata String where
+  toMetadata = Text
+
+--------------------------------------------------------------------------------
+-- AnyToMetadata
+--------------------------------------------------------------------------------
+
+newtype AnyToMetadata = AnyToMetadata
```
ngua, line 76: My assumption is that this type would be for a heterogeneous collection of `ToMetadata`s, correct (I do think this would be useful for users)? Could we add a doc comment for that?

src/Serialization/AuxiliaryData.js
```diff
@@ -15,6 +15,9 @@ exports.setTxAuxiliaryData = setter('auxiliary_data');
 exports.newAuxiliaryData = () =>
     lib.AuxiliaryData.new();
 
+exports.auxiliaryDataHashBytes = auxiliaryData =>
```
ngua, line 18: I think this should be renamed to `_hashAuxiliaryData` or something similar. The first thing I think of is converting an `AuxiliaryDataHash` to bytes, not hashing the `AuxiliaryData`

ngua (approved):
Looks good, thanks! Just a couple of suggestions below. As a sidenote, I'm hoping that using CSL to hash the auxiliary data will work, since we've had issues with other hashing functions. That doesn't need to be addressed here though

ngua (approved):

PR #383, opened 2022-05-06
Firefrorefiddle/bytestypes

firefrorefiddle:
Create Wrappers for `CborBytes` and `RawBytes` as proposed in #234 . Not sure where  `RawBytes` should be made explicit or which other places would need the `CborBytes` wrapper.


adamczykm:
According to CSL's docstrings hashes are the only types that don't use CBOR representation. So both .to_bytes() and .from_bytes() use RawBytes. While it's possible to assume that this is just a 'simplified CBOR' representation, I think that it will be more clear not to use CborBytes wrappers around hashes.

@ngua wdyt?

adamczykm:
According to CSL's docstrings hashes are the only types that don't use CBOR representation. So both .to_bytes() and .from_bytes() use RawBytes. While it's possible to assume that this is just a 'simplified CBOR' representation, I think that it will be more clear not to use CborBytes wrappers around hashes.

@ngua wdyt?

ngua:
@adamczykm I see, I was under the impression that CSL types indicated CBOR or raw bytes by the name of the method (for example `to_bytes` vs `bytes`). If that's the case then the hashes should be `RawBytes` though. I'm not surprised that hashes wouldn't be CBOR encoded though, there's really no point since they're just bytes anyway

src/Types/CborBytes.purs
```diff
@@ -0,0 +1,79 @@
...
+  , cborBytesToHex
+  , byteLength
+  , hexToCborBytes
+  , hexToCborBytesUnsafe
+  ) where
+
+import Data.Newtype (class Newtype, wrap, unwrap)
+
+import Types.ByteArray (ByteArray)
+import Types.ByteArray as BA
```
ngua, line 19: I think we've started avoiding one- or two-letter qualified module names. `ByteArray` would be better I think

src/Types/RawBytes.purs
```diff
@@ -0,0 +1,79 @@
...
+  , rawBytesToHex
+  , byteLength
+  , hexToRawBytes
+  , hexToRawBytesUnsafe
+  ) where
+
+import Data.Newtype (class Newtype, wrap, unwrap)
+
+import Types.ByteArray (ByteArray)
+import Types.ByteArray as BA
```
ngua, line 19: Same here

src/Wallet.purs
```diff
@@ -26,6 +26,7 @@ import Helpers ((<<>>))
 import Serialization as Serialization
 import Serialization.Address (Address, addressFromBytes)
 import Types.ByteArray (ByteArray, hexToByteArray, byteArrayToHex)
+import Types.CborBytes (CborBytes(..))
```
ngua, line 29: Explicit constructor imports

ngua (changes requested):
Looks pretty good. A few things:

- You're using the `CborBytes` constructor in several places, but it's probably better to use `wrap` instead
- There are a few more places where we can switch from `ByteArray` to `CborBytes`: 
  - anywhere with a record that contains `txCbor` (the record can be removed and the type changed)
  - `TokenName` -- the type itself should be `CborBytes` but the `mkTokenName` function should take `RawBytes`
  - in the `Wallet` module there are several effects that use `ByteArray`s, these should be `CborBytes`
- We should probably change the type of the hashes as @adamczykm suggests

ngua:
Oh we also re-export these types from `Contract.Prim.ByteArray`

src/Types/CborBytes.purs
```diff
@@ -0,0 +1,79 @@
...
+import Prelude
+import Aeson (class DecodeAeson, decodeAesonViaJson)
+import Data.Argonaut (class DecodeJson, decodeJson)
+import Test.QuickCheck.Arbitrary (class Arbitrary, arbitrary)
+
+-- | An array of Bytes containing CBOR data
+newtype CborBytes = CborBytes ByteArray
+
+instance Show CborBytes where
+  show rb = "(cborBytesFromIntArrayUnsafe" <> show (cborBytesToIntArray rb) <>
```
vvtran, line 30: Might be worth adding a space after the initial string `cborBytesFromIntArrayUnsafe`.
Also, we've transitioned to use its hex representation for `Show`, as it's a little easier to debug with so I would suggest `cborBytesToHex` instead of `cborBytesToIntArray`

src/Types/CborBytes.purs
```diff
@@ -0,0 +1,79 @@
...
+    ")"
+
+derive instance Newtype CborBytes _
+
+derive instance Eq CborBytes
+
+derive instance Ord CborBytes
```
vvtran, line 37: I believe
```
derive newtype instance Eq CborBytes
derive newtype instance Ord CborBytes
```
should be better. See my comment a few below this too.

src/Types/CborBytes.purs
```diff
@@ -0,0 +1,79 @@
...
+
+instance Semigroup CborBytes where
+  append r1 r2 = wrap $ unwrap r1 `append` unwrap r2
+
+instance Monoid CborBytes where
+  mempty = cborBytesFromIntArrayUnsafe []
```
vvtran, line 43: I would write this as `mempty = wrap mempty` but down to you. Although see the comment below.

src/Types/CborBytes.purs
```diff
@@ -0,0 +1,79 @@
...
+
+instance DecodeJson CborBytes where
+  decodeJson = map wrap <<< decodeJson
+
+instance DecodeAeson CborBytes where
+  decodeAeson = decodeAesonViaJson
+
+instance Arbitrary CborBytes where
+  arbitrary = wrap <$> arbitrary
```
vvtran, line 52: I think we can have something more efficient like this
```

derive instance Newtype CborBytes _
derive newtype instance Eq CborBytes
derive newtype instance Ord CborBytes
derive newtype instance Semigroup CborBytes
derive newtype instance Monoid CborBytes
derive newtype instance DecodeJson CborBytes
derive newtype instance DecodeAeson CborBytes
derive newtype instance Arbitrary CborBytes
```

src/Types/RawBytes.purs
```diff
@@ -0,0 +1,79 @@
...
+import Prelude
+import Aeson (class DecodeAeson, decodeAesonViaJson)
+import Data.Argonaut (class DecodeJson, decodeJson)
+import Test.QuickCheck.Arbitrary (class Arbitrary, arbitrary)
+
+-- | An array of Bytes with no information about the content format
+newtype RawBytes = RawBytes ByteArray
+
+instance Show RawBytes where
+  show rb = "(rawBytesFromIntArrayUnsafe" <> show (rawBytesToIntArray rb) <> ")"
```
vvtran, line 30: Again, a space after the initial string. And use the hex representation to `rawBytesToHex` show.

src/Types/RawBytes.purs
```diff
@@ -0,0 +1,79 @@
...
+  mempty = rawBytesFromIntArrayUnsafe []
+
+instance DecodeJson RawBytes where
+  decodeJson = map wrap <<< decodeJson
+
+instance DecodeAeson RawBytes where
+  decodeAeson = decodeAesonViaJson
+
+instance Arbitrary RawBytes where
+  arbitrary = wrap <$> arbitrary
```
vvtran, line 51: Again, I'd newtype derive where possible
```

derive newtype instance Eq RawBytes
derive newtype instance Ord RawBytes
derive newtype instance Semigroup RawBytes
derive newtype instance Monoid RawBytes
derive newtype instance DecodeJson RawBytes
derive newtype instance DecodeAeson RawBytes
derive newtype instance Arbitrary RawBytes
```

You'll need to clean up your redundant imports after this.

vvtran (commented):
LGTM, a few comments on newtype deriving and the `Show` instance.

src/Types/CborBytes.purs
ngua: Good catch! I'd say adding a space is important to make sure that the `Show` instance is a valid Purescript expression that will evaluate to the `show`n value

src/ToData.purs
```diff
@@ -200,6 +201,9 @@ instance ToData a => ToData (Ratio a) where
 instance ToData ByteArray where
   toData = Bytes
 
+instance ToData CborBytes where
```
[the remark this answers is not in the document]
firefrorefiddle: > but it's probably better to use wrap instead

I thought it would be clearer to use the constructor in cases like this one:

```
instance FromData BaseAddress where
  fromData (Bytes res) = baseAddressFromBytes (CborBytes res)
  fromData _ = Nothing
```

Because it makes it obvious to any reader that the bytes are hereby declared to contain cbor. `wrap` doesn't say anything. I am happy to change it to `wrap` though if you prefer it that way.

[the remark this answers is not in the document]
firefrorefiddle: >>>
@adamczykm I see, I was under the impression that CSL types indicated CBOR or raw bytes by the name of the method (for example to_bytes vs bytes). 
>>>

And that seems to be assumed in `Hash.js`/`Hash.purs` too, as there are two distinct functions.

```
exports.scriptHashAsBytes = h => h;
exports.scriptHashToBytes = hashToBytes;
```

So are they the same after all?

src/Wallet.purs
```diff
@@ -25,7 +25,14 @@ import FfiHelpers (MaybeFfiHelper, maybeFfiHelper)
 import Helpers ((<<>>))
 import Serialization as Serialization
 import Serialization.Address (Address, addressFromBytes)
-import Types.ByteArray (ByteArray, hexToByteArray, byteArrayToHex)
+import Types.ByteArray (byteArrayToHex)
+import Types.CborBytes
+  ( CborBytes
+  , hexToCborBytes
```
vvtran, line 31: Redundant import.

src/ToData.purs
```diff
@@ -36,6 +37,9 @@ import Prim.TypeError (class Fail, Text)
 import Record as Record
 import Type.Proxy (Proxy(Proxy))
 import Types.ByteArray (ByteArray(ByteArray))
+import Types.RawBytes (RawBytes)
```
vvtran, line 40: This is a duplicate import.

src/Serialization/Hash.purs
```diff
@@ -48,21 +49,21 @@ instance Ord Ed25519KeyHash where
...
+  show edkh = "(Ed25519KeyHash " <> rawBytesToHex (ed25519KeyHashToBytes edkh)
     <> ")"
 
 instance ToData Ed25519KeyHash where
-  toData = toData <<< ed25519KeyHashToBytes
+  toData = toData <<< unwrap <<< ed25519KeyHashToBytes
 
 instance FromData Ed25519KeyHash where
-  fromData (Bytes kh) = ed25519KeyHashFromBytes kh
+  fromData (Bytes kh) = ed25519KeyHashFromBytes <<< wrap $ kh
```
vvtran, line 59: If you like, you could write this as `ed25519KeyHashFromBytes $ wrap kh`

vvtran (commented):
Some redundant imports around and a comment.

vvtran:
There's quite a few conflicts too, we have a lot of recently merged PRs and more to come.

src/Serialization/Hash.purs
firefrorefiddle: If you say it's nicer, then I will. ;-)

firefrorefiddle:
> There's quite a few conflicts too, we have a lot of recently merged PRs and more to come.

Fixed the conflicts. Only imports yet.

src/Wallet.purs
```diff
@@ -192,6 +200,6 @@ foreign import _submitTxNami
   -> Effect (Promise String) -- Submitted tx hash
 
 foreign import _attachSignature
-  :: ByteArray -- CBOR bytes of tx
-  -> ByteArray -- CBOR bytes of witness set
-  -> Effect (ByteArray)
+  :: CborBytes -- CBOR bytes of tx
+  -> CborBytes -- CBOR bytes of witness set
+  -> Effect (CborBytes)
```
vvtran, line 205: Just a stylistic nitpick but the parenthesis are redundant, I appreciate this wasn't your code.

vvtran (approved):
LGTM, will approve with the optional stylistic change

src/Cardano/Types/Value.purs
```diff
@@ -220,7 +221,7 @@ unsafeAdaSymbol = CurrencySymbol mempty
 -- | constructor is not exported
 mkCurrencySymbol :: ByteArray -> Maybe CurrencySymbol
 mkCurrencySymbol byteArr =
-  scriptHashFromBytes byteArr *> pure (CurrencySymbol byteArr)
+  scriptHashFromBytes (wrap byteArr) *> pure (CurrencySymbol byteArr)
```
ngua, line 224: I know this wasn't you, but I think:

```suggestion
  scriptHashFromBytes (wrap byteArr) $> CurrencySymbol byteArr
```
is nicer

ngua (approved):
Looks good!

src/Cardano/Types/Value.purs
ngua: Actually I'm going to open a PR with some other stylistic changes for things that were recently discovered, I can take care of this then

PR #394, opened 2022-05-10
change show instance of ByteArray to hex

vvtran:
- Change `Show` instance of `ByteArray` to be hex instead of an int array.

### DIFF
--- a/src/Types/ByteArray.purs
+++ b/src/Types/ByteArray.purs
@@ -29,7 +29,7 @@ newtype ByteArray = ByteArray Uint8Array
 derive instance Newtype ByteArray _
 
 instance Show ByteArray where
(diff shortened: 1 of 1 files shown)


src/Types/ByteArray.purs
```diff
@@ -29,7 +29,7 @@ newtype ByteArray = ByteArray Uint8Array
 derive instance Newtype ByteArray _
 
 instance Show ByteArray where
-  show arr = "(byteArrayFromIntArrayUnsafe " <> show (byteArrayToIntArray arr)
+  show arr = "(byteArrayToHex " <> show (byteArrayToHex arr)
```
ngua, line 32:
```suggestion
  show arr = "(hexToByteArrayUnsafe " <> show (byteArrayToHex arr)
```

ngua (approved):

Issue #380, opened 2022-05-06
`PPViewHashesDontMatch` when paying without minting


ngua:
Strangely, we get a `PPViewHashesDontMatch` (or its Ogmios wrapper, `extraDataMismatch`) when paying to script inputs _without_ also minting something. That is, we need to mint something in order to get the tx to validate

vvtran:
I will have a crack at fixing this as it appeared under a project I'm in.

vvtran:
When changing the datum for sending value to a script, the error message changes
```
nexpected parse error on input:{"type":"jsonwsp/response","version":"1.0","servicename":"ogmios","methodname":"SubmitTx","result":{"SubmitFail":[{"extraDataMismatch":{"provided":"f88f09a0153e89ee57ed4a7385867307968332c76e62a6d61b689647cf6ba053","inferredFromParameters":"2f50ea2546f8ce020ca45bfcf2abeb02ff18af2283466f888ae489184b3d2d39"}}]},"reflection":{"id":"SubmitTx-l2uh8nrf","step":"INIT"}}
```
to
```
unexpected parse error on input:{"type":"jsonwsp/response","version":"1.0","servicename":"ogmios","methodname":"SubmitTx","result":{"SubmitFail":[{"extraDataMismatch":{"provided":"3965346f650a23bba6a6d791999a0e2db4fb8017ee421ea7c6034b9464102ab6","inferredFromParameters":"3aa207d08590b85a3003d45091ac302ed9c51101c9f967f448a05692401bf06a"}}]},"reflection":{"id":"SubmitTx-l2uhmpj7","step":"INIT"}}
```
notice how both the inferred and provided changes which is confusing.

Issue #340, opened 2022-04-22
Extra datum with simple Ada transfer


ngua:
Attempting to submit the transaction in `examples/Pkh2Pkh.purs` results in:

```
{
  "type": "jsonwsp/response",
  "version": "1.0",
  "servicename": "ogmios",
  "methodname": "SubmitTx",
  "result": {
    "SubmitFail": [
      {
        "extraDataMismatch": {
          "provided": "6c5f384a73bacc120b6b78c55d1078d0e422898b75b5434127ba028479747347",
          "inferredFromParameters": null
        }
      }
    ]
  },
  "reflection": { "id": "SubmitTx-l29wf6qu", "step": "INIT" }
}
```
Avoiding `finalizeTx` entirely solves this, but that is not a good solution

adamczykm:
Will try to tackle it now. I'd appreciate any hints.

adamczykm:
It seems that indeed server's transaction finalization adds extra scriptdatahash.

Before finalization:
```purescript
[...]

PR #395, opened 2022-05-10
add null redeemers logic to languages in finalizeTx

vvtran:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/380
In particular:
- We can now send value to a script address without requiring a dummy minting policy to mint something - it turns out that a redeemer was required with our current logic. The Alonzo cddl has specific requirements for empty redeemers found here https://github.com/input-output-hk/cardano-ledger/blob/master/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L123
Some `PlutusV2` info can be found here also https://github.com/input-output-hk/cardano-ledger/blob/81a309fa6410d4216724a3eb32e4141da40733e9/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L117

### DIFF
--- a/server/src/Api/Handlers.hs
+++ b/server/src/Api/Handlers.hs
@@ -102,7 +102,11 @@ finalizeTx (FinalizeRequest {tx, datums, redeemers}) = do
   decodedDatums <-
     throwDecodeErrorWithMessage "Failed to decode datums" $
       traverse decodeCborDatum datums
-  let languages = Set.fromList [PlutusV1]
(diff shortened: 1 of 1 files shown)


vvtran:
I tested this on a contract I'm working on elsewhere locally and the transaction completes without requiring the extra minting policy/redeemer.

adamczykm (approved):
nice catch

ngua (approved):

ngua:
This is probably going to conflict with #340, but it's needed critically so we should merge it sooner rather than later

PR #396, opened 2022-05-10
Add CLI options to connect to cardano-node

klntsky:
We need to access network id and node socket to calculate exunits.

### DIFF
--- a/server/exe/Main.hs
+++ b/server/exe/Main.hs
@@ -44,11 +44,13 @@ opts =
         "CBTx server. See the README for routes and request/response types"
 
 networkIdReader :: Options.ReadM C.NetworkId
-networkIdReader = Options.eitherReader $ \arg -> case arg of
-  "mainnet" -> return C.Mainnet
-  networkMagicStr -> case readMaybe networkMagicStr of
-    Just word32 -> return $ C.Testnet (C.NetworkMagic word32)
-    Nothing -> Left "Failed to parse network ID"
+networkIdReader = Options.eitherReader $ \case
+  "mainnet" -> pure C.Mainnet
+  networkMagicStr ->
+    maybe
+      (Left "Failed to parse network ID")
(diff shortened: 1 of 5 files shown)


server/exe/Main.hs
```diff
@@ -18,31 +19,39 @@ import Network.Wai.Handler.Warp (
...
-  Options.info (serverOptions <**> Options.helper) $
+  Options.info (serverOptionsParser <**> Options.helper) $
     Options.fullDesc
       <> Options.progDesc
         "CBTx server. See the README for routes and request/response types"
 
-serverOptions :: Options.Parser ServerOptions
-serverOptions =
+networkIdReader :: Options.ReadM C.NetworkId
+networkIdReader = Options.eitherReader $ \arg -> case arg of
```
ngua, line 47: Why not use lamda case here?

server/exe/Main.hs
```diff
@@ -53,3 +62,20 @@ serverOptions =
...
+          <> Options.showDefault
+          <> Options.value "./node/socket/node.socket"
+          <> Options.metavar "PATH"
+      )
+    <*> Options.option
+      networkIdReader
+      ( Options.long "network-id"
+          <> Options.help "Network id: either 'mainnet' or a NetworkMagic number"
+          <> Options.showDefault
+          <> Options.value C.Mainnet
```
ngua, line 79: Using `Mainnet` as the default is conceptually simpler, but CTL in general doesn't actually support mainnet at the moment. I'm wondering if this should be changed to testnet?

ngua (dismissed):
We need to adjust the Nix setup to pass the correct arguments to the server executable

Also I'm not sure what the default value for the socket path should be

ngua (approved):
I took care of my two suggestions. Again, I'm not sure if we should be using mainnet as the default network ID, but I don't want to hold this up for that

PR #397, opened 2022-05-10
fix extra datum with simple ada transfer

### LINKED ISSUE #340: Extra datum with simple Ada transfer

Luis-omega:
fixes #340


server/src/Api/Handlers.hs
```diff
@@ -102,7 +103,14 @@ finalizeTx (FinalizeRequest {tx, datums, redeemers}) = do
   decodedDatums <-
     throwDecodeErrorWithMessage "Failed to decode datums" $
       traverse decodeCborDatum datums
-  let languages = Set.fromList [PlutusV1]
+  let witnessScripts = (Tx.txscripts' . Tx.wits) decodedTx
+      (Value.Value ada maps) = (Tx.mint . Tx.body) decodedTx
+      languages =
+        if Map.null witnessScripts
+          && Map.null maps
+          && ada == 0
```
klntsky, line 111: This looks a bit different from a similar change in [another PR](https://github.com/Plutonomicon/cardano-transaction-lib/pull/395/files#diff-4f1478ccd6d4b99631ab8cf039548961315e8091fe4b5dafa1d6561a38bc670eR106).

What documentation this code is based on?

ngua: @klntsky These things are often undocumented and the changes from the other PR were a related but separate issue

ngua, line 111: I don't see why the Ada value would have an effect here

server/src/Api/Handlers.hs
```diff
@@ -102,7 +103,14 @@ finalizeTx (FinalizeRequest {tx, datums, redeemers}) = do
   decodedDatums <-
     throwDecodeErrorWithMessage "Failed to decode datums" $
       traverse decodeCborDatum datums
-  let languages = Set.fromList [PlutusV1]
+  let witnessScripts = (Tx.txscripts' . Tx.wits) decodedTx
+      (Value.Value ada maps) = (Tx.mint . Tx.body) decodedTx
```
ngua, line 107: I think `assets` might be a clearer name than `maps`

ngua (commented):
I'll merge this with the changes from #395 and then we can merge this. I don't think we need to bother with checking the Ada value in `mint` since only minting scripts should affect this

ngua (approved):

Issue #123, opened 2022-02-17
Figure out inconsistencies in fee calculation


ngua:
We've discussed this before, but there are discrepancies between the fee estimation endpoint merged in #55, certain Cardano CLI tools, etc... There are two unfortunate possible consequences to this:

- the fee is too low and validation fails
- the fee is too high and the user pays an unnecessary amount in excess

We can try to get around the first problem by adding a buffer or padding to the fees if they are consistently too low, but we really need to figure out the underlying issue at some point soon

Issue #360, opened 2022-04-29
Calculate ex units on server


bladyjoker:
To successfully run Plutus Scripts we need correct execution units (xu) budget calculations to be performed.

Options that we've scoped out so far that will get us there asap are:
1. Ogmios [evaluateTx](https://ogmios.dev/api/modules/_cardano_ogmios_client.TxSubmission.html) method
2. mlabs-haskell/bot-plutus-interface implementation (using cardano-node lib)

It would be good to know how the 2 methods compare in terms of xus calculated.

ngua:
Related: #174

ngua:
For reference, the BPI approach: https://github.com/mlabs-haskell/bot-plutus-interface/blob/91b18e1d73227d21a99007cb1568341dfa6c3391/src/BotPlutusInterface/CardanoCLI.hs#L514

ngua:
We'll go with the server now, it will be faster and less error-prone to implement

ngua:
Since the tx deserialization code is untested (and bugs are fairly likely given our past experiences with CSL integration), I'm going to use the server to take care of this. I'm planning on extending `finalizeTx` to take care of it

PR #401, opened 2022-05-11
Set transaction execution units, Calculate script execution fees

errfrom:
Context: https://github.com/Plutonomicon/cardano-transaction-lib/issues/360 and https://github.com/Plutonomicon/cardano-transaction-lib/issues/123


errfrom:
I haven't yet tested the new functionality introduced in this PR, and it would be great if someone could help me out with that.

server/README.md
```diff
@@ -82,111 +82,14 @@ The server executable can be built with `nix build .#cardano-trasaction-lib-serv
 "b8fe9f7f6255a6fa08f668ab632a8d081ad87983c77cd274e48ce450f0b349fd"
 ```
 
-## GET /fees
+## GET /eval-ex-units
```
klntsky, line 85: Looks like you accidentally removed some docs from README.

server/src/Api/Handlers.hs
```diff
@@ -65,6 +78,49 @@ estimateTxFees (WitnessCount numWits) cbor = do
...
+      throwM (CborDecode err)
+    Right (C.Tx txBody@(C.TxBody txBodyContent) _) -> do
+      sysStart <- queryNode C.QuerySystemStart
+      eraHistory <- queryNode (C.QueryEraHistory C.CardanoModeIsMultiEra)
+      pparams <- asks protocolParams
+      let eraInMode = C.AlonzoEraInCardanoMode
+          txInputs = Set.fromList . fmap fst $ Shelley.txIns txBodyContent
+          eval = C.evaluateTransactionExecutionUnits
+      utxos <- queryUtxos txInputs
+      case (eval eraInMode sysStart eraHistory pparams utxos txBody) of
```
klntsky, line 94:
```suggestion
      case eval eraInMode sysStart eraHistory pparams utxos txBody of
```

server/src/Types.hs
```diff
@@ -86,10 +89,25 @@ newEnvIO serverOptions =
...
+          C.CardanoModeParams (C.EpochSlots 21600)
+      , localNodeNetworkId = networkId opts
+      , localNodeSocketPath = nodeSocket opts
+      }
+
 newtype Cbor = Cbor Text
   deriving stock (Show)
   deriving newtype (Eq, FromHttpApiData, ToHttpApiData, FromJSON, ToJSON)
 
+newtype ExecutionUnitsMap = ExecutionUnitsMap ([(Cbor, Cbor)])
```
adamczykm, line 107: Since the ctor is exported it's possible to construct invalid values for this type. Wouldn't defining it as `ExecutionUnitsMap [(TxWitness.RdmrPtr, Scripts.ExUnits)]` be worthwhile?
Also, you don't need the parens here.

adamczykm, line 107: Since the ctor is exported it's possible to construct invalid values for this type. Wouldn't defining it as `ExecutionUnitsMap [(TxWitness.RdmrPtr, Scripts.ExUnits)]` be worthwhile?
Also, you don't need the parens here.

adamczykm (commented):
Overall I believe we need more code commentary especially docstrings for AppM actions.

server/src/Types.hs
```diff
@@ -86,10 +92,34 @@ newEnvIO serverOptions =
...
+          C.CardanoModeParams (C.EpochSlots 21600)
+      , localNodeNetworkId = networkId opts
+      , localNodeSocketPath = nodeSocket opts
+      }
+
 newtype Cbor = Cbor Text
   deriving stock (Show)
   deriving newtype (Eq, FromHttpApiData, ToHttpApiData, FromJSON, ToJSON)
 
+newtype ExecutionUnitsMap = ExecutionUnitsMap [RdmrPtrExUnits]
```
adamczykm, line 110: :+1:

server/README.md
errfrom: I've actually removed the part of the docs that were duplicated

ngua: Are you sure? It looks like a huge amount of the API docs are missing. How did you generate them?

src/QueryM.purs
```diff
@@ -468,14 +471,17 @@ instance Show ClientError where
       <> err
       <> ")"
 
+txToHex :: Transaction -> Effect String
```
ngua, line 474: Maybe we should move this to `Helpers`, since it's also defined in the `Wallet` module

ngua: Actually that might lead to cyclical module imports unfortunately

ngua (commented):
Great work! I'm just going to test this out with a few contracts. I also think some of the documentation needs to be restored

src/Types/ScriptLookups.purs
```diff
@@ -165,6 +171,7 @@ import Cardano.Types.Value
   , getNonAdaAsset
   )
 import TxOutput (transactionOutputToScriptOutput)
+import Data.Lens.Getter (view)
```
ngua, line 174: Redundant import here

ngua (commented):
A few more things that we need to do:
- update the `ctl-runtime` in the flake to use the correct socket path (edit: forgot to mount the volume when updating the ctl-server service)
- remove the padded ex units from the constraints processing. These can be set to zero now (this will also be necessary when/if we switch to Ogmios IIRC)

ngua:
OK I've run the `AlwaysMints` example and the ex units are indeed being calculated and updated in the redeemers

ngua:
However the tx fails to submit with a budget error:
```
The budget was overspent. Final negative state: ({ cpu: -369459 | mem: -394 })
```

errfrom:
@ngua Good news!

I finally managed to submit a valid transaction (`AlwaysMints`) with correctly calculated execution units and fees.

Some considerations: 

1. There seems to be an unfortunate mismatch between the formal description of `minfee` function in Alonzo specification _(Figure 4: Functions related to fees and collateral)_ and the way it is defined in `Cardano.Ledger.Shelley.LedgerState`. Namely the implementation doesn't take script fees into account.

```Haskell
minfee pp tx =
  Coin $
    fromIntegral (getField @"_minfeeA" pp)
      * getField @"txsize" tx + fromIntegral (getField @"_minfeeB" pp)
```

2. Because of this, I decided to extend the `eval-ex-units` server endpoint to also calculate script fees based on the execution units and their prices (which can be read from protocol parameters). 
Response format:

```JSON
{"rdmrPtrExUnitsList":[{"exUnitsSteps":0,"rdmrPtrTag":0,"exUnitsMem":0,"rdmrPtrIdx":0}],"txScriptsFee":"0"}
```

3. This PR might probably close https://github.com/Plutonomicon/cardano-transaction-lib/issues/123 as well, since the difference between expected fees and newly calculated ones (basic fee + scripts fee) is now pretty insignificant (though there is some). I tested it on the `AlwaysMints` example and got 2640 lovelace difference, which is more like a margin of error. Therefore I changed the fee "padding" to 10000 lovelace, which hopefully should be enough to cover this difference.

4. Execution units are now recalculated in `BalanceTx` each time `calculateMinFee` is called, as previously discussed in the Slack channel. 

```PureScript
evalExUnitsAndMinFee
  :: UnattachedUnbalancedTx
  -> QueryM (Either BalanceTxError (UnattachedUnbalancedTx /\ BigInt))
evalExUnitsAndMinFee uaubTx =
  runExceptT do
    let tx = uaubTx ^. _unbalancedTx <<< _transaction
    basicFee <- ExceptT $ calculateMinFee tx
      <#> bimap CalculateMinFeeError' unwrap
    tx' <- ExceptT $ reattachDatumsAndRedeemers uaubTx
      <#> lmap ReindexErrors'
    ExceptT $ evalTxExecutionUnits tx'
      <#> bimap EvalTxExUnitsError'
        \(EvalTxExUnitsResponse { rdmrPtrExUnitsList, txScriptsFee }) ->
          updateTxExecutionUnits uaubTx rdmrPtrExUnitsList
            /\ (basicFee + txScriptsFee)
```

GeorgeFlerovsky:
> There seems to be an unfortunate mismatch between the formal description of `minfee` function in Alonzo specification (Figure 4: Functions related to fees and collateral) and the way it is defined in `Cardano.Ledger.Shelley.LedgerState`. Namely the implementation doesn't take script fees into account.

Wait... Why are you using `Cardano.Ledger.Shelley.LedgerState`? That module defines the fee calculation for _Shelley-era_ transactions, which did not even have scripts. Thus, it would make sense for that era's `minfee` function not to include script costs.

Why not use `minfee` from `Cardano.Ledger.Alonzo.Tx`, which includes the script cost and should be what the remote node uses when validating our transactions on mainnet?

https://github.com/input-output-hk/cardano-ledger/blob/2acff66e84d63a81de904e1c0de70208ff1819ea/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs#L339-L346

GeorgeFlerovsky:
Are you also using Shelley-era ledger types elsewhere in CTL, where better types exist in the Alonzo/Babbage eras?

ngua:
@GeorgeFlerovsky No, we're using Alonzo-era types

server/README.md
ngua: Oh, never mind. Sorry @errfrom I see what you mean now

server/src/Api/Handlers.hs
```diff
@@ -42,18 +45,27 @@ import Types (
...
+    EraMismatchError,
+    ScriptExecutionError,
+    TxValidityIntervalError
+  ),
   Cbor (Cbor),
   CborDecodeError (InvalidCbor, InvalidHex, OtherDecodeError),
-  CtlServerError (CborDecode),
+  CtlServerError (CardanoError, CborDecode),
   Env (protocolParams),
+  ExecutionUnitsMap (..),
```
ngua, line 58: We should use explicit constructor/field imports. I see that we weren't actually following that convention 100% in this module, would you mind updating the other ones as well?

ngua (approved):
Excellent work @errfrom! 

I have some comments on style below, but I wanted to approve this to unblock merging

src/BalanceTx.purs
```diff
@@ -521,10 +639,11 @@ returnAdaChange changeAddr utxos (Transaction tx@{ body: TxBody txBody }) =
...
               tx' = wrap tx { body = txBody' }
 
-            fees'' <- lmap ReturnAdaChangeCalculateMinFee <$> calculateMinFee'
-              tx'
+            let uaubTx'' = uaubTx' # _unbalancedTx <<< _transaction .~ tx'
+            fees'' <- evalExUnitsAndMinFee uaubTx''
+              <#> lmap \_ -> ReturnAdaChangeCalculateMinFee
             -- fees should increase.
-            pure $ fees'' >>= \fees' -> do
+            pure $ fees'' >>= \(uaubTx''' /\ fees') -> do
```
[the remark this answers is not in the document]
errfrom: @ngua Going to fix this as part of the next PR

src/BalanceTx.purs
```diff
@@ -271,27 +403,31 @@ balanceTx (UnbalancedTx { transaction: unbalancedTx, utxoIndex }) = do
...
+    nonAdaBalancedCollTx <- ExceptT $ loop allUtxos ownAddr [] unattachedTx''
     -- Return excess Ada change to wallet:
     unsignedTx <- ExceptT $
       returnAdaChange ownAddr allUtxos nonAdaBalancedCollTx <#>
         lmap ReturnAdaChangeError'
-
-    -- Sort inputs at the very end so it behaves as a Set.
-    let sortedUnsignedTx = unsignedTx # _body <<< _inputs %~ Array.sort
+    let
+      unattachedTx = unsignedTx ^. _transaction'
```
vvtran, line 424: This is being shadowed.

vvtran (commented):
I haven't had enough time to review this properly, but left one comment for now.

adamczykm (commented):
:+1:

ngua (approved):
I went ahead and took care of the merge conflict and last comment from @vvtran, just so we can merge this. Thank you for the great work @errfrom!

PR #402, opened 2022-05-11
Add Plutus compatible Json instance for Types.Scripts

bladyjoker:


src/Plutus/Types/CurrencySymbol.purs
```diff
@@ -10,15 +10,25 @@ module Plutus.Types.CurrencySymbol
 
 import Prelude
 
+import Contract.Prelude (Either(..))
+import Data.Argonaut
+  ( class DecodeJson
+  , class EncodeJson
+  , JsonDecodeError(..)
```
ngua, line 17: Usually we use explicit constructor imports

src/Types/Scripts.purs
```diff
@@ -10,7 +10,17 @@ module Types.Scripts
 
 import Prelude
 
-import Data.Argonaut (class DecodeJson, class EncodeJson)
+import Contract.Prelude (Either(..))
+import Data.Argonaut
+  ( class DecodeJson
+  , class EncodeJson
+  , JsonDecodeError(..)
```
ngua, line 17: Same here

src/Types/Scripts.purs
```diff
@@ -10,7 +10,17 @@ module Types.Scripts
 
 import Prelude
 
-import Data.Argonaut (class DecodeJson, class EncodeJson)
+import Contract.Prelude (Either(..))
```
ngua, line 13: We should avoid importing `Contract` modules internally

ngua (changes requested):
The changes to the instances look fine, but a few things we need to do before merging this:
- remove the `Contract` imports
- update the examples to account for the changes to the JSON instances

src/Plutus/Types/CurrencySymbol.purs
bladyjoker: IDE does this for some reason like this, terrible I agree, fixing

src/Types/Scripts.purs
bladyjoker: How did that get here, I guess IDE being very smoort again

ngua (approved):

PR #407, opened 2022-05-12
ogmios based protocol params

### LINKED ISSUE #346: Ogmios based protocol params?
CTL server could set the current protocol parameters by querying the Ogmios instance for it. Now they are set from a static file which requires keeping it consistent or errors will happen. These errors might not be immediately recognized as PP mismatch errors which cost us dev time.

Luis-omega:
fixes #346


Luis-omega:
The only remain issues are : 

- The `server/config/pparams.json` has `"extraPraosEntropy": null` but ogmios returns `"extraEntropy": "neutral"`, should I leave it as `null` or use the function `makePraosNonce` over the string `"neutral"`?

- The server tests now have a collection of fixed parameters to avoid depending on living instances of `Ogmios` and `nodes`. This version of parameters doesn't have a version of the `costModels` dictionary, so we must be careful to delete them whenever test parameters are updated.

flake.nix
```diff
@@ -251,7 +251,7 @@
                   "-c"
                   ''
                     ${pkgs.ogmios}/bin/ogmios \
-                      --host 0.0.0.0 \
+                      --host ogmios \
```
ngua, line 254: Is this going to work when making websocket requests outside of the runtime (e.g. from the browser)? I'm not sure if it will

server/src/Ogmios/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+import Cardano.Api (
+  AnyPlutusScriptVersion (AnyPlutusScriptVersion),
+  CostModel,
+  ExecutionUnitPrices (ExecutionUnitPrices),
+  ExecutionUnits,
+  Lovelace (Lovelace),
+  PlutusScriptVersion (PlutusScriptV1),
+ )
+import Cardano.Api.Shelley (ProtocolParameters (ProtocolParameters))
+
```
ngua, line 12: Why the blank lines between imports? Fourmolu will sort the imports if you remove them

server/src/Ogmios/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+decodeProtocolParameters :: ByteString -> Either [Text.Text] ProtocolParameters
+decodeProtocolParameters response =
+  let value :: Maybe Aeson.Value
+      value = Aeson.decode response
+   in case parseValue parseResult <$> value of
+        Just (Right params) -> Right params
+        Just (Left e) -> Left $ displayError id e
+        _ -> Left ["Fail at converting ogmios response to cardano format"]
+
+--      wrapped :: Maybe ProtocolParametersWrapper
```
ngua, line 110: Can you remove the commented-out code?

server/src/Ogmios/Parser.hs
```diff
@@ -0,0 +1,146 @@
+module Ogmios.Parser (decodeProtocolParameters) where
```
ngua, line 1: This applies to most of this module: since you're just using `Text` as the error type, why bother with `Data.Aeson.BetterErrors`? We just use Aeson directly everywhere else and now we need to depend on parsec as well

server/src/Ogmios/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+    (,) <$> parseNatural "major" <*> parseNatural "minor"
+
+parseNatural :: Text.Text -> Parse e Natural
+parseNatural strKey =
+  naturalFromInteger <$> key strKey (asIntegral :: Parse e Integer)
+
+parseLovelace :: Text.Text -> Parse e Lovelace
+parseLovelace strKey =
+  key strKey $
+    Lovelace <$> (asIntegral :: Parse e Integer)
```
ngua, line 48: Is the type annotation necessary?

server/src/Ogmios/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+type LocalParser a = Parsec.Parsec String () a
+
+nonZeroDigit :: LocalParser Char
+nonZeroDigit = Parsec.Char.satisfy (\c -> (c /= '0') && isDigit c)
+
+nonZeroInteger :: LocalParser Integer
+nonZeroInteger = do
+  headDigit <- nonZeroDigit
+  remains <- Parsec.many Parsec.Char.digit
+  return $ read (headDigit : remains)
```
ngua, line 125: We favor `pure` over `return` everywhere else. There are several instances of this in this module and others

server/src/Ogmios/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+  do
+    Parsec.Char.spaces
+    numerator <- haskellInteger
+    Parsec.Char.spaces >> Parsec.Char.char '/' >> Parsec.Char.spaces
+    denominator <- haskellInteger
+    return $ numerator % denominator
+
+rationalParser :: String -> Either Text.Text Rational
+rationalParser s =
+  case Parsec.runParser rational () "ogmios.json" s of
```
ngua, line 144: `either` would work here instead of an explicit case expression

server/src/Ogmios/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+  Map.singleton (AnyPlutusScriptVersion PlutusScriptV1)
+    <$> key "plutus:v1" fromAesonParser
+
+parseResult :: Parse Text.Text ProtocolParameters
+parseResult =
+  key "result" $
+    ProtocolParameters
+      <$> parseVersion
+      <*> parseRational "decentralizationParameter"
+      -- TODO : How to parse the value `neutral` from ogmios?
```
ngua, line 75: That should be addressed in this PR I think

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,91 @@
...
+  makeRequest,
+  tryQueryUntilZero,
+  RequestOption (Query, Acquire),
+  defaultServerParameters,
+  ServerParameters (..),
+  QueryOption (..),
+) where
+
+import Control.Exception.Base (SomeException, try)
+
```
ngua, line 14: More blank lines

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,91 @@
...
+data QueryOption = CurrentProtocolParameters
+
+data RequestOption
+  = Query QueryOption
+  | -- | TODO : add point parameter to Acquire
+    Acquire
+
+data ServerParameters = ServerParameters
+  { getHost :: String
```
ngua, line 34: I think the `get` prefix on these fields isn't necessary

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,91 @@
...
+import Data.Text (Text)
+import Network.Socket (withSocketsDo)
+import Network.WebSockets (ClientApp)
+import Network.WebSockets qualified as WebSockets
+import System.IO (hFlush, stdout)
+import System.Time.Extra qualified as Time.Extra
+
+--------------------------------------------------------------------------------
+
+data QueryOption = CurrentProtocolParameters
```
ngua, line 26: I don't think we need a `QueryOption` type, since we won't be implementing any other Ogmios queries

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,91 @@
...
+getQueryName :: QueryOption -> Text
+getQueryName CurrentProtocolParameters = "currentProtocolParameters"
+
+buildRequestJSON :: RequestOption -> Text
+buildRequestJSON (Query CurrentProtocolParameters) =
+  "{ \"type\": \"jsonwsp/request\", \"version\": \"1.0\", \"servicename\": \"ogmios\","
+    <> "\"methodname\": \"Query\", \"args\": { \"query\":\""
+    <> getQueryName CurrentProtocolParameters
+    <> "\"} }"
+buildRequestJSON _ = error "Unsuported Ogmios request"
```
ngua, line 57: Following my suggestion for removing `QueryOption` would make this a total function

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,91 @@
...
+    , getPort = 1337
+    , getPath = "/"
+    }
+
+getQueryName :: QueryOption -> Text
+getQueryName CurrentProtocolParameters = "currentProtocolParameters"
+
+buildRequestJSON :: RequestOption -> Text
+buildRequestJSON (Query CurrentProtocolParameters) =
+  "{ \"type\": \"jsonwsp/request\", \"version\": \"1.0\", \"servicename\": \"ogmios\","
```
ngua, line 53: It might be easier to read and less error-prone to use Aeson to construct the object

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,91 @@
...
+  WebSockets.sendTextData
+    conn
+    requestJSON
+  msg <- WebSockets.receiveData conn
+  WebSockets.sendClose conn ("" :: Text)
+  return msg
+
+makeRequest :: ServerParameters -> RequestOption -> IO ByteString
+makeRequest ServerParameters {..} option =
+  withSocketsDo $
```
ngua, line 71: `withSocketsDo` isn't required (and only ever was on Windows IIRC)

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,91 @@
...
+tryQueryUntilZero query remainAttempts =
+  if remainAttempts <= 0
+    then do
+      return $ Left "Error trying to connect to Ogmios"
+    else do
+      msgOrError <- first show <$> try @SomeException query
+      case msgOrError of
+        Right msg -> return $ Right msg
+        Left e ->
+          do
```
ngua, line 85: Strange formatting

server/src/Ogmios/Parser.hs
```diff
@@ -0,0 +1,146 @@
...
+
+zeroInteger :: LocalParser Integer
+zeroInteger = read <$> many1 (Parsec.Char.satisfy (== '0'))
+
+haskellInteger :: LocalParser Integer
+haskellInteger = nonZeroInteger Parsec.<|> zeroInteger
+
+rational :: LocalParser Rational
+rational =
+  do
```
ngua, line 135: Putting `do` on a new line looks a bit odd to me

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,91 @@
...
+import System.Time.Extra qualified as Time.Extra
+
+--------------------------------------------------------------------------------
+
+data QueryOption = CurrentProtocolParameters
+
+data RequestOption
+  = Query QueryOption
+  | -- | TODO : add point parameter to Acquire
+    Acquire
```
ngua, line 31: Do we need this?

server/src/Types.hs
```diff
@@ -36,27 +39,27 @@ import Data.Aeson (FromJSON, ToJSON (toJSON))
...
-import Plutus.V1.Ledger.Scripts qualified as Ledger.Scripts
 import Servant (FromHttpApiData, QueryParam', Required, ToHttpApiData)
 import Servant.Docs qualified as Docs
 import Text.Read (readMaybe)
 import Utils (tshow)
 
+import Ogmios.Parser (decodeProtocolParameters)
+import Ogmios.Query qualified
+
+-- import Paths_ctl_server (getDataFileName)
```
ngua, line 61: You can just delete this

server/src/Types.hs
```diff
@@ -36,27 +39,27 @@ import Data.Aeson (FromJSON, ToJSON (toJSON))
...
 import Data.Text.Encoding qualified as Text.Encoding
 import GHC.Generics (Generic)
 import Network.Wai.Handler.Warp (Port)
-import Paths_ctl_server (getDataFileName)
-import Plutus.V1.Ledger.Api qualified as Ledger
-import Plutus.V1.Ledger.Scripts qualified as Ledger.Scripts
 import Servant (FromHttpApiData, QueryParam', Required, ToHttpApiData)
 import Servant.Docs qualified as Docs
 import Text.Read (readMaybe)
 import Utils (tshow)
```
ngua, line 57: More blank lines

server/src/Types.hs
```diff
@@ -77,14 +80,42 @@ data ServerOptions = ServerOptions
...
+   in do
+        eitherResponse <-
+          Ogmios.Query.tryQueryUntilZero
+            queryProtocolParams
+            maxConnectionAttempts
+        case eitherResponse of
+          Right response ->
+            case decodeProtocolParameters response of
+              Right params ->
+                (return . Right . Env serverOptions) params
```
ngua, line 111: It looks strange to me to use function composition and then use parens for application (instead of `$`)

server/test/Main.hs
```diff
@@ -2,6 +2,18 @@ module Main (main) where
...
 import Cardano.Api qualified as C
+import Cardano.Api.Shelley (
+  ExecutionUnitPrices (
+    ExecutionUnitPrices,
+    priceExecutionMemory,
+    priceExecutionSteps
+  ),
+  ExecutionUnits (ExecutionUnits, executionMemory, executionSteps),
+  Lovelace (Lovelace),
+  ProtocolParameters (..),
```
ngua, line 13: Explicit constructor imports

server/test/Main.hs
```diff
@@ -2,6 +2,18 @@ module Main (main) where
...
+  --  makePraosNonce,
```
ngua, line 14: Please remove

server/test/Main.hs
```diff
@@ -39,18 +50,27 @@ import Types (
...
+    networkId,
+    nodeSocket,
+    ogmiosHost,
+    ogmiosPort,
+    port
+  ),
   WitnessCount (WitnessCount),
-  newEnvIO,
   unsafeDecode,
  )
```
ngua, line 70: Blank lines

server/test/Main.hs
```diff
@@ -305,3 +329,60 @@ fullyAppliedScript =
...
+loadFile =
+  do
+    contents <- ByteString.readFile "test/ogmios.json"
+    return $ decodeProtocolParameters contents
+
+testParser :: Spec
+testParser =
+  it "Testing parser of ogmios parameters" $ do
+    value <- loadFile
+    shouldBe value (Right fixedProtocolParameters)
```
ngua, line 388: The convention we've used so far in the tests is to make `shouldBe` infix

ngua (changes requested):
I have some comments mostly regarding style and formatting below

I would have preferred using Aeson directly to parse the values, just for consistency with the rest of the server code, but you don't need to change that if it would take a long time to rewrite

The biggest potential issue is the change to Ogmios' host in the runtime configuration. I want to confirm that websockets are able to connect to Ogmios from outside the runtime system (e.g. from the browser)

flake.nix
Luis-omega: All I can say is that I can run Pkh2Pkh example.

I got help from @klntsky in this change.

server/src/Ogmios/Parser.hs
Luis-omega: I separate Imports by category : 

- Cardano/Plutus related
- Base Haskell modules 
- Rest of the packages 
- If some package has a lot of imports, then it has its own section 

Those can be in random order but are always separated. This helps me to maintain imports and looks better (to me) than having all of them together. I can change those for consistency or change the rest of the server to this format.

server/src/Ogmios/Parser.hs
Luis-omega: Those are two separate things coming from the same problem: 

- The original parser was written in pure Aeson but it was crashing silently, so, I introduced `BetterErrors` to get some feedback for debugging. 
- There were various problems, in particular, Ogmios returns `Rationals` as strings `"a/b"` and the `FromJSON` instance for `Rational` didn't expect this. That's the reason we now import `Parsec`, so we can parse these rationals.

At first, I thought of writing the `Rational` parser ad-hoc but that would mean writing part of the core of `Parsec` in the module anyways. 

I can come back to pure Aeson with some ease once I solve the `how to parse praosNonce` issue.

server/src/Ogmios/Parser.hs
Luis-omega: It's the result of code refactors, previously it was needed, but right now it isn't.

Fixed, thanks.

server/src/Ogmios/Parser.hs
Luis-omega: We want to return an `Either` here to propagate the error to `Aeson.BetterError`, so, maybe you were thinking of `first` from the `Bifunctor` instance of `Either`?

server/src/Ogmios/Query.hs
Luis-omega: fixed.

server/src/Ogmios/Query.hs
Luis-omega: I wrote them like that just in case, but now that I have confirmation that those won't be needed I'm deleting them.

server/src/Ogmios/Parser.hs
Luis-omega: This is a clash of styles since it looks odd to me to put it in the same line :sweat_smile:

server/src/Types.hs
Luis-omega: same as discussed above.

server/test/Main.hs
Luis-omega: Our code `STANDARS.md` is out of sync with https://github.com/mlabs-haskell/styleguide. 

In particular, we have now : 

```
An exception is granted if the number of fields or constructors is large; then, wildcard exports MAY be used:

-- This is fine if Baz has a lot of constructors or fields
module Foo.Bar (
  Baz(..)
  ) where 
```

And `ProcolParameters` definitely has a lot of fields.

Luis-omega: Right, I remembered wrong the part of the [style guide](https://github.com/mlabs-haskell/styleguide) that talks about imports, I thought the `exports` parts also plied here.

Fixed.

server/test/Main.hs
Luis-omega: fixed.

server/test/Main.hs
Luis-omega: It was part of the `extra entropy` issue, fixed.

server/src/Ogmios/Parser.hs
Luis-omega: yep, I stated it at the top but now is solved.

fixed.

server/src/Ogmios/Parser.hs
Luis-omega: The problem with `extra entropy` is solved now, so, I can convert the parser back to pure Aeson, it would take me around 4 hours, should I do it?

flake.nix
ngua: OK, thanks for confirming

server/src/Ogmios/Parser.hs
ngua: Just for consistency with the rest of the codebase (both Haskell and Purescript), I'd prefer if you don't separate them

server/src/Ogmios/Parser.hs
ngua: I would prefer using Aeson directly, for a few reasons:
- there's no extra dependencies
- possible contributors are less likely to know Parsec and aeson-better-errors, while we can assume that they will be comfortable using Aeson parsers
- we can remove the custom Parsec parsers
- Aeson uses attoparsec, which is more efficient than parsec

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,74 @@
...
+  if remainAttempts <= 0
+    then do
+      pure $ Left "Error trying to connect to Ogmios"
+    else do
+      msgOrError <- first show <$> try @SomeException query
+      case msgOrError of
+        Right msg -> pure $ Right msg
+        Left e ->
+          do
+            putStrLn $ "Error : " <> show e
```
ngua, line 69: Isn't this already a `String`? Since you're doing `first show`

server/src/Ogmios/Parser.hs
ngua: I understand that style is more or less arbitrary, but I would prefer that you make it consistent with the rest of the codebase

server/src/Ogmios/Parser.hs
ngua: That said, if you really don't want to rewrite the parser or if it will take too much time, you can leave it as-is

server/test/Main.hs
```diff
@@ -61,25 +107,26 @@ serverSpec = do
...
   describe "Api.Handlers.hashData" hashDataSpec
+  describe "Ogmios.Parser " testParser
 
 applyArgsSpec :: Spec
 applyArgsSpec = around withTestApp $ do
   clientEnv <- setupClientEnv
 
   context "POST apply-args" $ do
-    it "returns the same script when called without args" $ \port -> do
+    it "pures the same script when called without args" $ \port -> do
```
klntsky, line 117: A nice purescript/haskell joke here, 

I hope it was done intentionally and is not a result of bad string replace.

server/src/Ogmios/Query.hs
Luis-omega: thanks! 

This is done now

server/src/Ogmios/Parser.hs
Luis-omega: fixed.

server/src/Ogmios/Query.hs
Luis-omega: fixed

server/test/Main.hs
Luis-omega: gotcha, i forgot `return` is a valid English word.

server/src/Ogmios/Query.hs
Luis-omega: you are right, fixed.

server/src/Ogmios/Parser.hs
Luis-omega: Converted to Aeson. 

Fixed.

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,102 @@
...
+  defaultServerParameters,
+  ServerParameters (..),
+) where
+
+import Control.Exception.Base (SomeException, try)
+import Data.Aeson qualified as Aeson
+import Data.Bifunctor (first)
+import Data.ByteString.Lazy (ByteString)
+import Data.Text (Text)
+import GHC.Generics
```
ngua, line 16: We always use explicit imports, unless it's something like `Prelude` in Purescript (which always needs to be imported)

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,102 @@
...
+app conn = do
+  WebSockets.sendTextData
+    conn
+    $ Aeson.encode defaultRequest
+  msg <- WebSockets.receiveData conn
+  WebSockets.sendClose conn ("" :: Text)
+  pure msg
+
+makeRequest :: ServerParameters -> IO ByteString
+makeRequest ServerParameters {..} =
```
ngua, line 84: I'd avoid `RecordWildCards` in general. `NamedFieldPuns` is OK

server/src/Types.hs
```diff
@@ -83,14 +85,39 @@ data ServerOptions = ServerOptions
...
+  , ogmiosPort :: Int
   }
   deriving stock (Generic)
 
 newEnvIO :: ServerOptions -> IO (Either String Env)
-newEnvIO serverOptions =
-  getDataFileName "config/pparams.json"
-    >>= Aeson.eitherDecodeFileStrict @Shelley.ProtocolParameters
-    <&> second (Env serverOptions)
+newEnvIO serverOptions@ServerOptions {..} =
```
ngua, line 94: Same thing regarding `RecordWildCards`

server/src/Types.hs
```diff
@@ -83,14 +85,39 @@ data ServerOptions = ServerOptions
...
+   in do
+        eitherResponse <-
+          Ogmios.Query.tryQueryUntilZero
+            queryProtocolParams
+            maxConnectionAttempts
+        case eitherResponse of
+          Right response ->
+            case decodeProtocolParameters response of
+              Right params ->
+                return . Right . Env serverOptions $ params
```
ngua, line 114: `pure`

ngua (commented):
Looks good for the most part, I have a few remaining comments on style below

I believe you're off until Monday so I can take care of this and merge it

ngua (commented):
I think the parsing should be simplified (we don't actually need attoparsec directly, Aeson uses it internally). I will take care of that as well

server/src/Ogmios/Query.hs
```diff
@@ -0,0 +1,102 @@
+{-# LANGUAGE OverloadedStrings #-}
```
ngua, line 1: These pragmas are redundant

ngua (approved):
I would've also liked to adjust how you're parsing the cost models as well -- I'd say it's better to avoid directly manipulating the underlying `HashMap` (and even if you wanted to use this approach, `update` or `alter` would be better IMO)

Also, in general, we should avoid partial functions like `read`. `readMaybe` should always be preferred

PR #408, opened 2022-05-12
Use ogmios-datum-cache reflection

### LINKED ISSUE #363: Use ogmios-datum-cache reflection in our code
Currently, to match ODC responses to requests we don't use reflection that is implemented in https://github.com/mlabs-haskell/ogmios-datum-cache/pull/13 because it was not available earlier.

We need to address that.

This may make it possible to unify some code for ogmios and ogmios-datum-cache requests.

klntsky:
Closes #363 

- Update Ogmios to the recent revision
- Fix incorrect handling of Json parse errors for ODC (cancel Aff instead of letting it freeze).
- Remove deleted endpoints (`datumFilter*`)
- (Partially) unify ODC and Ogmios WS code
- Do not rely on response type to match responses with requests for ODC, use reflection instead

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,7 +11,6 @@
     ogmios.url = "github:mlabs-haskell/ogmios/c4f896bf32ad066be8edd8681ee11e4ab059be7f";
     ogmios-datum-cache = {
       url = "github:mlabs-haskell/ogmios-datum-cache";
-      flake = true;
     };
     # so named because we also need a different version of the repo below
     # in the server inputs and we use this one just for the `cardano-cli`
--- a/flake.nix
+++ b/flake.nix
@@ -9,9 +9,7 @@
 
     # for the purescript project
     ogmios.url = "github:mlabs-haskell/ogmios/c4f896bf32ad066be8edd8681ee11e4ab059be7f";
-    ogmios-datum-cache = {
-      url = "github:mlabs-haskell/ogmios-datum-cache";
-    };
+    ogmios-datum-cache.url = "github:mlabs-haskell/ogmios-datum-cache";
     # so named because we also need a different version of the repo below
     # in the server inputs and we use this one just for the `cardano-cli`
     # executables
@@ -194,7 +192,7 @@
               };
               autoStart = true;
               startFromLast = false;
-              filter = "{ \"const\": true }";
+              filter = builtins.toJSON { const = true; };
             };
           }
         }:
@@ -303,7 +301,7 @@
                   useHostStore = true;
                   ports = [ (bindPort datumCache.port) ];
                   restart = "on-failure";
-                  depends_on = [ "postgres" ];
+                  depends_on = [ "postgres" "ogmios" ];
                   command = [
                     "${pkgs.bash}/bin/sh"
                     "-c"
--- a/flake.nix
+++ b/flake.nix
@@ -189,8 +189,8 @@
               ];
             blockFetcher = {
               firstBlock = {
-                slot = 44366242;
-                id = "d2a4249fe3d0607535daa26caf12a38da2233586bc51e79ed0b3a36170471bf5";
+                slot = 54066900;
+                id = "6eb2542a85f375d5fd6cbc1c768707b0e9fe8be85b7b1dd42a85017a70d2623d";
               };
               autoStart = true;
               startFromLast = false;
--- a/flake.nix
+++ b/flake.nix
@@ -11,7 +11,7 @@
     ogmios.url = "github:mlabs-haskell/ogmios/c4f896bf32ad066be8edd8681ee11e4ab059be7f";
     ogmios-datum-cache = {
       url = "github:mlabs-haskell/ogmios-datum-cache";
-      flake = false;
+      flake = true;
     };
     # so named because we also need a different version of the repo below
     # in the server inputs and we use this one just for the `cardano-cli`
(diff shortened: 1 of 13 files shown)


flake.nix
```diff
@@ -11,7 +11,7 @@
     ogmios.url = "github:mlabs-haskell/ogmios/c4f896bf32ad066be8edd8681ee11e4ab059be7f";
     ogmios-datum-cache = {
       url = "github:mlabs-haskell/ogmios-datum-cache";
-      flake = false;
+      flake = true;
```
ngua, line 14: You can just delete this line, it's `true` unless otherwise specified for flake inputs

src/Types/Transaction.purs
```diff
@@ -875,6 +876,13 @@ instance Show DataHash where
 -- To help with people copying & pasting code from Haskell to Purescript
 type DatumHash = DataHash
 
+newtype BlockId = BlockId ByteArray
```
ngua, line 879: Isn't this the same as the `BlockHeaderHash`? Maybe we could deduplicate the the types (the other one is a `String`, but perhaps should be a `ByteArray`). We'd just need to adjust the `getChainTip` return type

test/AffInterface.purs
```diff
@@ -40,6 +52,46 @@ suite = do
...
+      testOgmiosDatumCacheGetDatumByHash
+    test "Can process GetDatumsByHashes" do
+      testOgmiosDatumCacheGetDatumsByHashes
+    test "Fetcher works" do
+      testOgmiosDatumCacheFetcher
+
+testOgmiosDatumCacheGetDatumByHash :: Aff Unit
+testOgmiosDatumCacheGetDatumByHash =
+  traceQueryConfig >>= flip runQueryM do
+    -- Use this to trigger block fetching in order to actually get the datum:
```
ngua, line 66: See my larger comment about the flake. We can just automate this part, I think this will be error-prone (IME a lot of users have found the explicit block fetching unclear)

src/QueryM.purs
```diff
@@ -848,7 +789,42 @@ mkOgmiosRequest jsonWspCall getLs inp = do
...
+            cont $ lmap dispatchErrorToError result
+        )
+      respLs.addRequest id sBody
+      _wsSend ws (logString config.logLevel Debug) sBody
+      pure $ Canceler $ \err -> do
+        liftEffect $ respLs.removeMessageListener id
+        liftEffect $ throwError $ err
+  liftAff $ makeAff $ affFunc
+
+-- TODO after ogmios-datum-cache implements reflection this could be generalized to make request for the cache as well
```
ngua, line 801: I thought using ODC's new reflection mechanism was the rationale for this PR?

ngua (changes requested):
Looks good, I have some minor suggestions below

My request for changes is because the ODC configuration in the `buildCtlRuntime` needs to be reworked pretty significantly. Unfortunately I can't point out the exact places since they're not part of the PR's changes, but we need to do at least the following:
- change the `datumCache` attrset argument. There are new configuration options
- update how the `config.toml` gets generated accordingly, as I believe it will now be invalid
- set up autofetching (for some reasonable starting block, the one we used during Seabug deployment should work for now) so that we don't need to manually start the block fetcher. This is one of ODC's best new improvements I think

src/QueryM.purs
klntsky: Yep, a copy-paste fail on my side.

test/AffInterface.purs
klntsky: Do you want to remove the start/stop calls from our code?

klntsky:
> set up autofetching (for some reasonable starting block, the one we used during Seabug deployment should work for now) so that we don't need to manually start the block fetcher. This is one of ODC's best new improvements I think

I changed the parameters to values provided by you in slack 1 month ago.

ngua:
@klntsky Those values should be OK, they still sync pretty quickly IME

test/AffInterface.purs
ngua: Hmmm, good question. I suggested removing these because we should use the automated fetching that ODC now supports. I think users (and frankly even CTL developers who haven't had to deal with ODC yet) won't know to run the block fetcher. Maybe mention the autofetching in the comment?

Also, we could move `testOgmiosDatumCacheFetcher` to before any other tests, then the block fetcher will have been stopped even if is automatically run from the runtime config

At any rate, I'll leave the ultimate decision up to you

flake.nix
```diff
@@ -190,10 +187,14 @@
...
-              slot = 44366242;
-              id = "d2a4249fe3d0607535daa26caf12a38da2233586bc51e79ed0b3a36170471bf5";
+            blockFetcher = {
+              firstBlock = {
+                slot = 54066900;
+                id = "6eb2542a85f375d5fd6cbc1c768707b0e9fe8be85b7b1dd42a85017a70d2623d";
+              };
+              autoStart = true;
+              startFromLast = false;
+              filter = "{ \"const\": true }";
```
ngua, line 197: Is this because of the typo in ODC? I just merged Tomasz' fix if so

ngua (approved):
If you bump the ODC input to latest master, you can delete the default filter workaround and we can merge this

flake.nix
ngua: Also it would be nicer to use `builtins.toJSON` for this I think

ngua:
I'd like to merge this ASAP so I took care of the last remaining things

PR #413, opened 2022-05-17
Upgrade Ogmios to latest master

ngua:
Adds `aarch64-{linux|darwin}` support for Ogmios

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,7 +8,7 @@
     };
 
     # for the purescript project
-    ogmios.url = "github:mlabs-haskell/ogmios/c4f896bf32ad066be8edd8681ee11e4ab059be7f";
+    ogmios.url = "github:mlabs-haskell/ogmios";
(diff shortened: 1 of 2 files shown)


firefrorefiddle (approved):
Looks good!

flake.nix
```diff
@@ -8,7 +8,7 @@
     };
 
     # for the purescript project
-    ogmios.url = "github:mlabs-haskell/ogmios/c4f896bf32ad066be8edd8681ee11e4ab059be7f";
+    ogmios.url = "github:mlabs-haskell/ogmios";
```
klntsky, line 11: Why did you add revision here and not to `locked.rev`?

ngua: I had to specify a specific commit previously as I hadn't merged the changes into Ogmios master yet. You can also do that with the `rev` attribute, but that requires the long form:
```
ogmios = {
  type = "github";
  owner = "...";
  repo = "...";
  rev = "...";
};
```

`locked.rev` is in the lockfile, which you shouldn't edit by hand

PR #417, opened 2022-05-17
purescript-bridge compatibility changes (cont of #405)

bladyjoker:
c/p from #405 ...
Hi CTL folks, 

This PR contains fixes/changes to the `ToData/FromData` machinery that are necessary for the correct operation of `purescript-bridge`. The changed modules should be well-documented, but the general idea is: 

When one extracts a `RowList` from a `Row`, the `RowList` is always lexicographically sorted. Plutus Data records are _not_ lexicographically sorted, so the existing generic To/FromData code in CTL does not work correctly for Records right now. Because both Haskell and PureScript permit (for some reason...seriously who does this?!) data types such as this gem

``` haskell
data Foo = FooA {foo :: String, fooo :: Int}
                | FooB {fooo :: Bool, foo :: Bool}
```
We have to capture the relation between a constructor and its record fields (if any exist) in order to ensure on-chain compatibility. In order to integrate with most of the existing generic To/FromData code, this has to be done at the type level, so that's where we do it. 

I will attempt to pull/merge CTL's `master` branch ASAP to resolve conflicts, but that may take me a day (this is by far the most complex merge I've attempted, I will probably screw it up at least once). In the meantime, please review this PR and suggest relevant/necessary changes. 

Edit: 

Actually there are way fewer conflicts than I expected. Will come back to this first thing tomorrow and sort that out.


bladyjoker:
@gnumonik please apply the commenting style as instructed in https://github.com/Plutonomicon/cardano-transaction-lib/pull/405#discussion_r873518985

src/Plutus/Types/DataSchema.purs
```diff
@@ -0,0 +1,262 @@
...
+       ( "G0" := PNil @@ Z
+       :+ "G1" := PNil @@ (S Z)
+       :+ "G2" :+ PNil @@ (S (S Z))
+       :+ PNil)
+
+   The sole purpose of the inner list is to ensure correct translation to and from Plutus Data for record types
+   when using the generic functions in ToData/FromData. Since GType's constructors do not have record arguments,
+   no additional information is needed in the inner lists.
+-}
+-- | See the source (DataSchema.purs) for a detailed explanation of how this works / what it is for / examples
```
ngua, line 156: I think this should be removed in favor of the non-doc comment above. We have searchable docs now, so I'd like to avoid directing users towards reading the source

src/Plutus/Types/DataSchema.purs
```diff
@@ -0,0 +1,262 @@
...
+-- | Type synonym used to represent MkField as a type operator.
+type MkField_
+  :: forall (f :: Type -> Type) (k :: Type). Symbol -> f k -> Field k
+type MkField_ lbl ixty = MkField lbl ixty
+
+infixr 8 type MkField_ as :=
+
+{- <<< Constraints used to provide compile-time validation of a PSchema >>> -}
+
+-- TODO: See if we can optimize this. Slows down compilation considerably for complex types.
```
ngua, line 208: We should add an issue for this and put the URL here

src/FromData.purs
```diff
@@ -2,190 +2,232 @@ module FromData
   ( FromDataError(..)
   , class FromData
   , class FromDataArgs
+
```
ngua, line 5: Extra blank line here

ngua (commented):
Just a few comments

src/Plutus/Types/DataSchema.purs
gnumonik: Done

src/Plutus/Types/DataSchema.purs
gnumonik: Done (also added references to the issue to the other Uniqueness classes)

src/FromData.purs
gnumonik: fixed

ngua (approved):

PR #415, opened 2022-05-17
Use Plutus-style data for Address and Value in Contract.*

jy14898:
Closes #298

PR #421, opened 2022-05-18
Viet/298 contract plutus types

### LINKED ISSUE #298: Convert types in public API to Plutus-style representations
We can use CSL/Cardano-style types internally, but our `Contract` interface should take and return `Value`s, `Address`es, etc... that are consistent with Plutus. For example, the `Contract.Utxos.UtxosAt` should take a Plutus-style address

Currently blocked on #193

vvtran:
- Closes #298
- Taken from https://github.com/Plutonomicon/cardano-transaction-lib/pull/415

Things I've implemented:
- Added `AssocMap`, `Credential` to `Contract` name space.
- Re-exported helpers from `Plutus.Types.Address` into `Contract.Address`
- Refined current helpers from `Contract.Address` to use Plutus `Address` in function domain/range - so we have two sets of helpers, there are comments explaining this.
- Created `Plutus.Types.Value.Coin` to mirror the CSL type, this is analagous to `Ada` from Plutus. E.g. functions like fee calculation will use the Plutus `Coin` in `Contract`
- Migrated much of `Types.Transaction` over to `Cardano.Types.Transaction` as part of an incremental refactor. Keeping common types in `Types.Transaction` e.g. `TransactionInput`, `TransactionHash`.
- Created `Plutus.Types.TransactionUnspentOutput` to mirror the CSL type, since it's output of a few functions from the `Contract` namespace, even if these functions are unlikely to be used.
- For `TransactionOutput`, I initially created the following in the common types `Types.Transaction`
```
newtype TxOutput address value = TxOutput
  { address :: address
  , amount :: value
  , dataHash :: Maybe DataHash
  }
```
then had
```
newtype TransactionOutput = TransactionOutput (TxOutput Cardano.Address Cardano.Value)
newtype TransactionOutput = TransactionOutput (TxOutput Plutus.Address Plutus.Value)
```
But decided against it because of the extra unwrapping, furthermore, if we used the same pattern for another common type like `TransactionUnspentOutput` (which uses `TransactionOutput`), then unwrapping would get annoying. Another example is, `Utxo`/`UtxoM`. If we used the above technique, we'd have to unwrap twice for `UtxoM`, then twice for `TransactionOutput`.
- Created `Plutus.Types.Transaction.Utxo(M)` to mirror the CSL type, which is now the output of `UtxosAt` in `Contract`, using the Plutus `TransactionOutput` of course.
- Removed many type aliases, I initially added these as I thought they would be helpful for Plutus people moving over, now with the two `TransactionOutput`s, they are probably confusing. In particular `type DatumHash = DataHash`. `TxOutRef`, and `TxOut` are gone. Everything is `TransactionOutput`, `TransactionInput` and `DataHash` now (whether from `Plutus.Types` or `Cardano.Types`.
- Functions that return the type `UnattachedUnbalancedTx` in `Contract` are still exposing the CSL `Redeemer` type but this is needed so that redeemers can be reattached.
- `Transaction` and many of its components are still exposed in `Contract`, the user generally isn't expected to use these and they are CSL related but I think they're fine to keep - we don't need Plutus equivalents do we? This includes various lenses like `_mint` `_collateral` which will return a CSL `Value`. Again, the user probably shouldn't play with these fields directly.
- I tried to make sure all constraints and lookups are using Plutus style, then converting to CSL internally. Balancing still uses CSL of course.
- `Contract.PlutusData`  now re-exports helpers from Generic `FromData`, `ToData` and `Plutus.Types.DataSchema`


src/Cardano/Types/Transaction.purs
```diff
@@ -0,0 +1,799 @@
...
+derive instance Generic UtxoM _
+derive instance Newtype UtxoM _
+derive newtype instance Eq UtxoM
+
+instance Show UtxoM where
+  show = genericShow
+
+type Utxo = Map TransactionInput TransactionOutput
+
+-- Option<Certificates>,
```
ngua, line 770: It's really time for these comments to go haha. They've somehow survived since the beginning

ngua (approved):
I'm going to remove the old comments (see my previous comment above) and then I think we can merge this. Something may have been overlooked, but we can open another PR if needed

vvtran:
I've already removed them. I also re-exported with the new ToData/FromData stuff.

ngua:
Oh great, I thought you hadn't started working yet

ngua:
We also need to update the Seabug metadata, but we're going to move that to a new repo at any rate and can do it there

PR #422, opened 2022-05-18
buildCtlRuntime: better config management

Synthetica9:
Previously, it was not possible to override only part of a config item (for example, only setting the port for postgres). This PR fixes that, and as a bonus, also allows you to refer to the final value of config items.

A few examples to clear this up:

```
# Correctly changes the dbConnectionString
buildCtlRuntime ({ postgres.port = 9001; postgres.user = "asdf"; })
# Refer back to the final config:
buildCtlRuntime (final: { node.port = postgres.port + 1; })
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Here is an example that uses the overlay to launch runtime services:
       # The configuration for the CTL runtime, which will be passed to the
       # expression that builds the JSON file used by Arion. This value can be
       # shared between `buildCtlRuntime` and `launchCtlRuntime`, as shown below
-      runtimeConfig = {
+      runtimeConfig = final: with final; {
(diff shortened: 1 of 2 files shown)


ngua (commented):
LGTM, thanks! Could you also update the example in the readme that shows how to use `buildCtlRuntime` please?

klntsky (approved):

Synthetica9:
Like this, or do you want more doc?

klntsky:
@Synthetica9 Would be cool if both ways to call `buildCtlRuntime` were mentioned (with attrset and with function as argument).

ngua:
I'm going to close this and re-open it from a branch on the main repo

PR #423, opened 2022-05-18
Add generic Hashing endpoint to server and the corresponding Frontend functions. 

MangoIV:
### Adds 
- [x] api changes that replace `blake2bhash` with `plutusHash`
- [x] fix documentation for server endpoint
- [x] corresponding frontend changes
- [x] clean up everything, add the wrapper functions in `Contract.Crypto`
- [x] add integration test 
- [x] actually run the integration test - succeeds

### Issues 
- none


server/README.md
```diff
@@ -61,9 +61,13 @@ The server executable can be built with `nix build .#cardano-trasaction-lib-serv
     - `application/json`
 
 - Bytes to hash as hexadecimal string (`application/json;charset=utf-8`, `application/json`):
+- Hashing method from PlutusTx to hash with; one of `Sha2_256`, `Sha3_256`, `Blake2b_256`
 
 ```javascript
-"666f6f"
+{
+  "bytesToHash" : "68656C6C6F20776F726C64"
```
ngua, line 68: I think the field names are a bit verbose. Why not just `bytes` and `method`?

server/README.md
```diff
@@ -77,9 +81,13 @@ The server executable can be built with `nix build .#cardano-trasaction-lib-serv
     - `application/json`
 
 - Hash bytes are returned as hexidecimal string (`application/json;charset=utf-8`, `application/json`):
+- Hashing Method used; one of `Sha2_256`, `Sha3_256`, `Blake2b_256`
 
 ```javascript
-"b8fe9f7f6255a6fa08f668ab632a8d081ad87983c77cd274e48ce450f0b349fd"
+{
+  "hash":"644bcc7e564373040999aac89e7622f3ca71fba1d972fd94a31c3bfbf24e3938",
+  "methodUsed":"Sha3_256"
```
ngua, line 89: Same here

server/README.md
```diff
@@ -61,9 +61,13 @@ The server executable can be built with `nix build .#cardano-trasaction-lib-serv
     - `application/json`
 
 - Bytes to hash as hexadecimal string (`application/json;charset=utf-8`, `application/json`):
+- Hashing method from PlutusTx to hash with; one of `Sha2_256`, `Sha3_256`, `Blake2b_256`
```
ngua, line 64: I think requiring the method to be capitalized might lead to user errors (if for some reason they don't want to use the client call in `QueryM.Crypto`)

server/src/Types.hs
```diff
@@ -153,16 +156,37 @@ newtype AppliedScript = AppliedScript Ledger.Script
...
+newtype HashedBytes = HashedBytes ByteString
   deriving stock (Show, Generic)
   deriving newtype (Eq)
   deriving (FromJSON, ToJSON) via JsonHexString
 
+data HashMethod = Blake2b_256
+                | Sha2_256
+                | Sha3_256
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (FromJSON, ToJSON)
```
ngua, line 180: I think using something like: 
```haskell
instance FromJSON HashMethod where
  parseJSON = 
    genericParseJSON 
      defaultOptions 
      { constructorTagModifier = fmap toLower }
```
would make it nicer to use the endpoint (so that users can write the hash method in a more natural way)

src/Contract/Crypto.purs
```diff
@@ -0,0 +1,44 @@
...
+  , datumHash
+  , plutusHash
+  , HashedData(..)
+  , HashMethod(..)
+  ) where
+
+import Contract.Monad (Contract, wrapContract)
+import Data.Either (Either)
+import Data.Maybe (Maybe)
+import Prelude ((<<<))
```
ngua, line 13: We usually import `Prelude` without explicit imports, and place it above every other import separated by a blank line

src/Contract/Crypto.purs
```diff
@@ -0,0 +1,44 @@
...
+import QueryM (ClientError)
+import QueryM.Crypto as Crypt
```
ngua, line 15: Is this a typo? Otherwise I think `Crypto` should be the qualified name

src/QueryM/Crypto.purs
```diff
@@ -0,0 +1,135 @@
...
+import Prelude
+
+import Affjax as Affjax
+import Affjax.RequestBody as Affjax.RequestBody
+import Affjax.ResponseFormat as Affjax.ResponseFormat
+import Control.Monad.Reader.Trans (asks)
+import Data.Argonaut
+  ( class DecodeJson
+  , Json
+  , JsonDecodeError(..)
```
ngua, line 19: We also use explicit constructor imports

src/QueryM/Crypto.purs
```diff
@@ -0,0 +1,135 @@
...
+  , decodeJson
+  , caseJsonString
+  , (:=)
+  , (~>)
+  , encodeJson
+  )
+import Data.Bifunctor (bimap)
+import Data.Either (Either(..), hush, note, either)
```
ngua, line 27: Here too

src/QueryM/Crypto.purs
```diff
@@ -0,0 +1,135 @@
...
+import Data.Generic.Rep (class Generic)
+import Data.Maybe (Maybe(..), maybe')
```
ngua, line 29: And here

src/QueryM/Crypto.purs
```diff
@@ -0,0 +1,135 @@
...
+import Data.Newtype (class Newtype, unwrap, wrap)
+import Data.Show.Generic (genericShow)
+import Effect.Aff.Class (liftAff)
+import Effect.Class (liftEffect)
+import Effect.Exception (throw)
+import QueryM (ClientError(..), QueryM, mkServerEndpointUrl, scriptToJson)
```
ngua, line 35: And here

test/Crypto.purs
```diff
@@ -0,0 +1,58 @@
...
+          ( \(Tuple meth goal) -> do
+              qbs <- throwMaybe (error "bytestring is not hex format") origBS
+              res <- throwMaybe (error "hashing method was not successfull") =<<
+                plutusHash meth qbs
+              shouldEqual goal $ byteArrayToHex res
+          )
+          hashtups
+    )
+
+throwMaybe :: forall a m e. MonadThrow e m => e -> Maybe a -> m a
```
ngua, line 43: You can also use `Test.Utils.errMaybe` and skip using `error` explicitly above

On a general note: we always use kind signatures on type parameters, even when this can be clearly inferred (so we would even explicitly add `(m :: Type -> Type)` even though it's obvious from the `MonadThrow` constraint)

MangoIV:
Hi, thank you for your review. Can you tell me how I run the integration tests? Does it automatically pick them up if I have the #ctl-env thingy running? (Had some permission issues with that)

server/README.md
MangoIV: I was expecting naming clashes when not using specific names, and indeed, `method` from `BytesToHash` would conflict with `method` from `HashBytesRequest`. I could rename it by writing custom `FromJSON` instances, do you want me to do so?

src/QueryM/Crypto.purs
MangoIV: I have to configure my lsp properly at some point, it always introduces those :/

test/Crypto.purs
MangoIV: Thank you, noted.

klntsky:
> Can you tell me how I run the integration tests?

@MangoIV you need to:

- execute `nix run -L .#ctl-runtime`
- execute `spago test -m Test.Integration`

server/src/Api/Handlers.hs
```diff
@@ -178,6 +176,14 @@ evalTxExecutionUnits cbor =
               , exUnitsSteps = C.executionSteps exUnits
               }
 
+plutusHash :: HashBytesRequest -> AppM ByteStringHash
+plutusHash (HashBytesRequest meth (BytesToHash hs)) = 
+  let hf = case meth of
```
ngua, line 181: I think possibly hasn't been formatted? In the future, you can run `make format` from the `server` directory (after entering the Haskell dev shell with `nix develop .#hsDevShell`). You can also run `make format` from the root of the repo to format the Purescript code

We have a CI check for formatting at any rate, so if you'll be contributing a lot I'll give you write access and you'll see if it fails due to formatting

server/README.md
ngua: Do you mean record name clashes? You could also just use `DuplicateRecordFields`, I have no problem with that (and in fact would be my first choice)

server/README.md
```diff
@@ -61,9 +61,13 @@ The server executable can be built with `nix build .#cardano-trasaction-lib-serv
     - `application/json`
 
 - Bytes to hash as hexadecimal string (`application/json;charset=utf-8`, `application/json`):
+- Hashing method from PlutusTx to hash with; one of `sha2_256`, `sha3_256`, `blake2b_256`
```
ngua, line 64: Did you write this manually? We generate the docs using `servant-docs`, so it will get lost the next time we update it

ngua (commented):
I'm going to close this and open a new PR from a branch on the main repo

server/src/Api/Handlers.hs
MangoIV: makes sense, I was running make format always from the root, thanks

server/README.md
MangoIV: makes sense, I didn't know that, yeah, I wrote those manually, didn't know

PR #424, opened 2022-05-18
Fix inaccuracies in fee calculation

errfrom:
Context: https://github.com/Plutonomicon/cardano-transaction-lib/issues/123

The solution was to take into account both script integrity hash and the fee value when estimating fees. 
Now all examples are successfully submitted with accurate fees.

### DIFF
--- a/server/src/Api/Handlers.hs
+++ b/server/src/Api/Handlers.hs
@@ -19,6 +19,7 @@ import Cardano.Ledger.Alonzo.Data qualified as Data
 import Cardano.Ledger.Alonzo.Language (Language (PlutusV1))
 import Cardano.Ledger.Alonzo.Tx qualified as Tx
 import Cardano.Ledger.Alonzo.TxWitness qualified as TxWitness
+import Cardano.Ledger.Core qualified as Ledger (TxBody)
 import Cardano.Ledger.Crypto (StandardCrypto)
 import Cardano.Ledger.Mary.Value qualified as Value
 import Cardano.Ledger.SafeHash qualified as SafeHash
@@ -32,11 +33,15 @@ import Data.ByteString (ByteString)
 import Data.ByteString.Base16 qualified as Base16
 import Data.ByteString.Lazy qualified as BL
 import Data.Kind (Type)
+import Data.List qualified as List (find)
 import Data.Map qualified as Map
+import Data.Maybe (fromJust)
+import Data.Maybe.Strict (StrictMaybe)
 import Data.Proxy (Proxy (Proxy))
 import Data.Set qualified as Set
 import Data.Text.Encoding qualified as Text.Encoding
 import Data.Traversable (for)
+import Math.NumberTheory.Logarithms qualified as Math (integerLog2)
 import Plutus.V1.Ledger.Scripts qualified as Ledger.Scripts
 import PlutusTx.Builtins qualified as PlutusTx
 import Types (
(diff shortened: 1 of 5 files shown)


src/BalanceTx.purs
```diff
@@ -532,13 +532,11 @@ logTx msg utxos (Transaction { body: body'@(TxBody body) }) =
     , "Fees: " <> show body.fee
     ]
 
--- Nami provides a 5 Ada collateral that we should add the tx before balancing
+-- Nami provides a 5 Ada collateral that we should add the tx before balancing.
```
adamczykm, line 535:
```
- -- Nami provides a 5 Ada collateral that we should add the tx before balancing.
+ -- Nami provides a 5 Ada collateral that we should add to the tx before balancing.
```

?

adamczykm (approved):
:tada:

src/BalanceTx.purs
errfrom: @adamczykm Hm, just added a period to the end of the docstring :)

klntsky (approved):

server/src/Api/Handlers.hs
```diff
@@ -75,13 +80,59 @@ import Types (
...
+
 estimateTxFees :: WitnessCount -> Cbor -> AppM Fee
 estimateTxFees (WitnessCount numWits) cbor = do
-  decoded <-
-    either (throwM . CborDecode) pure $
-      decodeCborTx cbor
-  pparams <- asks protocolParams
-  pure . Fee $ estimateFee pparams numWits decoded
+  tx <- decodeCborTx cbor & either (throwM . CborDecode) pure
+  case tx of
```
ngua, line 90: Why not just use a pattern directly in the binding? `C.Tx txBody' keyWits <- decode...`

server/src/Api/Handlers.hs
```diff
@@ -75,13 +80,59 @@ import Types (
...
+              -- be able to handle a possible integer overflow
+              feeBytes =
+                bytesNeeded (fee + bytesNeeded fee * feePerByte)
+           in fee + feeBytes * feePerByte
+      where
+        -- Calculates the number of bytes that a given integer will take
+        -- when CBOR encoded.
+        bytesNeeded :: Integer -> Integer
+        bytesNeeded n =
+          let predicate = (>= (succ $ Math.integerLog2 n `div` 8))
```
ngua, line 119: I think we can avoid `$` by moving the parens

src/BalanceTx.purs
```diff
@@ -581,9 +578,9 @@ returnAdaChange changeAddr utxos unattachedTx =
           ReturnAdaChangeImpossibleError
             "Not enough Input Ada to cover output and fees after prebalance."
             Impossible
-      EQ ->
-        except $ Right $ unattachedTx' # _transaction' .~
-          wrap tx { body = wrap txBody { fee = wrap fees } }
+      EQ -> do
```
ngua, line 581: Why add `do` here?

server/src/Api/Handlers.hs
```diff
@@ -239,14 +271,61 @@ queryUtxos txInputs =
...
+-- Set script integrity hash
+--------------------------------------------------------------------------------
+
+setScriptIntegrityHash ::
+  Shelley.ProtocolParameters ->
+  Shelley.TxBody C.AlonzoEra ->
+  Shelley.TxBody C.AlonzoEra
+setScriptIntegrityHash pparams txBodyAlonzo =
+  case txBodyAlonzo of
+    Shelley.ShelleyTxBody _e txBody scripts scriptData _a _v ->
```
ngua, line 284: I think the `_` prefixes might be a bit misleading, since that's conventionally used to indicate a discarded value, but they're actually used below

server/src/Api/Handlers.hs
ngua: Hmmm apparently that's a fallible pattern, interesting

ngua (commented):
Nice work! I will take care of the style things below now to merge this immediately

ngua (approved):

PR #425, opened 2022-05-19
fix encodeJson minting policy

vvtran:
- small fix, although I think we're switching to Aeson soon


vvtran:
Closing as this will be included in Aeson switch.

PR #427, opened 2022-05-19
Use purescript-aeson instead of argonaut

### LINKED ISSUE #429: Use purescript-aeson instead of argonaut

klntsky:
CC @bladyjoker 

Closes #429


bladyjoker:
Hi @klntsky I need to look carefully into this as it conflicts in some way with #419 . This is on our critical path for a project so there will be lots of movement arounds Aeson instances until we get them aligned with Haskell.

src/Cardano/Types/Value.purs
```diff
@@ -47,16 +47,16 @@ module Cardano.Types.Value
...
-import Data.Argonaut
-  ( class DecodeJson
-  , class EncodeJson
-  , JsonDecodeError(TypeMismatch)
-  , caseJsonObject
-  , encodeJson
+import Aeson
+  ( class DecodeAeson
+  , class EncodeAeson
+  , JsonDecodeError(..)
```
ngua, line 53: Explicit constructor imports. It seems like everywhere `JsonDecodeError` is being imported like this. I'm not going to add a comment in each module, but please fix each of them

ngua (changes requested):
Looks fine except for the comment below

src/Serialization/Address.purs
```diff
@@ -128,15 +121,8 @@ instance Semigroup Slot where
...
-instance DecodeJson Slot where
-  decodeJson = jsonToAeson >>>
-    caseAesonObject
-      (Left $ TypeMismatch "Expected object")
-      (flip getField "getSlot" >=> Slot >>> pure)
-
-instance EncodeJson Slot where
-  encodeJson (Slot uint) = fromNumber (UInt.toNumber uint)
+instance EncodeAeson Slot where
+  encodeAeson' (Slot uint) = encodeAeson' (UInt.toNumber uint)
```
bladyjoker, line 125: Why did you remove decoding for Slot?

src/Types/Natural.purs
```diff
@@ -56,10 +50,6 @@ instance DecodeAeson Natural where
           fromBigInt bi
       )
 
--- This is needed for `ApplyArgs`.
-instance DecodeJson Natural where
-  decodeJson = decodeAeson <<< jsonToAeson
-
```
bladyjoker, line 62: Why remove?

src/Types/Rational.purs
```diff
@@ -43,11 +43,11 @@ derive newtype instance Semiring Rational
 derive newtype instance Ring Rational
 derive newtype instance CommutativeRing Rational
 
-instance EncodeJson Rational where
-  encodeJson r = encodeJson
+instance EncodeAeson Rational where
```
bladyjoker, line 46: We need a DecodeAeson for this as well...

src/Types/Scripts.purs
```diff
@@ -67,12 +67,12 @@ derive instance Newtype MintingPolicy _
...
-  decodeJson = decodeJsonHelper "getMintingPolicy" MintingPolicy
+instance DecodeAeson MintingPolicy where
+  decodeAeson = decodeAesonHelper "getMintingPolicy" MintingPolicy
 
-instance EncodeJson MintingPolicy where
-  encodeJson (MintingPolicy script) = encodeJson
-    { "getValidator": encodeJson script }
+instance EncodeAeson MintingPolicy where
+  encodeAeson' (MintingPolicy script) = do
+    encodeAeson' { "getMintingPolicy": script }
```
bladyjoker, line 76: Thanks and sorry for messing this up!

bladyjoker (dismissed):
Thanks for doing this @klntsky 

I noticed that sometimes you remove a Decode instance and I'm not sure why. One other thing that I noticed is that we sometimes have only an Encode but no Decode, if you can add those it would be great, but if you don't have time I'm already doing that in #419

src/Serialization/Address.purs
klntsky: I confirmed this with @vvtran , it's leftover code.

src/Types/Natural.purs
klntsky: Same as above

src/Types/Rational.purs
ngua: Don't we have one?

src/QueryM/JsonWsp.purs
```diff
@@ -17,20 +17,17 @@ import Prelude
 
 import Aeson
   ( class DecodeAeson
+  , class EncodeAeson
   , Aeson
+  , JsonDecodeError(..)
```
ngua, line 22: I think you missed this one

ngua (commented):
Small comment

src/Contract/Scripts.purs
```diff
@@ -15,6 +15,11 @@ module Contract.Scripts
   , module TypesScripts
   ) where
 
+import Aeson (class DecodeAeson)
+import Address
+  ( enterpriseAddressMintingPolicyHash
+  , enterpriseAddressScriptHash
+  )
```
vvtran, line 22: I think we can remove this import completely, this has been refactored into
```
import Contract.Address
  ( enterpriseAddressScriptHash
  , enterpriseAddressStakeValidatorHash
  , enterpriseAddressValidatorHash
  , scriptHashAddress -- Directly uses Plutus `Address`
  , toValidatorHash -- Directly uses Plutus `Address`
  , typedValidatorBaseAddress
  , typedValidatorEnterpriseAddress
  , validatorHashBaseAddress
  , validatorHashEnterpriseAddress
  ) as Address
```
below and I don't think we should include `enterpriseAddressMintingPolicyHash`

src/Plutus/Types/Address.purs
```diff
@@ -70,7 +70,7 @@ pubKeyHashAddress :: PaymentPubKeyHash -> Maybe StakePubKeyHash -> Address
 pubKeyHashAddress (PaymentPubKeyHash pkh) skh = wrap
   { addressCredential: PubKeyCredential pkh
   , addressStakingCredential:
-      map (StakingHash <<< PubKeyCredential <<< unwrap) skh
+      map (StakingHash <<< PubKeyCredential <<< wrap <<< unwrap) skh
```
vvtran, line 73: I think this just just be `unwrap`.  See my comment on `StakePubKeyHash`

src/Types/ByteArray.purs
```diff
@@ -51,17 +55,14 @@ instance Semigroup ByteArray where
...
-    (note (Json.UnexpectedValue j) <<< hexToByteArray)
-    j
-
-instance EncodeJson ByteArray where
-  encodeJson ba = fromString (byteArrayToHex ba)
-
 instance DecodeAeson ByteArray where
-  decodeAeson = decodeAesonViaJson
+  decodeAeson = caseAesonString (Left err)
+    (note err <<< hexToByteArray)
```
vvtran, line 60: We previously had separate errors for this but merging into one now?

src/Types/PubKeyHash.purs
```diff
@@ -115,14 +114,14 @@ instance Show PaymentPubKeyHash where
...
+instance DecodeAeson PaymentPubKeyHash where
+  decodeAeson = caseAesonObject
     (Left $ TypeMismatch "Expected object")
     ( flip getField "unPaymentPubKeyHash" >=>
-        decodeJson >>> map PaymentPubKeyHash
+        decodeAeson >>> map PaymentPubKeyHash
     )
 
-newtype StakePubKeyHash = StakePubKeyHash PubKeyHash
+newtype StakePubKeyHash = StakePubKeyHash Ed25519KeyHash
```
vvtran, line 124: I think this a bad merge, this was recently changed as a newtype over `PubKeyHash` in line with Plutus, `master` still has this change. See https://github.com/Plutonomicon/cardano-transaction-lib/pull/432

src/Types/Scripts.purs
```diff
@@ -41,22 +41,22 @@ derive instance Generic PlutusScript _
...
-derive newtype instance EncodeJson PlutusScript
+derive newtype instance DecodeAeson PlutusScript
+derive newtype instance EncodeAeson PlutusScript
 
 instance Show PlutusScript where
   show = genericShow
 
-decodeJsonHelper
+decodeAesonHelper
   ∷ ∀ (a ∷ Type) (b :: Type)
```
vvtran, line 51: This is not you but are we okay with unicode? We should probably stick to one since we're generally trying to standardise our code @ngua

vvtran (commented):
Comments, main thing is the `StakePubKeyHash`

src/Types/Scripts.purs
vvtran: There's also usage in `Cardano.Types.Value.union`

bladyjoker: I think this is lsp being smart :/

ngua: Oh, I would prefer not to use unicode. Not sure where that was introduced

ngua (approved):

ngua:
@vvtran We can fix the unicode thing in another PR

Issue #243, opened 2022-03-29
Correct implementation to get an `Address` from `ScriptHash`


vvtran:
Related to https://github.com/Plutonomicon/cardano-browser-tx/issues/242 I'll copy and paste some of the issue over:

`PublicKey` is a newtype wrapper over Bech32String which is an alias of `String`, the issue is the string can be different lengths so some further newtype wrappers over String would probably make sense (see above issue)

For example, we can use `ed25519KeyHashFromBech32 :: Bech32String -> Maybe Ed25519KeyHash` on an input like `"addr_vkh1zuctrdcq6ctd29242w8g84nlz0q38t2lnv3zzfcrfqktx0c9tzp"` but not
`"addr1qyc0kwu98x23ufhsxjgs5k3h7gktn8v5682qna5amwh2juguztcrc8hjay66es67ctn0jmr9plfmlw37je2s2px4xdssgvxerq"`. However,
we can use `addressBech32 :: Address -> Bech32String` on the latter.

I believe this is because the former being an enterprise address (no staking credentials), and hence shorter. Feel free to correct my understanding on this.

The issue here is we can't get an `Address` from a `ScriptHash` (i.e. `ValidatorHash` contained in a `TypedValidator a`). Our currently implementation 
`typedValidatorAddress :: forall (a :: Type). NetworkId -> TypedValidator a -> Address` goes via the `BaseAddress`. This seemingly always produces  an address with stake credentials. Perhaps we need to reconsider and go via `EnterpriseAddress` which is an `Address` without staking credentials.

This mirrors the `Maybe` type in the definition of Plutus' `Address`
https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger/html/Ledger-Address.html#t:Address

vvtran:
Can we close this with our enterprise address implementation?

vvtran:
Or perhaps switch to "bug" as we would need to do cherry-pick commits over to master.

ngua:
Changed to a "bug" since we need to apply the changes to master as well

ngua:
Fixed on Seabug branch, will merge into master soon

PR #435, opened 2022-05-20
Hashing (was #243)

ngua:
Replay of #243 from @MangoIV


MangoIV:
Thank you! I see you already resolved the missing issues yourself, sorry for not being quick enough.

MangoIV:
I already integrated the changes from #419 but I don't have permission to push to this branch. (just wanted to leave a note here so that work doesn't need to be duplicated)

ngua:
I'm going to close this since it seems to be part of #419

MangoIV:
I mean we could've integrated it to decrease  the amount of stuff  to review in #419 but your call :)

PR #437, opened 2022-05-23
Add key wallet for testing

### LINKED ISSUE #140: Key wallet support for testing
The `Wallet` type should be extended with support for public and private keys

This should be integrated with a testing framework that has the following properties:

- Throws an exception if run in the browser
- If run in node, uses `fs` to write transactions generated during testing to disk
- The transactions should (probably?) be CBOR-encoded
- They can then be provided to Plutip

jy14898:
Closes #140


src/Wallet.purs
```diff
@@ -32,21 +43,95 @@ import Effect.Exception (throw)
...
+  selectCollateral :: Utxo -> Maybe TransactionUnspentOutput
+  selectCollateral utxos = unwrap <<< unwrap <$> flip
+    foldMapWithIndex
+    utxos
+    \input output ->
+      let
+        txuo = AdaOut $ TransactionUnspentOutput { input, output }
+        Value ada (NonAdaAsset naa) = _value txuo
+        onlyAda = all (all ((==) zero)) naa
+        bigAda = ada >= mkCoin 5_000_000
```
klntsky, line 102: Is hardcoding 5 ADA value safe?

jy14898: Not sure, but given Nami has the same collateral it's probably safe. Currently collateral is 150% of the tx fee, and I don't think fees can get that high?

jy14898: But also, this method can be overridden if necessary for big transactions

klntsky: No I mean is it determined by protocol parameters? We need to support private testnets too.

jy14898: Oh true. Once again it can be overridden, alternatively we can make it a parameter mkKeyWallet

jy14898: We could try and supply a collateral estimate to this function based on an estimated fee and scaling by collateral scale?

src/Serialization.purs
```diff
@@ -571,6 +593,9 @@ convertWithdrawals mp =
 publicKeyFromBech32 :: Bech32String -> Maybe PublicKey
 publicKeyFromBech32 = _publicKeyFromBech32 maybeFfiHelper
 
+privateKeyFromBytes :: ByteArray -> Effect (Maybe PrivateKey)
```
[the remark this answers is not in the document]
jy14898: `_privateKeyFromBytes` really does work on bytes, not cbor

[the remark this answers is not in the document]
ngua: We have `RawBytes` for that

ngua:
Shouldn't we change the base branch to `develop`?

examples/Pkh2PkhKeyWallet.purs
```diff
@@ -0,0 +1,136 @@
...
+  , submit
+  )
+import Contract.TxConstraints as Constraints
+import Contract.Value as Value
+import Control.Monad.Error.Class (catchError, liftMaybe)
+import Control.Monad.Logger.Trans (runLoggerT)
+import Control.Monad.Reader (runReaderT)
+import Data.BigInt as BigInt
+import Data.Log.Formatter.Pretty (prettyFormatter)
+import Data.Log.Level (LogLevel(..))
```
ngua, line 36: Explicit constructor imports

src/BalanceTx.purs
```diff
@@ -107,14 +111,12 @@ import Serialization.Address
...
 import Types.Natural (toBigInt) as Natural
 import Types.ScriptLookups (UnattachedUnbalancedTx(UnattachedUnbalancedTx))
 import Types.Transaction (DataHash, TransactionInput)
-import Types.UnbalancedTransaction
-  ( UnbalancedTx(UnbalancedTx)
-  , _transaction
-  )
-import TxOutput (utxoIndexToUtxo)
+import Types.UnbalancedTransaction (UnbalancedTx(UnbalancedTx), _transaction)
+import Wallet (Wallet(..))
```
ngua, line 119: Explicit imports

src/BalanceTx.purs
```diff
@@ -403,11 +405,18 @@ balanceTx unattachedTx@(UnattachedUnbalancedTx { unbalancedTx: t }) = do
...
+    wallet <- asks _.wallet
     utxos <- ExceptT $ utxosAt ownAddr <#>
       (note (UtxosAtError' CouldNotGetUtxos) >>> map unwrap)
-
+    collateral <- case wallet of
+      Just (Nami _) -> Just <$>
+        ( ExceptT $ getWalletCollateral <#>
+            note (GetWalletCollateralError' CouldNotGetNamiCollateral)
+        )
+      -- TODO(jy14898) Supply with fee estimate?
```
ngua, line 416: Please open issues corresponding to the TODOs (and add their URLs in the comment)

examples/Pkh2PkhKeyWallet.purs
```diff
@@ -0,0 +1,136 @@
...
+
+    launchAff_ $ flip catchError
+      ( \e -> liftEffect $ logError e
+          *> log' "crimson" ("[ERROR] " <> message e)
+          *> unlock
+      )
+      do
+        priv <- (liftMaybe $ error "Failed to parse private key") =<<
+          ( liftEffect $ map join $ traverse privateKeyFromBytes $
+              hexToByteArray input.private_key
```
ngua, line 103: This is a very minor thing, but I'd like to avoid snake_case

examples/Pkh2PkhKeyWallet.purs
```diff
@@ -0,0 +1,136 @@
...
+          )
+        pkh <- liftMaybe (error "Failed to parse public key hash") $
+          ed25519KeyHashFromBech32 input.to_pkh
```
ngua, line 106: Same here

examples/Pkh2PkhKeyWallet.purs
```diff
@@ -0,0 +1,136 @@
...
+import Data.Log.Message (Message)
+import Effect.Exception (Error, error, message)
+import QueryM (QueryConfig)
+import Serialization (privateKeyFromBytes)
+import Serialization.Hash (ed25519KeyHashFromBech32)
+import Types.ByteArray (hexToByteArray)
+import Wallet (mkKeyWallet)
+
+type Form =
+  { private_key :: String
```
ngua, line 46: See my comments about snake_case below

ngua (changes requested):
I think it would be helpful to document how to generate the key pair

src/Wallet.purs
```diff
@@ -15,12 +17,21 @@ import Cardano.Types.Transaction
   , TransactionWitnessSet(TransactionWitnessSet)
   , Vkey(Vkey)
   , Vkeywitness(Vkeywitness)
+  , Utxo
+  , _vkeys
+  , TransactionOutput(..)
```
klntsky, line 22: > `(..)`

src/Wallet.purs
klntsky: Let's ignore this for now, I will create an issue to investigate.

src/Wallet.purs
jy14898: Thanks, HLS keeps inserting .. rather than the name when I use an unimported constructor. I'll have to see if I can configure it

ngua:
Also @jy14898 could you add something to the changelog please (under `## Unreleased`)?

ngua (approved):
Looks good. Just a heads up: with #533 merged, you'll need to fix `utxosAt` again (sorry)

PR #438, opened 2022-05-23
Make cardano-configurations configurable

klntsky:
This change makes it more convenient to change cardano network used by CTL

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,16 @@ Furthermore, CTL exposes an `overlay` from its flake. You can use this in the Ni
 
 We have recenly set up a small scaffolding repository for projects wishing to adopt CTL: https://github.com/mlabs-haskell/ctl-scaffold. More documentation and resources will be added soon to the repo
 
+### Changing network configurations.
+
+CTL supports using networks other than testnet to provide development environment.
+
+First you need to specify alternative way of getting the network configuration.
+
+[cardano-configurations](https://github.com/input-output-hk/cardano-configurations) repo provides configs for `mainnet`, `staging` and a few other networks. If you want one of these, simply replace `network` parameter in `buildCtlRuntime`.
+
+Providing custom repo with similar structure is also possible: edit `flake.nix` and change `inputs.cardano-configurations` providing the URL of your fork. You can use `path:/path/to/repo` as URL as well. After that, edit `network` parameter (it should point to a subdirectory of `./network/` in your fork).
+
 ## Architecture
 
 CTL is directly inspired by the Plutus Application Backend (PAB). Unlike PAB, however, CTL is a library and not a standalone process. Over the course of CTL's development, several questions have been raised as to how best create PAB-as-a-library:
--- a/README.md
+++ b/README.md
@@ -445,6 +445,8 @@ You can specify your own fork of `cardano-configurations` like this:
 inputs.cardano-transaction-lib.inputs.cardano-configurations.follows = "...";
 ```
 
+When changing networks, make sure that `network.magic` is correctly synchronized with value in config (see `protocolConsts.protocolMagic` in `byron.json`).
+
 ## Architecture
 
 CTL is directly inspired by the Plutus Application Backend (PAB). Unlike PAB, however, CTL is a library and not a standalone process. Over the course of CTL's development, several questions have been raised as to how best create PAB-as-a-library:
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Here is an example that uses the overlay to launch runtime services:
       # shared between `buildCtlRuntime` and `launchCtlRuntime`, as shown below
       #
       # You can refer to the final configuration value by passing a function
-      # that takes a single arugment. Alternately, you can pass an attrset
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -425,6 +425,16 @@ Furthermore, CTL exposes an `overlay` from its flake. You can use this in the Ni
...
 
+### Changing network configurations.
+
+CTL supports using networks other than testnet to provide development environment.
+
+First you need to specify alternative way of getting the network configuration.
+
+[cardano-configurations](https://github.com/input-output-hk/cardano-configurations) repo provides configs for `mainnet`, `staging` and a few other networks. If you want one of these, simply replace `network` parameter in `buildCtlRuntime`.
+
+Providing custom repo with similar structure is also possible: edit `flake.nix` and change `inputs.cardano-configurations` providing the URL of your fork. You can use `path:/path/to/repo` as URL as well. After that, edit `network` parameter (it should point to a subdirectory of `./network/` in your fork).
```
ngua, line 436: I think it would be better to illustrate something like 
```
inputs.cardano-transaction-lib.inputs.cardano-configurations.follows = "...";
```

since users can't edit our flake unless they fork the repo or point it towards a local copy, which would be very inconvenient

Also, I think it might be better to use `git+file://...` as an example for a local directory as IIRC `path:...` requires the input to be a flake 

See my other comments below about adding a configurations directory as an argument for building the CTL runtime

flake.nix
```diff
@@ -206,8 +208,10 @@
             fix (final: recursiveUpdate
               (defaultConfig final)
               (if isFunction extraConfig then extraConfig final else extraConfig));
-          nodeDbVol = "node-db";
-          nodeIpcVol = "node-ipc";
+          # Network subdirectory in cardano-configurations repo provided above
+          network = "testnet";
```
ngua, line 212: Move this line to `defaultConfig`, you wouldn't be able to override it here. Then you need to add `inherit (config) network` or replace all of the occurrences with the qualified name (i.e. `config.network`)

flake.nix
```diff
@@ -16,7 +16,9 @@
     cardano-node-exe = {
       url = "github:input-output-hk/cardano-node/ea8b632820db5546b22430bbb5ed8db4a2fef7dd";
     };
+    # Repository with network parameters
     cardano-configurations = {
+      # Override with "path:/path/to/cardano-configurations";
```
ngua, line 21: I can't put this comment where it should really go as the lines aren't part of the changes yet, but the easiest way to override `cardano-configurations` is to just add it to `defaultConfig` below:

```
defaultConfig = final: with final; {
  inherit (inputs) cardano-configurations;
  network = "testnet";
  node = { port = 3001; };
  # ...
};

```

ngua (dismissed):
Good idea to make it configurable. A few things should be adjusted though

README.md
```diff
@@ -425,6 +429,24 @@ Furthermore, CTL exposes an `overlay` from its flake. You can use this in the Ni
...
+
+First you need to specify an alternative way of getting the network configuration.
+
+[cardano-configurations](https://github.com/input-output-hk/cardano-configurations) repo provides configs for `mainnet`, `staging` and a few other networks.
+
+Which of the network config directories is chosen is determined by `network.name` parameter of `buildCtlRuntime`
+
+You can specify your own fork of `cardano-configurations` like this:
+
+```nix
```
ngua, line 444: I'd remove `nix` here, since it's a fragment of something larger and not a valid Nix expression on its own, GH will render the markdown with highlighted syntax errors, like this:

```nix
foo = "...";
```

flake.nix
```diff
@@ -279,7 +286,7 @@
                     ${server}/bin/ctl-server \
                       --port ${toString ctlServer.port} \
                       --node-socket ${nodeSocketPath} \
-                      --network-id 1097911063
+                      --network-id ${toString config.network.magic}
```
ngua, line 289: I don't think this will work for mainnet. You need to check if the `magic` attribute is set or if it's `null`. Also we need to add something to the documentation indicating that `mainnet` should not have `magic` set

ngua: To elaborate: the correct flag for ctl-server is `--network-id mainnet`

README.md
```diff
@@ -425,6 +429,24 @@ Furthermore, CTL exposes an `overlay` from its flake. You can use this in the Ni
...
+inputs.cardano-transaction-lib.inputs.cardano-configurations.follows = "...";
+```
+
+When changing networks, make sure that `network.magic` is correctly synchronized with value in config (see `protocolConsts.protocolMagic` in `byron.json`).
```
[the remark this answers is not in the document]
klntsky: No, because we need to distinguish between mainnet and other networks (see comment by @ngua about ctl-server `--network-id`), and without full knowledge about all protocol parameters we can't determine that - network magic is just a nonce by itself.

adamczykm (commented):
Lgtm

klntsky:
@adamczykm could you please approve via button

adamczykm (approved):
Yes

PR #439, opened 2022-05-23
Remove additional thunk from publicKeyFromBech32 foreign implementation

jy14898:
~~The foreign import was incorrectly not in Effect, this PR makes the necessary changes to the import and the impacted code~~
EDIT: Removing the thunk
https://github.com/Plutonomicon/cardano-transaction-lib/blob/d3ce766c4718c6924356c952dec0f1c70be92a89/src/Serialization.js#L97

### DIFF
--- a/src/Serialization.js
+++ b/src/Serialization.js
@@ -94,7 +94,7 @@ exports.addVkeywitness = witnesses => witness => () =>
 exports.newVkeyFromPublicKey = public_key => () =>
     lib.Vkey.new(public_key);
 
-exports._publicKeyFromBech32 = maybe => bech32 => () => {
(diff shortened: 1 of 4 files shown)


klntsky:
Thanks for catching this!

But I think we can remove Effect entirely here and stick to just `Maybe`. Since we catch all exceptions, it would be safe.

jy14898:
@klntsky I thought that too, but wasn't sure if the reference is mutable in the wasm side of csl

klntsky:
Technically it's mutable: there's a `free` method. However, we don't use it from PS, so we can treat it as pure.

We do that already for some other types.

adamczykm (approved):

Issue #410, opened 2022-05-13
Construct exact POSIXTimeRange given to TxInfo


t1lde:
There are cases where we need to be able to obtain the *exact* POSIXTimeRange that gets passed to scripts with TxInfo - for example if the there are time-dependent values on Datums.

The problem is that there is a round-trip conversion from 

- the POSIXTimeRange provided to `MustValidateIn`
- Conversion to `Slot`
- conversion to `TransactionValiditySlot`
- conversion *back* to POSIXTimeRange, when the ledger constructs TxInfo

See pg 35 of the [Alonzo spec](https://hydra.iohk.io/job/Cardano/cardano-ledger/specs.alonzo-ledger/latest/download-by-type/doc-pdf/alonzo-changes) for part of the conversion. This may also require some reverse-engineering of the cardano-ledger

Saizan:
I did some digging and `cardano-ledger` implements `slotToPOSIXTime` [1], using `epochInfoSlotToUTCTime`
https://github.com/input-output-hk/cardano-ledger/blob/2acff66e84d63a81de904e1c0de70208ff1819ea/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs#L186-L203

That function in turn is provided by `cardano-slotting`, with addtional parameters about the protocol and epoch:
https://github.com/input-output-hk/cardano-base/blob/8fe904d629194b1fbaaf2d0a4e0ccd17052e9103/slotting/src/Cardano/Slotting/EpochInfo/API.hs#L80-L85

Most things can be directly queried from the node, and the `EpochInfo` can be calculated from `EraHistory` which can also be calculated.
https://github.com/input-output-hk/cardano-node/blob/91a11ba26707d2de1c72abfe4a15f59f84e8e349/cardano-cli/src/Cardano/CLI/Shelley/Run/Query.hs#L1313-L1316

Saizan:
Ultimately it comes down to properly using the outputs of the following `slotToWallclock` query. In that module we also have `wallclockToSlot` which would be useful in the other direction.

https://github.com/input-output-hk/ouroboros-network/blob/bd9e5653647c3489567e02789b0ec5b75c726db2/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Qry.hs#L461-L481

PR #440, opened 2022-05-23
Add `eraSummaries` Ogmios Query

vvtran:
- Initial part of https://github.com/Plutonomicon/cardano-transaction-lib/issues/410
- Add `eraSummaries` endpoint of Ogmios query to current API https://ogmios.dev/api/ with a test.
- Got rid of replicate json wsp request and response types in `QueryM.DatumCacheWsp` EDIT: this was wrong and corrected in https://github.com/Plutonomicon/cardano-transaction-lib/pull/447
- A few stylistic changes.
- This doesn't have an issue but I noticed `Slot` usage in `ChainTip` query could be problematic, our `Slot` is `UInt` but the one returned by Ogmios can potentially cause precision errors, so I've switched to `BigInt`. Any thoughts on this would be good.

### DIFF
--- a/src/QueryM/Ogmios.purs
+++ b/src/QueryM/Ogmios.purs
@@ -318,7 +318,7 @@ newtype UtxoR = UtxoR UtxoQueryResult
 derive newtype instance Show UtxoR
 
 instance DecodeAeson UtxoR where
-  decodeAeson j = UtxoR <$> parseUtxoQueryResult j
+  decodeAeson = map UtxoR <<< parseUtxoQueryResult
 
 -- the inner type for Utxo Queries
 type UtxoQueryResult = Map.Map OgmiosTxOutRef OgmiosTxOut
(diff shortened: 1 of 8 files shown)


src/QueryM/Ogmios.purs
```diff
@@ -209,7 +303,8 @@ instance Show ChainOrigin where
 
 -- | A point on the chain, identified by a slot and a block header hash
 type ChainPoint =
-  { slot :: Slot
+  { slot :: BigInt -- I think we need to use `BigInt` here, 18446744073709552000
```
adamczykm, line 306: Maybe we should change Slot to wrap BigInt instead. Both node uses Integer, ogmios U64 and CSL goes with U32.
I guess similar thing wrt `Epoch`. At least we could reclaim our newtype wrappers and make the query result 
more safe

adamczykm: Alternatively, we could set up another layer of newtype wrappers along with descriptions/docstrings for "securing" ogmios schema.
Idk.
I feel like we're missing an overarching unified approach to 'transporting' types between services.

adamczykm (commented):
Left some comments, otherwise lgtm

vvtran:
Tempted to close in favour of https://github.com/Plutonomicon/cardano-transaction-lib/pull/447

ngua:
@vvtran Agreed

PR #441, opened 2022-05-24
AlwaysSucceeds spending example

Luis-omega:


examples/AlwaysSucceeds.purs
```diff
@@ -37,26 +46,60 @@ main = launchAff_ $ do
...
+    liftedM "Failed to balance/sign tx" $ balanceAndSignTx ubTx
+  txId <- submit bsTx.signedTxCbor
+  logInfo' $ "Tx ID: " <> show txId
+
+spendFromAlwaysSucceeds :: ValidatorHash -> Validator -> Contract () Unit
+spendFromAlwaysSucceeds vhash validator = do
+  let
+    arbitraryRedeemer = unitRedeemer
+    scriptAddress = scriptHashAddress vhash
+  (UtxoM utxos) <-
```
ngua, line 81: Redundant parens here

examples/AlwaysSucceeds.purs
```diff
@@ -37,26 +46,60 @@ main = launchAff_ $ do
...
+    lookups = Lookups.validator validator
+      <> Lookups.unspentOutputs utxos
+
+    orefs :: List Transaction.TransactionInput
+    orefs = fst <$> Map.toUnfoldable utxos
+
+    constraints :: TxConstraints Unit Unit
+    constraints =
+      fold
+        ( do
```
ngua, line 94: These are also redundant. Purescript enables the equivalent of `-XBlockArguments` by default, so you can write

```
constraints = fold do
  ...
```
or `fold $ do ...` is also OK

examples/AlwaysSucceeds.purs
```diff
@@ -37,26 +46,60 @@ main = launchAff_ $ do
...
+    arbitraryRedeemer = unitRedeemer
+    scriptAddress = scriptHashAddress vhash
+  (UtxoM utxos) <-
+    fromMaybe (UtxoM Map.empty) <$> utxosAt scriptAddress
+  let
+    lookups :: Lookups.ScriptLookups PlutusData
+    lookups = Lookups.validator validator
+      <> Lookups.unspentOutputs utxos
+
+    orefs :: List Transaction.TransactionInput
```
ngua, line 88: You can use an `Array` instead of a `List`, which is a primitive, builtin type in Purescript

examples/AlwaysSucceeds.purs
```diff
@@ -37,26 +46,60 @@ main = launchAff_ $ do
...
+    orefs = fst <$> Map.toUnfoldable utxos
+
+    constraints :: TxConstraints Unit Unit
+    constraints =
+      fold
+        ( do
+            oref <- orefs
```
ngua, line 95: You should try to spend just one unspent output at the script address I think. This will try to spend _all_ of them, no? There's likely a _lot_ locked at this address. This is the standard Plutus always succeeds script, it's not CTL-specific at all

examples/AlwaysSucceeds.purs
ngua: But this is kind of a moot point since you should adjust how your building the constraints anyway. See my comment below. You should really just have one `mustSpendScriptOutput`

examples/AlwaysSucceeds.purs
```diff
@@ -37,26 +46,60 @@ main = launchAff_ $ do
     logInfo' "Running Examples.AlwaysSucceeds"
     validator <- liftContractM "Invalid script JSON" $ alwaysSucceedsScript
     vhash <- liftedM "Couldn't hash validator" $ validatorHash validator
+    logInfo' $ "Attempt to lock testada"
+    payToAlwaysSucceeds vhash validator
+    logInfo' $ "Try to spend locked values"
```
vvtran, line 51: redundant `$`

examples/AlwaysSucceeds.purs
Luis-omega: Yes, I was just translating the Plutus pioneer example for spending script outputs. I haven't looked that currently are around 14k founds locked by the script.

Luis-omega:
Right now I'm unable to debug it but I guess this still lacks a constraint about the payment to the wallet owner unless the balancing solves it by itself.  Anyway, I would delay it until I can debug/test the example.

Luis-omega:
It seems that the problem with the test address is solved now, so I can verify it now.

examples/AlwaysSucceeds.purs
```diff
@@ -25,37 +27,112 @@ import Contract.Transaction
...
+    liftAff <<< delay $ Milliseconds 1000.0
+    countToZero (n - 1)
+
+payToAlwaysSucceeds
+  :: ValidatorHash
+  -> Validator
+  -> Contract () TransactionHash
+payToAlwaysSucceeds vhash validator = do
+  let
+    -- Note that CTL does not have explicit equivalents of Plutus'
```
klntsky, line 72: Could you please move this comment to `Types.TxConstraints`

Luis-omega: done.

Luis-omega:
Right now I can't test it as Ogmios return an empty list of utxos in my wallet. 

```
ogmios_1              | {"severity":"Info","timestamp":"2022-05-30T19:57:24.141529599Z","thread":"479","message":{"WebSocket":{"contents":{"era":"Alonzo","tag":"StateQueryRequest","query":{"utxo":["addr_test1qzf7urzwfexnpqakjl0p9ejupr07jcxt4r6r8txfgkpg85wyuwa620klsgupl7l0lthlx22uv08thdztnsw8s25p33nsapwugn"]},"point":null},"tag":"WebSocketStateQuery"}},"version":"v5.2.0"}
ogmios_1              | {"severity":"Info","timestamp":"2022-05-30T19:57:24.660287229Z","thread":"479","message":{"WebSocket":{"contents":{"tag":"StateQueryResponse","result":[]},"tag":"WebSocketStateQuery"}},"version":"v5.2.0"}
```

But the example has been working before.

examples/AlwaysSucceeds.purs
```diff
@@ -25,37 +27,107 @@ import Contract.Transaction
...
+    logInfo' "Try to spend locked values"
+    spendFromAlwaysSucceeds vhash validator txId
+
+countToZero :: Int -> Contract () Unit
+countToZero n =
+  if n <= 0 then
+    pure unit
+  else do
+    logInfo' $ "Waiting before we try to unlock : " <> show n
+    liftAff <<< delay $ Milliseconds 1000.0
```
ngua, line 63: `Milliseconds` has a `Newtype` instance, so you can use `wrap` instead of the constructor. Also saves you an import

examples/AlwaysSucceeds.purs
```diff
@@ -25,37 +27,107 @@ import Contract.Transaction
...
-    logInfo' $ "Tx ID: " <> show txId
+    logInfo' "Attempt to lock value "
+    txId <- payToAlwaysSucceeds vhash validator
+    countToZero 20
+    logInfo' "Try to spend locked values"
+    spendFromAlwaysSucceeds vhash validator txId
+
+countToZero :: Int -> Contract () Unit
+countToZero n =
+  if n <= 0 then
```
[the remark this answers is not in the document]
ngua: Oh wait, this should use `unless`, there's no need for `if` (`unless (n <= 0) $ do ...`)

ngua (commented):
Nice, I think this really improves the example and makes it more illustrative of the `Contract` interface

Just a few suggestions below

examples/AlwaysSucceeds.purs
```diff
@@ -25,37 +27,107 @@ import Contract.Transaction
...
+      in
+        void
+          $ buildSingnAndSubmitTx lookups constraints
+    _ ->
+      logInfo' $ "The id "
+        <> show txId
+        <> " does not have output locked at : "
+        <> show scriptAddress
+  where
+  hasTransactionId :: forall t. TransactionInput -> t -> Boolean
```
ngua, line 116: Missing a kind signature here

examples/AlwaysSucceeds.purs
```diff
@@ -25,37 +27,107 @@ import Contract.Transaction
...
-
-      lookups :: Lookups.ScriptLookups PlutusData
-      lookups = Lookups.validator validator
-
-    ubTx <- liftedE $ Lookups.mkUnbalancedTx lookups constraints
-    BalancedSignedTransaction bsTx <-
-      liftedM "Failed to balance/sign tx" $ balanceAndSignTx ubTx
-    txId <- submit bsTx.signedTxCbor
-    logInfo' $ "Tx ID: " <> show txId
+    logInfo' "Attempt to lock value "
```
vvtran, line 51: redundant space at the end of the string.

examples/AlwaysSucceeds.purs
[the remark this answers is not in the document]
vvtran: Yeah I agree with the `unless`, you can even drop the `$` like `BlockArguments`
```
unless (n <= 0) do
    logInfo' $ "Waiting before we try to unlock : " <> show n
    liftAff <<< delay $ Milliseconds 1000.0
    countToZero (n - 1)
```

examples/AlwaysSucceeds.purs
```diff
@@ -25,37 +27,107 @@ import Contract.Transaction
...
+  -> Contract () Unit
+spendFromAlwaysSucceeds vhash validator txId = do
+  let
+    scriptAddress = scriptHashAddress vhash
+  UtxoM utxos <-
+    fromMaybe (UtxoM Map.empty) <$> utxosAt scriptAddress
+
+  let
+    filteredById =
+      Map.filterWithKey hasTransactionId utxos
```
vvtran, line 95: Just aesthetically, we may as well keep these on oneline, since they're all below 80 characters anyway
```
let scriptAddress = scriptHashAddress vhash
UtxoM utxos <- fromMaybe (UtxoM Map.empty) <$> utxosAt scriptAddress
let filteredById = Map.filterWithKey hasTransactionId utxos
```

examples/AlwaysSucceeds.purs
```diff
@@ -25,37 +27,107 @@ import Contract.Transaction
...
+        lookups :: Lookups.ScriptLookups PlutusData
+        lookups = Lookups.validator validator
+          <> Lookups.unspentOutputs utxos
+
+        constraints :: TxConstraints Unit Unit
+        constraints =
+          Constraints.mustSpendScriptOutput txInput unitRedeemer
+      in
+        void
+          $ buildSingnAndSubmitTx lookups constraints
```
vvtran, line 109: Again, I would keep this to one line `void $ buildSingnAndSubmitTx lookups constraints`

examples/AlwaysSucceeds.purs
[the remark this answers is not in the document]
vvtran: Wait, this is outdated.

examples/AlwaysSucceeds.purs
Luis-omega: I keep forgetting we have the `Newtype` class in PS, is just nice to have it!

examples/AlwaysSucceeds.purs
Luis-omega: It was the result of refactoring the original example that takes all the locked UTxOs.   I still forget that PS is a strict language, this is a nice way to simplify the code.

Luis-omega:
I can test it again (thanks for the help with that). All seems to be ok, so unless we have more refactors to do, this pr is ready.

ngua (approved):

PR #443, opened 2022-05-24
Viet/further ogmios queries

vvtran:
- Based on https://github.com/Plutonomicon/cardano-transaction-lib/pull/440
- Add `systemStart` and `currentEpoch` queries to aid with slot arithmetic.
- Added more type safety compared to the PR this is based on instead of just using `BigInt`s everywhere.

PR #447, opened 2022-05-24
Viet/410 construct exact posix time range

vvtran:
Based on https://github.com/Plutonomicon/cardano-transaction-lib/pull/443 which is based on https://github.com/Plutonomicon/cardano-transaction-lib/pull/440 - I am tempted to remove these old PRs although it means reviewing a larger one.
- Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/410
Includes
- Added Ogmios Queries for EraSummaries, CurrentEpoch (turns out this wasn't needed but useful as `Contract` code anyway), `SystemStart` - these have been lifted to `Contract.Time`.
~- Got rid of replicate json wsp request and response types in QueryM.DatumCacheWsp.~
- A lot of  integer types in Ogmios queries return integers up to `18446744073709552000` which is well out of `UInt` precision, in particular they return absolute slots of these values which don't fit in inside our `Serialization.Address.Slot`. Similarly for `Cardano.Types.Epoch`. I've created extra datatypes for now (e.g. `AbsSlot`) but it may be good to change `Serialization.Address.Slot` etc. into  `BigInts`s. The Alonzo.cddl says they are uInts with some unknown precision. I will likely leave this for a separate PR if so. Any thoughts?
- Removed `Types.Interval` and refactored into `Time` namespace (there are Plutus types so may not be ideal). Edit: I've left this for now per @klntsky saying it has already been refactored and shouldn't be undone. My purpose was always to improve the CTL framework and refactor things incrementally (to reduce the size of a refactor later on). I didn't realise the knock on effect it had to Purs bridge.
- There is a new datatype called `newtype OnchainPOSIXTimeRange = OnchainPOSIXTimeRange POSIXTimeRange`  defined for the round trip on "off chain CTL" `POSIXTimeRange`:
1) `POSIXTimeRange` -> `SlotRange`
2) `SlotRange` -> `TransactionValidity` - I kept this type alias hidden.
3) `TransactionValidity` -> `OnchainPOSIXTimeRange`
- Most of the conversion functions that close this PR are defined in `Time.Conversion`; e.g. `slotToPosixTime`, `posixTimeToSlot`, `posixTimeRangeToSlotRange`, `slotRangeToPosixTimeRange`, `slotRangeToTransactionValidity`, `posixTimeRangeToTransactionValidity`, `toOnchainPosixTimeRange` with _some_ lifted to `Contract.Time`. Note that `toOnchainPosixTimeRange` is intended to convert the off-chain contract PTR to its on-chain counterpart (found in `ScriptContext`) if all works out.
- There are a few helpers like `absSlotFromSlot` and `slotFromAbsSlot` for effectively converting between `UInt` and `BigInt`.
- The slot <> POSIX machinery require `EraSummariesQR` and `SystemStartQR` given by the Ogmios queries `Contract.Time.getEraSummaries` and `Contract.Time.getSystemStart`. I felt it was best to have these as parameters (although less ergonomical) to encourage one source of truth in a contract. I.e. we query the above parameters once and use as inputs for the rest of the contract, instead of each helper `slotRangeToTransactionValidity` and `posixTimeRangeToTransactionValidity` querying each time. Of course, I would _hope_ the output of each query is static in a Contract anyway. There is one exception to this which is when it is queried inside `Types.ScriptLookups` - we could improve this.
- I've removed `SlotConfig` from the contract config altogether, many of the helpers implemented in `Types.Interval` were broken, this included.
~- **Please note:** it is currently untested. I need to figure out how to test this.~ This now has some basic tests.


src/Time/Conversion.purs
```diff
@@ -0,0 +1,255 @@
...
+derive instance Generic AbsTime _
+derive instance Newtype AbsTime _
+derive newtype instance Eq AbsTime
+derive newtype instance Ord AbsTime
+
+instance Show AbsTime where
+  show = genericShow
+
+-- | Find the relative slot provided we know the `AbsSlot` for an absolute slot
+-- | given an `EraSummary`. We could relax the `Maybe` monad if we use this
```
klntsky, line 153:
```suggestion
-- | given an `EraSummary`. We could relax the `Either` monad if we use this
```

klntsky, line 153:
```suggestion
-- | given an `EraSummary`. We could relax the `Either` monad if we use this
```

bladyjoker (commented):
Hi Viet, thanks for doing this. Can we just be conservative with shuffling modules around? Concretely, Types.Interval is a plutus-ledger-api type, and change to this breaks the whole chain (ctl -> purs-bridge -> haskell contracts).

src/QueryM/DatumCacheWsp.purs
```diff
@@ -57,25 +54,6 @@ newtype WspFault = WspFault Aeson
...
-  , mirror :: ListenerId
-  }
-
-type JsonWspResponse =
-  { type :: String
-  , version :: String
-  , servicename :: String
-  , methodname :: String
-  , result :: Maybe Aeson
-  , fault :: Maybe WspFault
```
klntsky, line 75: This is not identical to `QueryM.JsonWsp.JsonWspResponse`

klntsky (changes requested):

MangoIV:
thank you for the efforts, that's really neat!

vvtran:
I may have overcomplicated things by have some many things in QueryM/Contract. This was because I originally queried inside the function. I can attempt to simplify this tomorrow.

vvtran:
I added some basic inversion tests which helped me debug a few things. Furthermore, I'm using Milliseconds precision. So if you provide your `POSIXTime` with milliseconds precision (of course POSIXTime is defined to be with Millisecond precision), using `toOnchainPosixTimeRange` will do some rounding so that your "on chain" `POSIXTime` will be in mllliseconds but rounded to the nearest second, I believe this is expected behaviour.

All of the conversion functions are re-export from `Contract.TIme` as `Effect` instead of `QueryM` now.

src/Time/Conversion.purs
```diff
@@ -0,0 +1,475 @@
...
+  convertBounds
+    :: Extended Slot
+    -> Effect (Either SlotToPosixTimeError (Extended POSIXTime))
+  convertBounds (Finite pt) = slotToPosixTime eraSummaries sysStart pt
+    <#> map Finite
+  convertBounds NegInf = pure $ Right NegInf
+  convertBounds PosInf = pure $ Right PosInf
+
+type TransactionValiditySlot =
+  { validityStartInterval :: Maybe Slot, timeToLive :: Maybe Slot }
```
Saizan, line 398: What do you mean by `timeToLive` here? in the function `slotRangeToTransactionValidity` that field is filled with the upper bound, but the field name almost seems to suggest a length?

How is `TransactionValiditySlot` used to build the transaction sent to the node?

Is this supposed to correspond to `ValidityInterval` from `cardano-ledger` ?
Note that describes an half-open interval: https://github.com/input-output-hk/cardano-ledger/blob/94b1ae3d6b66f4232f34060d89c2f12628998ef2/eras/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Timelocks.hs#L100-L106

test/AffInterface.purs
```diff
@@ -97,14 +111,105 @@ testOgmiosDatumCacheFetcher =
...
+      Left err -> throw $ show err
+      Right slot -> do
+        ePosixTime <- slotToPosixTime es ss slot
+        either (throw <<< show) (shouldEqual $ transf posixTime) ePosixTime
+
+  mkPosixTime :: String -> POSIXTime
+  mkPosixTime = POSIXTime <<< unsafePartial fromJust <<< BigInt.fromString
+
+testSlotToPosixTime :: Aff Unit
+testSlotToPosixTime = do
```
bladyjoker, line 188: How do you feel about introducing Quickcheck for tests? Probably out of scope for this PR, but generally speaking it would be great to start building out a good support for quickcheck in CTL.

bladyjoker (approved):
Hi @vvtran thanks for doing this! I left a couple of comments of a 'nice to have' nature...

- Symmetric and tested Aeson instances (adding EncodeAeson and a roundtrip test),
- Provide tests that trigger all of the declared errors in PosixTimeToSlotError (if feasible),
- Quickcheck tests...

test/AffInterface.purs
klntsky: I made an issue for that
https://github.com/Plutonomicon/cardano-transaction-lib/issues/466

src/Types/Interval.purs
```diff
@@ -380,161 +430,434 @@ after h (Interval { to: to' }) = upperBound h > to'
...
+derive instance Generic PosixTimeToSlotError _
+derive instance Eq PosixTimeToSlotError
+
+instance Show PosixTimeToSlotError where
+  show = genericShow
+
+-- | Converts a `POSIXTime` to `Slot` given an `EraSummariesQR` and
+-- | `SystemStartQR` queried from Ogmios.
+posixTimeToSlot
+  :: EraSummariesQR
```
klntsky, line 636: Maybe the `QR` suffix should be dropped from these type names since we use them in domain functions?

klntsky (approved):

test/AffInterface.purs
bladyjoker: Great! Resolving this and tackling in #466

vvtran:
As I was writing extra Aeson instances/tests, I noticed one the error constructors is redundant so removed.

ngua (approved):
I think we should merge this since the remaining issues will be addressed in #467

Issue #212, opened 2022-03-22
Implement client-side blake2b (256) hashing


ngua:
We have several server endpoints now for hashing, which either directly or indirectly make use of blake2b hashing. We should try to implement this on the client side in order to
- [ ] general hashing functionality (this came up for Seabug's token name hashing, for example)
- [ ] hash scripts
- [ ] hash datums

PR #448, opened 2022-05-24
Implement client-side hashing

errfrom:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/212


src/Hashing.js
```diff
@@ -0,0 +1,29 @@
+/* global require exports BROWSER_RUNTIME */
+
+const Blake2 = require('blakejs');
```
klntsky, line 3: I wonder if we could use `blake2.wasm` and `node-blake2` packages for speed, since we have conditional compilation anyway.

This is not a change request, performance is probably acceptable here.

package-lock.json
```diff
@@ -1,8 +1,5130 @@
 {
```
klntsky, line 1: Looks like you accidentally changed permissions to this file to 755 instead of 644.

klntsky (changes requested):
Looks great to me except of one seemingly unwanted change

vvtran (commented):
I haven't read it in detail but a few comments.

src/Contract/PlutusData.purs
```diff
@@ -128,4 +128,4 @@ cancelFetchBlocks = wrapContract QueryM.cancelFetchBlocks
 
 -- | Hashes a Plutus-style Datum
 datumHash :: forall (r :: Row Type). Datum.Datum -> Contract r (Maybe DataHash)
-datumHash = wrapContract <<< QueryM.datumHash
+datumHash = pure <<< Hashing.hashDatum
```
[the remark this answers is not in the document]
vvtran: Perhaps we can redefine as `datumHash = Hashing.hashDatum` or changed the original name  `Hashing.hashDatum` -> `Hashing.datumHash`

klntsky (approved):

PR #457, opened 2022-05-26
Hashing: Add sha256 and sha3_256 hash functions

errfrom:


klntsky (approved):
LGTM

ngua (approved):
My only concern is that the lockfile was generated with a different version of npm than the one provided in the dev shell

ngua:
OK, ignore my previous comment, just tried `npm i --package-lock-only` and it produced the same output (is it because the lockfile specifies the different version now?)

Issue #460, opened 2022-05-27
Ogmios protocol parameters query response changed in a weird way


Luis-omega:
Now it returns a more compact response : 

```
{
  "type": "jsonwsp/response",
  "version": "1.0",
  "servicename": "ogmios",
  "methodname": "Query",
  "result": {
    "minFeeCoefficient": 44,
    "minFeeConstant": 155381,
    "maxBlockBodySize": 65536,
    "maxBlockHeaderSize": 1100,
    "maxTxSize": 16384,
    "stakeKeyDeposit": 2000000,
    "poolDeposit": 500000000,
    "poolRetirementEpochBound": 18,
    "desiredNumberOfPools": 500,
    "poolInfluence": "3/10",
    "monetaryExpansion": "3/1000",
    "treasuryExpansion": "1/5",
    "decentralizationParameter": "1/4",
    "extraEntropy": "neutral",
    "protocolVersion": {
      "major": 3,
      "minor": 0
    },
    "minUtxoValue": 1000000,
    "minPoolCost": 340000000
  },
  "reflection": null
}

```

This makes the current Aeson parser fail and the Haskell server fails.

Luis-omega:
And without changing anything we have again 

```
{
  "severity": "Info",
  "timestamp": "2022-05-27T03:39:55.875969662Z",
  "thread": "418",
  "message": {
    "WebSocket": {
      "contents": {
        "tag": "StateQueryResponse",
        "result": {
          "poolDeposit": 500000000,
          "extraEntropy": "neutral",
          "maxValueSize": 5000,
          "minFeeCoefficient": 44,
          "minFeeConstant": 155381,
          "minPoolCost": 340000000,
          "monetaryExpansion": "3/1000",
          "desiredNumberOfPools": 500,
          "maxExecutionUnitsPerTransaction": {
            "memory": 10000000,
            "steps": 10000000000
          },
          "maxExecutionUnitsPerBlock": {
            "memory": 50000000,
            "steps": 40000000000
          },
          "maxBlockHeaderSize": 1100,
          "maxCollateralInputs": 3,
          "protocolVersion": {
            "minor": 0,
            "major": 5
          },
          "treasuryExpansion": "1/5",
          "poolInfluence": "3/10",
          "decentralizationParameter": "0/1",
          "coinsPerUtxoWord": 34482,
          "collateralPercentage": 150,
          "prices": {
            "memory": "577/10000",
            "steps": "721/10000000"
          },
[...]

PR #461, opened 2022-05-27
genralized Aeson parser for ogmios protocol parameters in Haskell server

Luis-omega:
Related to #460


Luis-omega:
closed by the same reason as #460

PR #467, opened 2022-05-30
Viet/aeson for posix time range construction

vvtran:
- Aeson types for https://github.com/Plutonomicon/cardano-transaction-lib/pull/447
- Roundtrip testing for Aeson.
- Added extra testing for POSIXTimeRange construction, some of the errors were difficult to reproduce so I only did a subset.


bladyjoker (approved):
Awesome thanks!

ngua (approved):

PR #469, opened 2022-05-31
plutus ctl comparative example

Luis-omega:
This is still a draft, the organization of the content can change at any time but most of the info would be the same unless we discuss otherwise.


examples/PlutusComparison.md
```diff
@@ -0,0 +1,288 @@
...
+The definition of `LoggerT` is : 
+
+```PureScript
+newtype LoggerT m a = LoggerT ((Message -> m Unit) -> m a)
+```
+
+As CTL has been written in PureScript and it's intended to be used inside a 
+browser, it makes little sense to allow more general logging capabilities. 
+This is the reason why `QueryMExtended` uses directly `LoggerT` in its 
+definition instead of allowing an arbitrary user writer.
```
ngua, line 81: I think this might be a bit misleading: we added `LoggerT` to the stack to allow for structured logging, but we actually allow arbitrary effects inside our `Contract`, unlike Plutus. In fact, Plutus adds `Writer` capabilities because it doesn't support arbitrary effects like logging. In CTL, you don't need to use the structured logging -- you could also just log to the console directly

examples/PlutusComparison.md
```diff
@@ -0,0 +1,288 @@
...
+    ]
+
+type ContractEnv = (IterationID, RequestID)
+
+newtype Contract w (s :: Row *) e a = Contract { unContract :: Eff (ContractEffs w e) a }
+  deriving newtype (Functor, Applicative, Monad)
+```
+
+The Plutus `Contract` environment is specialized to just two values and is fixed. 
+Also, Plutus `Contract` uses a phantom type to allow users
```
ngua, line 68: The phantom type parameter `s` is for the contract schema, which CTL doesn't have

examples/PlutusComparison.md
```diff
@@ -0,0 +1,288 @@
...
+both of them.
+
+
+The purpose of `Contract` in CTL is to bring the capabilities of 
+`QueryMExtended` to the public API. 
+
+The current definition of `Contract` in CTL is :
+
+```PureScript
+type QueryConfig (r :: Row Type) =
```
ngua, line 15: I'm not sure if we need to reproduce the entire `QueryConfig` here. It will probably lead to the docs drifting away from the actual definitions. In fact, `slotConfig` has already been removed (recently)

examples/PlutusComparison.md
```diff
@@ -0,0 +1,288 @@
+
```
ngua, line 1: Please move this file to a `doc` directory in the repo root. Also, I would change the name of file to something like `side-by-side-comparison.md`, since you're directly comparing two programs

ngua, line 1: Could you add a document title as well please?

ngua (commented):
I think the side-by-side comparison of the two contracts is very helpful. Please add more about how CTL requires a more explicit build-balance-sign-submit flow than Plutus does (that's a big difference I think)

I also have some suggestions for your initial discussion of the differences between the two `Contract` monads. I'm also working on a more prose-oriented document that discusses this at a more abstract level, so feel free to keep that section brief

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+and parameters `w` for a writer and `e` for errors. 
+In the case of CTL we don't have the contract schema parameter or the writer 
+parameter since CTL definition allows performing arbitrary effects. 
+This is possible  since the definition of `LoggerT` is: 
+
+```PureScript
+newtype LoggerT m a = LoggerT ((Message -> m Unit) -> m a)
+```
+
+The use of `Aff` inside `LoggerT` allows us to use asynchronous effects
```
vvtran, line 73: We could easily remove `LoggerT` and the stack would still hold (up to logging changes). You can use `Effect.Class.Console (log)` to log to console.

There is no need for writer as we can perform arbitrary async. effects since we have `Aff` at the bottom of the stack?

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+    a
+
+newtype Contract (r :: Row Type) (a :: Type) = Contract (QueryMExtended r a)
+```
+
+
+In CTL we have a general environment type `forall r . QueryConfig r`, it stores
+the needed parameters to connect to an `Ogmios` server  and
+uses the PureScript native row polymorphism to make it extensible for both 
+CTL developers and users.
```
vvtran, line 40: Let's be specific that it is the `r :: Row Type` that allows this extensibility as not everyone will be familiar with this.

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+`mustPayToOtherScript`, according to Plutus ledger, it :
+
+```Haskell
+-- | mustPayToOtherScript vh d v locks the value v with the given script 
+-- | hash vh alonside a datum d.
+-- | If used in OffChain, this constraint creates a script output with 
+-- | vh, d and v, and adds d in the 
+-- | transaction's datum witness set.
+mustPayToOtherScript 
+  :: forall i o. ValidatorHash -> Datum -> Value -> TxConstraints i oSource#
```
vvtran, line 145: remove `Source#`

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+
+
+### MustPayTo functions
+
+
+In the case of Plutus `Contract`, we use the function 
+`mustPayToOtherScript`, according to Plutus ledger, it :
+
+```Haskell
+-- | mustPayToOtherScript vh d v locks the value v with the given script
```
klntsky, line 139:
```suggestion
-- | `mustPayToOtherScript vh d v` locks the value v with the given script 
```

Please also backtick-quote all variable names.

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+type Utxo = Map TransactionInput TransactionOutput
+newtype UtxoM = UtxoM Utxo
+```
+
+```PureScript
+module Contract.Utxos ...
+.
+.
+.
+-- | This module defines query functionality via Ogmios to get utxos. Gets
```
klntsky, line 255: Please use `UTxO` instead of `utxo` spelling in docs and comments.

I will do the same change to existing docs soon.

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+newtype UtxoM = UtxoM Utxo
+```
+
+```PureScript
+module Contract.Utxos ...
+.
+.
+.
+-- | This module defines query functionality via Ogmios to get utxos. Gets 
+-- | utxos at an (internal) Address in terms of a Plutus Address.
```
klntsky, line 256: It look like you copy-pasted the docs from generated documentation instead of the code, losing backticks for identifiers.

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+    tx.transactionId == txId
+```
+
+Notice the explicit signature in:
+
+```
+  fst <$> find hasTransactionId (Map.toUnfoldable utxos :: Array (_ /\ _))
+```
+
+Since PureScript has JS as the backend, `Arrays` are a data type available as
```
klntsky, line 312:
```suggestion
Since PureScript has JS as the backend, `Array` is the most used container (instead of lists as in Haskell)
```

klntsky:
I changed the base, because this branch has pulled commits from master that are not in develop.

Since this PR does not contain breaking changes, we are OK with merging it to `master`.

klntsky (changes requested):

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+```Haskell
+-- Haskell
+grab :: forall w s e. AsContractError e => Contract w s e ()
+grab = do
+   utxos <- utxoAt scrAddress
+   let orefs   = fst <$> Map.toList utxos
+      lookups  = Constraints.unspentOutputs utxos      <>
+                 Constraints.otherScript validator
+      tx :: TxConstraints Void Void
+      tx       = mconcat [mustSpendScriptOutput oref $ Redeemer $ I 17 | oref <- orefs]
```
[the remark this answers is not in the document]
Luis-omega: this is a copy of the source code, so this should be a thing with the autoformat. I will remove them here aniway.

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+grab
+  :: ValidatorHash
+  -> Validator
+  -> TransactionHash
+  -> Contract () Unit
+grab vhash validator txId = do
+  let scriptAddress = scriptHashAddress vhash
+  UtxoM utxos <- fromMaybe (UtxoM Map.empty) <$> utxosAt scriptAddress
+  case
+    fst <$> find hasTransactionId (Map.toUnfoldable utxos :: Array (_ /\ _))
```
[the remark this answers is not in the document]
Luis-omega: Right. I still believe that being specific by using`(Array (_ /\ _))` helps to understand the example.

doc/side-by-side-CTL-and-Plutus-comparison.md
Luis-omega: I have my doubts here since I now copying the comments from the source and it uses the haddock style:

`@mustPayToOtherScript vh d v@`

doc/side-by-side-CTL-and-Plutus-comparison.md
Luis-omega: I'm not sure how to add this to the document without going into a discussion about `Haskell` vs `PS` type systems instead of `Plutus` vs `CTL`, so I just add a Wikipedia link and a link to [other source ](https://hgiasac.github.io/posts/2018-11-18-Record-Row-Type-and-Row-Polymorphism.html)

doc/side-by-side-CTL-and-Plutus-comparison.md
```diff
@@ -0,0 +1,323 @@
...
+  txId <- submit bsTx.signedTxCbor
+  logInfo' $ "Tx ID: " <> show txId
+  pure txId
+```
+
+
+This function takes our lookups and constraints, then constructs
+an unbalanced transaction using them, after that it tries to balance 
+the transaction and attempts to get a signature for this transaction, then it 
+submits the transaction and if all got right, it returns the Id
```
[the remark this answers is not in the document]
Luis-omega: In the end, I write a list of actions being performed and then write about the kind of errors that could happen here.

doc/side-by-side-CTL-and-Plutus-comparison.md
klntsky: Purescript docs use markdown.

Luis-omega: this part is Haskell code.

ngua (approved):
Just going to rename the markdown file to be in line with our other docs

PR #471, opened 2022-05-31
Add docs

### LINKED ISSUE #192: Add migration guide with Plutus `Contract` comparison
Once we have our own `Contract` interface set, we should add something to the docs to illustrate the similarities and differences with Plutus' `Contract` API

ngua:
Closes #192 (mostly, we will also add @Luis-omega's side-by-side comparison). In addition to the Plutus comparison, I split the readme into smaller documents


doc/plutus-comparison.md
```diff
@@ -0,0 +1,179 @@
...
+**CTL**:
+
+```purescript
+newtype Contract (r :: Row Type) (a :: Type) = Contract (QueryMExtended r a)
+```
+
+where
+
+- `QueryMExtended` is an internal monad transformer stack based on `ReaderT` over `Aff` (Purescript's monad for asynchronous effects, with no Haskell analogue).
+- `r` extends the record serving as the reader environment
```
klntsky, line 40: Not related to PR: do we really need this `r`? Why cant people throw another `ReaderT` on top for their configs?

doc/plutus-comparison.md
```diff
@@ -0,0 +1,179 @@
...
+
+**Plutus**:
+
+```haskell
+newtype Contract (w :: Type) (s :: Row Type) (e :: Type) (a :: Type) = Contract { ... }
```
klntsky, line 45: `Row *`

doc/plutus-comparison.md
ngua: That would also be possible, but I'm quite fond of the extensible record approach

doc/plutus-comparison.md
ngua: I didn't reproduce the `Contract` definition verbatim. `*` is a soon-to-be-deprecated synonym for `Type` anyway

klntsky (approved):

Issue #463, opened 2022-05-27
Ensure redeemers are indexed correctly before setting ex units 


errfrom:
In some cases the redeemer indexes don't match in the arguments of this call:
https://github.com/Plutonomicon/cardano-transaction-lib/blob/7a7d8f442c4dc3b6e5d48469de806ef7cb50e6a4/src/BalanceTx.purs#L284-L285

1. We probably need to reindex redeemers before setting ex units
2. Provide examples / tests that would cover the case of having multiple redeemers

MangoIV:
I've looked into this issue because I hoped to help solve it; I didn't quite understand what you meant with your first bullet point, I think the redeemers already get reindexed before setting the ex units, however, looking at the reindex function [here](https://github.com/Plutonomicon/cardano-transaction-lib/blob/7a7d8f442c4dc3b6e5d48469de806ef7cb50e6a4/src/ReindexRedeemers.purs#L49) I found that currently, if the `txOutRef` is Nothing, a (Spending) redeemer will just be skipped as you will then get into the `mintRef` case. Is that intentional?

PR #473, opened 2022-05-31
Fix redeemers reindexing and ex units budget calculation (In Progress)

errfrom:
Context: https://github.com/Plutonomicon/cardano-transaction-lib/issues/463

### DIFF
--- a/src/BalanceTx.purs
+++ b/src/BalanceTx.purs
@@ -97,7 +97,11 @@ import QueryM
   , evalTxExecutionUnits
   )
 import QueryM.Utxos (utxosAt)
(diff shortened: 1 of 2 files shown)


ngua:
I want to merge master in so we can run the new `AlwaysSucceeds` example, which requires a `Spend` redeemer

ngua:
Although I suppose that you and @bladyjoker were able to run an example with multiple redeemers already

ngua (approved):
This seems to be correct. Is a minimal example that includes multiple redeemers going to be part of this PR or a separate one?

bladyjoker:
> Although I suppose that you and @bladyjoker were able to run an example with multiple redeemers already

That's correct, we were able to successfully run a contract yesterday.

PR #477, opened 2022-05-31
add space for some `Show` instances

vvtran:
- A few `Show` instances were missing a space, `NonAdaAsset` is already taken care of in https://github.com/Plutonomicon/cardano-transaction-lib/pull/476
- On the end of a comment, I changed all `BigInt` newtype wrappers to have extra parenthesis on `Show` instances.


klntsky:
Thanks! Another annoying thing is that some types miss parentheses in `Show`. E.g. `(Coin fromString "351823")`. But we can sort of address that in wrapper instance: `(Coin (fromString "351823"))`. Could you please look for problematic instances of this kind?

vvtran:
@klntsky wait, this is true for every `BigInt` newtype? Oh dear.

vvtran:
`UInt` is different e.g. `(Epoch 5u)` which is okay I guess

ngua (commented):
Too bad we can't use `genericShow` for these

Can we abstract away this common pattern of adding parens around an inner, `Show`able type? Something like:

```purescript
showWithParens
  :: forall (t :: Type) (a :: Type)
   . Newtype t a
  => Show a
  => String
  -> t
  -> String
showWithParens ctorName x = "(" <> ctorName <> " (" <> show (unwrap x) <> "))"
```
We could probably get the constructor name generically but I'm not sure if we want to go that far. Anyway, maybe this isn't worth the effort, but I think doing it manually might lead to more issues with various `Show` instances over time

ngua (approved):

PR #481, opened 2022-06-01
Add example contract for multiple redeemers

MangoIV:
- this should work reliably now and as soon as we can also mint more than one token in a go we can easily add this. don't forget to comment out the always succeeds contract if you don't wanna run it.


locals.dhall
```diff
@@ -0,0 +1 @@
+/nix/store/igps0na390g1brb8d4k1fm7062214260-locals-dhall
```
ngua, line 1: How was this generated? At any rate, it should go into .gitignore

examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,254 @@
+module Examples.MultipleRedeemers (threeRedeemerContract) where
+
+import Prelude
+
+import Cardano.Types.Value
```
ngua, line 5: Why use `Cardano.Types.Value`? It should be possible to do everything with the public `Contract` interface

examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,254 @@
...
+        liftContractM "could not make tokennames with values" $ traverse
+          (bitraverse (mkTokenName <=< byteArrayFromAscii) pure)
+          tokens
+
+      logInfo' $ "trying to create" <> show tok
+
+      let
+        toCsValue :: Array (Tuple TokenName Int) -> Value.Value
+        toCsValue t =
+          unwrap <<< toPlutusType <<< Value mempty <<< NonAdaAsset
```
ngua, line 101: You should use the Plutus-style `Value` from `Contract.Value` instead. There's no need to use the Cardano ones, unless I'm missing something (and even if this is necessary at the moment, that's a bug that needs to be fixed -- we don't want users to have to think about the Cardano vs. Plutus thing)

examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,254 @@
...
+        ]
+
+    logInfo' $ "Found " <> show orefs <> " at alwaysSucceeeds address"
+
+    ubTx <- liftedE $ Lookups.mkUnbalancedTx lookups constraints
+    BalancedSignedTransaction bsTx <-
+      liftedM "Failed to balance/sign tx" $ balanceAndSignTx ubTx
+    hash2 <- submit bsTx.signedTxCbor
+    logInfo' $ "Hash of second transaction " <> show hash2
+    pure unit
```
ngua, line 183: This is redundant

examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,254 @@
...
+  pure unit
```
ngua, line 184: This is also redundant

examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,254 @@
...
+          $ Map.keys
+          $ unwrap utxos
+
+      if (length orefs == numTokens) then
+        pure $ Tuple (unwrap utxos) orefs
+      else do
+        logInfo' "Could not find utxos, trying again"
+        getAlwaysSucceedsUtxos
+
+  runContract cfg $ do
```
ngua, line 163:
```suggestion
  runContract_ cfg $ do
```

ngua: But actually this should be moved anyway (see my comment about a single `runContract_` call above)

examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,254 @@
+module Examples.MultipleRedeemers (threeRedeemerContract) where
+
+import Prelude
```
ngua, line 3: I'd recommend adjusting the imports. Having these examples is partially to show how we intend the library to be used. Our goal is that users only need to import `Contract.Prelude` as a replacement for `Prelude` and then don't need to worry about things like `Maybe`, `Either`, `(/\)`, etc...

Also there are several imports of internal modules from `Types.*`, etc... For example, `TokenName` should be imported from `Contract.Value`

examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,254 @@
...
+-- FIXME: this doesn't work without a browser
+threeRedeemerContract :: Aff Unit
+threeRedeemerContract = do
+  log "creating utxos"
+  ContractConfig defaults <- traceContractConfig
+  wallet <- mkNamiWalletAff
+  let
+    cfg :: ContractConfig ()
+    cfg = ContractConfig $ defaults { wallet = pure wallet }
+
```
ngua, line 67: Why not have a single `Contract` action and run it here with `runContract_`? As it is you're calling `runContract` twice -- once to get the tuple and then to build and submit the transaction

examples/AlwaysSucceeds.purs
```diff
@@ -28,9 +28,15 @@ import Contract.TxConstraints as Constraints
 import Contract.Value as Value
 import Contract.Wallet (mkNamiWalletAff)
 import Data.BigInt as BigInt
+import Examples.MultipleRedeemers as MultipleRedeemers
 
 main :: Effect Unit
 main = launchAff_ $ do
+  alwaysSucceedsContract
+  MultipleRedeemers.threeRedeemerContract
```
ngua, line 36: I don't think this should be added to `AlwaysSucceeds`. You can just add a `main` to `MultipleRedeemers` instead. To run it, you can edit the `ps-entrypoint` variable in the Makefile

ngua (changes requested):

examples/MultipleRedeemers.purs
MangoIV: I wanted to differentiate between two contracts. One to create tokens and one to spend those tokens. 
I think that makes things clearer and easier to understand.

examples/MultipleRedeemers.purs
MangoIV: Got it. Currently I'm searching the documentation - pursuit doesn't tell me which modules reexport the things I need hence I'm just taking the module they're defined in.

examples/MultipleRedeemers.purs
ngua: In that case, my recommendation would be to split the top-level function up and then just have:

```purescript
createTokens :: Contract () (TransactionHash /\ Int)

spendTokens ::  Contract () Unit
```

I think that would help clarify what's going on a bit, then both of them can be called within a single `runContract_`

MangoIV: I'll do it like that then :)

examples/MultipleRedeemers.purs
ngua: I see, unfortunately we can't do anything about that AFAIK. For the examples, we should be writing them as a user would. So we should also only use `Contract.*` for CTL stuff

MangoIV: I think it would be sensible to just export a lot of the really commonly used stuff from the Contract.Prelude. Then you wouldn't have to worry about that and which types to import.

examples/AlwaysSucceeds.purs
MangoIV: I'll add that as documentation to the main function.

examples/MultipleRedeemers.purs
MangoIV: So this is still an issue - I can not directly construct a Plutus Value, also I could not purge all the non-Contract imports.

klntsky: > So this is still an issue - I can not directly construct a Plutus Value

There's a Monoid instance and a `singleton` function, which should be enough to construct anything. However, if you can think of other helper functions missing, please share.

.gitignore
```diff
@@ -12,6 +12,7 @@ result
 result-*
 .envrc
 .direnv
+locals.dhall
```
klntsky, line 15: This is not a part of our setup, please remove

MangoIV: copied.

PR #485, opened 2022-06-02
Expose functions for getting wallet balance

nrutledge:
Seabug (and other dApps) need the ability to display the connected wallet's balance. This can be done by calling the `getBalance()` function on the wallet directly and using CSL to decode the cbor that is returned, but this is not ideal. 

This PR exposes functions required for getting the balance of the connected Nami wallet as a CSL `Value`. The actual function being used by Seabug exists in the `seabug-contracts` repo:
```purescript
getWalletBalance :: Effect (Promise (Maybe Value))
getWalletBalance = fromAff $ do
  (Nami nami) <- mkNamiWalletAff
  callNami nami _.getBalance
```

We may want to add something similar in a module within this repo (but it should include all the CIP-30 stuff and perhaps other helpers). 

The frontend can utilize the above `getWalletBalance` function without having to depend directly on CSL. For example, getting the current lovelace balance in the wallet:
```ts
const getWalletLovelace = async () => {
  const balance = await seabugContracts.getWalletBalance();
  try {
    return BigInt(balance.value0.coin().to_str());
  } catch (err) {
    console.log('getWalletBalance() failed', err);
    return 0n;
  }
},
```

### DIFF
--- a/.github/workflows/add-to-project.yaml
+++ b/.github/workflows/add-to-project.yaml
@@ -1,0 +1,16 @@
+name: Add issues to CTL project
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@8d66869b6c4b9b217ec079b4cfbd4cde3fce35f3
+        with:
+          project-url: https://github.com/orgs/Plutonomicon/projects/3
+          github-token: ${{ secrets.ADD_TO_PROJECT }}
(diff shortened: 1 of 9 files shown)


src/Wallet.purs
```diff
@@ -102,6 +106,10 @@ mkNamiWalletAff = do
           Deserialization.UnspentOuput.convertUnspentOutput
             <$> fromBytesEffect (unwrap bytes)
 
+  getBalance :: NamiConnection -> Aff (Maybe Value)
+  getBalance nami = fromNamiHexString _getNamiBalance nami >>=
+    (_ >>= unwrap >>> fromBytes) >>> pure
```
nrutledge, line 111: `fromBytes` here should probably change to `fromBytesEffect` (haven't had time yet to refactor).

klntsky:
-> https://github.com/Plutonomicon/cardano-transaction-lib/pull/625

PR #492, opened 2022-06-03
Automate adding issues to new project

ngua:
This will add all newly opened issues to the new CTL project (can only be done via GH Actions)


klntsky (approved):

ngua:
I need to change some settings to fully enable this, but we can still merge it for now

Issue #466, opened 2022-05-30
QuickCheck support


klntsky:
It may be a good idea to add `Arbitrary` instances to all of our types (in PureScript orphans are disallowed, which makes testing a bit harder because of the need to define newtypes).

Luis-omega:
This seems like something that I could help with and something that would help me either. So unless stated otherwise I will work on it. 

I probably would need some help to spot some corner cases but it could be done as part of the review process.

klntsky:
I doubt our clients are testing their Contract-related code in ways that would let them benefit from arbitrary value generation. Usually, the tests involve running async code and can't be implemented as unit tests or property tests. We should close this until we see anyone actually interested in Arbitrary instances.

PR #501, opened 2022-06-07
Add Arbitrary instances to types.

Luis-omega:
I don't know how to proceed with `BigInt`, it doesn't have an instance of `Arbitrary` but we can use a variation of 

```
instance Arbitrary BigInt where 
  arbitrary = fromSomeIntTypeWithInstance <$> arbitrary
```
Then a lot of the new instances can be auto derived by : 

```
deriving newtype Arbitrary SomeWrapper 
```

I don't know the right place to put such an instance. 

Apart from it, I have doubts in the following types :  

- redeemerHash
- Serialization Address

- CurrencySymbol
- NonAdaAsset (depends on CurrencySymbol)
- Value (depends on NonAdaAsset)
- ValueSpentBalances (depends on Value)
- OutputConstraint (depends on Value)
- TxConstraints (depends on OutputConstraint)

- UnattachedUnbalancedTx
- MkUnbalancedTxError

- PlutusScript
- MintingPolicy (depends on PlutusScript)
- Validator (depends on PlutusScript)
- StakeValidator (depends on PlutusScript)

- ScriptHash
- MintingPolicyHash (depends on ScriptHash)
- ValidatorHash (depends on ScriptHash)
- StakeValidatorHash (depends on ScriptHash)

- BlockHeaderHash
- ChainTip (depends on BlockHeaderHash)
- Tip (depends on ChainTip)

- TransactionMetadatum
- GeneralTransactionMetadata (depends on TransactionMetadatum)

- TxConstraint

- TypedTxOut (depends on Cardano.Types.TransactionOutput)
- TypedTxOutRef (depends on TypedTxOut)

- TypeCheckError (depends on Address and PlutusData)

- TypedValidator (depends on Validator, MintingPolicy)

- Cardano.Types.Transaction.Transaction
- UnbalancedTx (depends on Cardano.Types.Transaction.Transaction)

Most of them are related to `create arbitrary Datum or Redeemer` or `create an arbitrary script for validator or mint` or `create arbitrary transaction`. The ones whose counterparts in `Plutus` I checked, don't have an  `Arbitrary` instance. 

So, I think those must be left without `Arbitrary` instances unless we have a good reason.


klntsky:
Regarding types that don't have instances in Plutus, it's not worth the effort I think (especially for values that should point to some existing on-chain state, like hashes). Let's skip them.

bladyjoker:
Hi @Luis-omega ,
I've found some resources for you that we used on the Haskell side of things to provide sensible Arbitrary instances:
- `plutus-extra` https://github.com/Liqwid-Labs/plutus-extra/blob/master/quickcheck-plutus-instances/src/Test/QuickCheck/Plutus/Instances.hs
- `purescript-bridge` https://github.com/mlabs-haskell/purescript-bridge/blob/master/roundtrip/RoundTrip/ArbitraryLedger.hs

bladyjoker:
Could I suggest that you start with tests first, migrating any unit test cases to use quickcheck, rather than going module by module? For instance I'd start with https://github.com/Plutonomicon/cardano-transaction-lib/blob/master/test/Data.purs but there's plenty of test modules that are obviously a candidate for using quickcheck.

Another reason that I'd suggest starting with tests is that writing Arbitrary instances often lead to unintended recursion and the best way to know whether they actually work is to use them in a test immediately.

Regardless of the route you wish to take, I'd suggest at least to split the effort across smaller PRs, each containing an Arbitrary instance accompanied by a test that uses is.

(P.S. Are you aware of any code coverage tools for Purs/Js? That would be amazing to have)

Luis-omega:
Since this way of doing it looks much more complete, I would like to do it, but I will probably have to open a new ticket or modify the current one #466. Also, if the task is too big maybe it should be spread over multiple tickets. I think this is a good thing worth having in the test suit, but it maybe takes some time, so,  I would like to ask @klntsky  and @ngua  If I should be spending too much time in it first.

Luis-omega:
Sorry, I think I misunderstood the comment before, yeah, porting as many tests as I can to use QuickChec is a good idea, I going to do it asap.

Luis-omega:
I'm closing this branch to follow the tests aproach.

klntsky:
Reopening as Arbitrary instances may be useful for ToData/FromData roundtrip

Luis-omega:
We want to lower the number of open prs, this pr is just too old and I'm focusing on ODC right now.

Issue #465, opened 2022-05-30
Implement E2E in-browser tests for wallets


klntsky:
Here's an instruction to test extensions with puppeteer that looks very convincing:

https://www.gokatz.me/blog/automate-chrome-extension-testing/

- Investigate how can we apply it to Nami and Gero right away (they should have stable extension IDs)
- Investigate how can we preserve extensions data (wallets and wallet passwords) between test runs

What we need to consider:

- Assuming we are going to have wallets stored in a public repo, someone might steal our testnet ada (not terrible, but annoying)
- waiting for testnet tx confirmations might take a lot of time

ngua:
We can use https://github.com/justinwoo/purescript-toppokki for this. We might need to write a bit of FFI stuff but it shouldn't be too bad (I'm guessing extensions aren't support atm)

PR #456, opened 2022-05-26
Lock Transactions inputs when signing, sign multiple transactions

firefrorefiddle:

PR #502, opened 2022-06-07
Combine examples into a single module

ngua:
We should be running all examples locally when opening PRs (at the very least until #465 is addressed). As it is, it's inconvenient to edit the Makefile or issue all of the commands manually. This adds an `Examples.Combined` module for convenience

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ SHELL := bash
 
 ps-sources := $(shell fd -epurs)
 nix-sources := $(shell fd -enix --exclude='spago*')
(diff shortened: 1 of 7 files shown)


ngua:
Let's add this after #456 is merged, it might simplify things

firefrorefiddle:
Alternatively, we could configure the Webpack server so that each example has it's own URL?

ngua:
@firefrorefiddle If you're willing to deal with webpack I'd be OK with that :)

ngua:
This has been superseded by the approach used in #649

PR #515, opened 2022-06-08
Modify `currencyMPSHash` to not return a Maybe

Luis-omega:
This pr introduces an unsafe internal function.


src/Plutus/Types/CurrencySymbol.js
```diff
@@ -0,0 +1,13 @@
+/* global require exports BROWSER_RUNTIME */
+
+var CardanoWasm;
+if (typeof BROWSER_RUNTIME != 'undefined' && BROWSER_RUNTIME) {
+    CardanoWasm = require('@ngua/cardano-serialization-lib-browser');
+} else {
+    CardanoWasm = require('@ngua/cardano-serialization-lib-nodejs');
+}
+
+exports._unsafeScriptHashFromBytes = input => {
```
Luis-omega, line 10: I have my doubts about introducing this module since a change can be out of sync with the definitions on `src/Serialization/Hash.js`. I introduced this module as I didn't want to export `_unsafeScriptHashFromByte` or its variants from the module `Serialization.Hash` to avoid its abuse.

ngua (changes requested):
I don't think we're gaining anything by introducing the new FFI function, as `currencyScriptHash` just calls it directly anyway. Also `Serialization.Hash.scriptHashFromBytes` would be redundant now, there's no need to have two versions both with and without the `Maybe` context

My recommendation: just remove the `Maybe` FFI helper from the original FFI function

src/Plutus/Types/CurrencySymbol.js
klntsky: Use `Deserialization.FromBytes` typeclass. Add a new instance for `ScriptHash`. This whole module can be removed.

klntsky (changes requested):

src/Plutus/Types/CurrencySymbol.js
Luis-omega: In the end, I deleted it. 

We can't use the `FromBytes` module as we would get an `E r CurrencySymbol` and we want just a plain `CurrencySymbol`, we can get rid of the possibility of error in the signature as the smart constructor for `CurrencySymbol` should already do the checks.

src/Plutus/Types/CurrencySymbol.purs
```diff
@@ -10,28 +10,17 @@ module Plutus.Types.CurrencySymbol
...
-import Aeson
-  ( class DecodeAeson
-  , class EncodeAeson
-  , caseAesonObject
-  , decodeAeson
-  , encodeAeson'
-  , getField
-  , JsonDecodeError(TypeMismatch)
-  )
+import Aeson (class DecodeAeson, class EncodeAeson, caseAesonObject, decodeAeson, encodeAeson', getField, JsonDecodeError(TypeMismatch))
```
klntsky: Please `make format`

klntsky (changes requested):

src/Plutus/Types/CurrencySymbol.purs
Luis-omega: It's already formated, maybe this is from a previous commit?

PR #519, opened 2022-06-08
Tx deserialization test and bugfixes

### LINKED ISSUE #513: Complete `convertCertificate`
`Deserialization.Transaction.convertCertificate` currently contains a hole (see also: `type CertConvHelper`).

We need full support of deserialization at this point.

klntsky:
Closes #513 

There has been a few fixes on the way.

### Pre-review checklist


src/Deserialization/Transaction.js
```diff
@@ -84,13 +91,40 @@ exports._convertCert = certConvHelper => cert => {
     // StakeRegistration,
```
jy14898, line 91: The comment `type CertConvHelper r =` above is outdated

src/Deserialization/Transaction.js
```diff
@@ -84,13 +91,40 @@ exports._convertCert = certConvHelper => cert => {
...
     var r = cert.as_stake_registration();
     if (r) return certConvHelper.stakeRegistration(r.stake_credential());
```
jy14898, line 93: I think it would be better to use a switch block here in conjunction with [CSL.CertificateKind](https://input-output-hk.github.io/cardano-js-sdk/enums/_cardano_sdk_core.CSL.CertificateKind.html) and [Certificate.Kind](https://input-output-hk.github.io/cardano-js-sdk/classes/_cardano_sdk_core.CSL.Certificate.html#kind)

src/Deserialization/Transaction.js
```diff
@@ -167,6 +201,7 @@ exports._unpackCostModel = cm => {
         catch(_){
             err=true;
         }
+        op++;
```
jy14898, line 204: Little bit more idiomatic

```suggestion
    try {
      for(var op = 0;; op++) {
        res.push(cm.get(op).to_str());
      }
    } catch { }
```

src/Deserialization/Transaction.purs
```diff
@@ -275,32 +306,139 @@ convertTxBody txBody = do
...
+    )
+    paramUpdates
   pure
     { epoch
-    , proposedProtocolParameterUpdates: ProposedProtocolParameterUpdates $
+    , proposedProtocolParameterUpdates: T.ProposedProtocolParameterUpdates $
         M.fromFoldable ppus
     }
 
 -- | TODO partially implemented
```
jy14898, line 325: Outdated comment?

src/Deserialization/Transaction.purs
```diff
@@ -709,42 +855,71 @@ foreign import _txBodyNetworkId
...
 
 foreign import _unpackMint
-  :: ContainerHelper -> Csl.Mint -> Array (Tuple ScriptHash Csl.MintAssets)
+  :: ContainerHelper -> Csl.Mint -> Array (ScriptHash /\ Csl.MintAssets)
 
 foreign import _unpackMintAssets
-  :: ContainerHelper -> Csl.MintAssets -> Array (Tuple Csl.AssetName Csl.Int)
+  :: ContainerHelper -> Csl.MintAssets -> Array (Csl.AssetName /\ Csl.Int)
 
 type CertConvHelper (r :: Row Type) =
```
jy14898, line 873: Maybe move to
```haskell
type CertConvHelper' (r :: Type) =
  { stakeDeregistration :: Csl.StakeCredential -> r
  ...
  }

type CertConvHelper (r :: Row Type) = CertConvHelper' (Err r T.Certificate)
```

to be DRY

src/Serialization/Types.js
```diff
@@ -1,2 +1,5 @@
-exports._to_str = x => x.to_str();
-exports._to_bech32 = x => x.to_bech32();
+exports._to_str = x => {
+    return x.to_str();
+};
```
jy14898, line 3: Might be accidental but I prefer the old version

src/Serialization/Types.js
```diff
@@ -1,2 +1,5 @@
...
+exports._vrfKeyHashShow = byteArrayToHex => vrfKeyhash =>
+    byteArrayToHex(vrfKeyhash.to_bytes());
```
jy14898, line 5: Why do the application in JS? ie

```
foreign import vrfKeyHashToBytes :: VRFKeyHash -> RawBytes
```

src/Deserialization/Transaction.purs
```diff
@@ -275,32 +306,139 @@ convertTxBody txBody = do
...
+  -> Err r T.Certificate
+convertPoolRegistration params = do
+  relays <- traverse convertRelay $ poolParamsRelays containerHelper params
+  pure $ T.PoolRegistration
+    { operator: poolParamsOperator params
+    , vrfKeyhash: poolParamsVrfKeyhash params
+    , pledge: poolParamsPledge params
+    , cost: poolParamsCost params
+    , margin: _unpackUnitInterval $ poolParamsMargin params
+    , reward_account: poolParamsRewardAccount params
```
ngua, line 370: Let's change this (and all other occurrences) to `rewardAccount`. Not sure if this got skipped during the great snake_case to camelCase migration, but if we don't change it now we're stuck with it until v2

src/Deserialization/Transaction.js
```diff
@@ -1,3 +1,10 @@
+/* global exports */
```
ngua, line 1: This is a more general comment, but I recall when the `Deserialization.Transaction` module was first introduced we wanted to reduce the amount of logic written in JS

src/Deserialization/Transaction.js
klntsky: We don't use cardano SDK. These are `cardano-serialization-lib` bindings. Here `kind()` outputs a number (no named constants are provided). I don't like using magic numbers in code, plus the overhead is constant and negligible.

jy14898: Those docs (from what I've observed) are 1 to 1, so the enum should be available. It's not really for performance, I think it would make it significantly more readable. We can entirely forgo the as_ and use the value directly

jy14898: (Just did a quick sanity check and `lib.CertificateKind.GenesisKeyDelegation` for eg is indeed available)

jy14898: Ah, I don't think as_* are simple kind checks and type coercion, ie you still need to call it first

src/Deserialization/Transaction.js
```diff
@@ -191,26 +223,21 @@ exports._unpackCostModels = containerhelper => containerhelper.unpackKeyIndexed;
...
-        catch(_){
-            err=true;
-        }
-        op++;
-    }
+    try {
+      for(var op = 0;; op++) {
+        res.push(cm.get(op).to_str());
+      }
+    } catch (_) { }
```
jy14898, line 232: The `(_)` in catch is optional, up to you though as it makes little difference

src/Deserialization/Transaction.js
```diff
@@ -191,26 +223,21 @@ exports._unpackCostModels = containerhelper => containerhelper.unpackKeyIndexed;
...
     return res;
 };
 
 // foreign import _convertLanguage
 //   :: forall r.ErrorFfiHelper r -> { plutusV1:: Language } -> CSL.Language -> E r Language
 exports._convertLanguage = errorHelper => langCtors => cslLang => {
     try{
-        if(cslLang.kind()==0){
+        if(cslLang.kind()==lib.LanguageKind.PlutusV1){
```
jy14898, line 240: Nice catch

jy14898 (approved):
Looks good to me

src/Deserialization/Transaction.js
```diff
@@ -53,8 +68,6 @@ exports._txBodyRequiredSigners = maybeGetterMulti("required_signers");
 // network_id(): number | void;
 exports._txBodyNetworkId = maybeGetter_(o => o.kind())("network_id");
```
jy14898, line 69: Just did a quick check for any more `kind`s, we can use NetworkIdKind here

jy14898 (changes requested):

ngua (approved):
LGTM

Issue #462, opened 2022-05-27
Consider using faster blake2 implementations


klntsky:
In #448 client side hashing has been added. It uses a pure JS implementation which is very portable, but is not as performant as native solutions.

Alternatives: 

- https://github.com/nazar-pc/blake2.wasm for the browser
- https://github.com/vrza/node-blake2 for the node

Since we already have conditional compilation, it would not be hard to eventually adapt these

PR #520, opened 2022-06-08
Client-side hashing: Use `blake2b-wasm` package for better performance

errfrom:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/462


src/Hashing.js
```diff
@@ -1,32 +1,67 @@
...
+const withBlake2bWasmObject = bytesToHash => digestLength => {
+  let obj = Blake2bWasm(digestLength);
+  obj.update(Buffer.from(bytesToHash));
+  return obj;
+};
+
+const blake2bHash = bytesToHash => digestLength => {
+  return new Promise(resolve => {
+    Blake2bWasm.ready(error => {
+      if (error || !Blake2bWasm.SUPPORTED) {
```
klntsky, line 28: Does SUPPORTED mean exactly that WASM is available? Or is the condition when SUPPORTED is `true` more strict than that?

Since CSL depends on WASM anyway, and we can't meaningfully use the library without it, we can just drop the pure JS implementation from the bundle.

klntsky (changes requested):

src/Hashing.js
errfrom: @klntsky Yes, SUPPORTED is only used to check if WASM is available, so I'll drop the pure js part

errfrom: @klntsky But on the other hand, we have to check if there is no error, so maybe it would be reasonable to have a fallback for this case?

klntsky (approved):

PR #527, opened 2022-06-10
Use `psa` for compilation in Nix derivations

### LINKED ISSUE #518: Use `psa` for compilation
This way we can censor warnings from dependencies during compilation

ngua:
Closes #518. Now we can compile with `psa --strict --censor-lib` which will silence the superfluous warnings from libraries and cause building to fail if any compiler warnings are reported (very helpful to catch things like redundant imports, etc...)

### Pre-review checklist

- [-] All Purescript imports are explicit, including constructors (n/a)
- [-] **All** existing examples have been run locally against a fully-synced testnet node (n/a)
- [-] The integration and unit tests have been run (`npm run test`) locally (n/a, except I ran the `checks.ctl-unit-test` via Nix to make sure it passes (Hydra is having issues)

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -71,35 +71,6 @@ let
           + shellHook;
       };
 
-  buildPursDocsSearch =
-    { name ? "purescript-docs-search"
-    , ...
-    }:
-    pkgs.stdenv.mkDerivation {
-      inherit name;
-      srcs = [
-        (pkgs.fetchurl {
-          url = "https://github.com/purescript/purescript-docs-search/releases/download/v0.0.11/docs-search-app.js";
-          sha256 = "17qngsdxfg96cka1cgrl3zdrpal8ll6vyhhnazqm4hwj16ywjm02";
-        })
-        (pkgs.fetchurl {
-          url = "https://github.com/purescript/purescript-docs-search/releases/download/v0.0.11/purescript-docs-search";
-          sha256 = "1hjdprm990vyxz86fgq14ajn0lkams7i00h8k2i2g1a0hjdwppq6";
-        })
-      ];
-      buildInputs = [ nodejs ];
-      unpackPhase = ''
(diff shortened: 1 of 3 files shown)


nix/default.nix
```diff
@@ -164,6 +142,73 @@ let
...
+      });
+
+  buildPursDocsSearch =
+    { name ? "purescript-docs-search"
+    , ...
+    }:
+    let
+      docsUrl =
+        "https://github.com/purescript/purescript-docs-search/releases/download";
+      docsVersion = "v0.0.11";
```
bladyjoker, line 185: Consider using flake inputs for this and then `dependency inject` this like:

```
    purs-docs-search-app = {
      url  = "https://github.com/purescript/purescript-docs-search/releases/download/v0.0.11/docs-search-app.js";
      flake = false;
      };
    purs-docs-search = {
      url  = "https://github.com/purescript/purescript-docs-search/releases/download/v0.0.11/purescript-docs-search";
      flake = false;
      };
```
But I understand the convenience of the current approach, so which ever you find nicer...

nix/default.nix
```diff
@@ -134,7 +111,8 @@ let
           install-spago-style
         '';
       buildPhase = ''
-        build-spago-style "./**/*.purs"
+        psa --strict --censor-lib --is-lib=.spago ${spagoGlobs} \
+          --censor-codes=UserDefinedWarning "./**/*.purs"
```
bladyjoker, line 115: This is great! I wish psa had config file to manage reporting like hlint does. Consider adding `censorCodes` argument with `[ "UserDefinedWarning" ]` as default

bladyjoker (approved):
Added some comments, mostly nice-to-haves. Didn't know PSA was a thing, very useful!

nix/default.nix
ngua: Yes, that would be ideal

PR #564, opened 2022-06-16
Implement getProtocolParameters

### LINKED ISSUE #541: Get protocol parameters from Ogmios
Currently we have a `ProtocolParametersAlonzo` module, but we should get the params dynamically from Ogmios instead

klntsky:
Closes #541 #566

### Pre-review checklist


test/ProtocolParams.purs
```diff
@@ -0,0 +1,31 @@
...
+import Mote (group, test)
+import Node.Encoding (Encoding(UTF8))
+import Node.FS.Aff (readTextFile)
+import QueryM.Ogmios (ProtocolParametersR)
+import Test.Spec.Assertions (shouldSatisfy)
+import TestM (TestPlanM)
+
+suite :: TestPlanM Unit
+suite = do
+  eAeson <- lift $ parseJsonStringToAeson <$> readTextFile UTF8
```
bladyjoker, line 22: You can use Utils.readAeson

bladyjoker (approved):
This looks good to me, let me know if you need me to look closer at certain bits.

src/Cardano/Types/Transaction.purs
```diff
@@ -374,6 +384,21 @@ derive instance Eq Nonce
...
+instance DecodeAeson Nonce where
+  decodeAeson aeson = (HashNonce <$> decodeAeson aeson) <|>
+    caseAesonString err
+      ( case _ of
+          "neutral" -> pure IdentityNonce
+          _ -> err
+      )
+      aeson
+    where
+    err = Left (TypeMismatch "Nonce")
```
ngua, line 396: Type signature

src/QueryM/Ogmios.purs
```diff
@@ -432,6 +447,248 @@ instance DecodeAeson TxEvaluationR where
...
+    caseAesonString (Left err)
+      \string -> do
+        case String.split (Pattern "/") string of
+          [ numeratorStr, denominatorStr ] -> note err do
+            numerator <- BigInt.fromString numeratorStr
+            denominator <- BigInt.fromString denominatorStr
+            PParamRational <$> numerator % denominator
+          _ -> Left err
+    where
+    err = TypeMismatch "PParamRaional"
```
ngua, line 473: Type signature

ngua (commented):
I think we need to remove the `ProtocolParametersAlonzo` file and get the protocol parameters from Ogmios during contract initialization. We can do that in a separate PR if you'd like

src/QueryM/Ogmios.purs
```diff
@@ -432,6 +447,248 @@ instance DecodeAeson TxEvaluationR where
...
+  , "monetaryExpansion" :: PParamRational
+  , "treasuryExpansion" :: PParamRational
+  , "decentralizationParameter" :: PParamRational
+  , "extraEntropy" :: Nonce
+  , "protocolVersion" ::
+      { "major" :: BigInt
+      , "minor" :: BigInt
+      }
+  , "minPoolCost" :: BigInt
+  , "coinsPerUtxoWord" :: BigInt
```
ngua, line 495: Some of these could be `Coin`s, like this field

src/QueryM/Ogmios.purs
```diff
@@ -432,6 +447,248 @@ instance DecodeAeson TxEvaluationR where
...
+        case String.split (Pattern "/") string of
+          [ numeratorStr, denominatorStr ] -> note err do
+            numerator <- BigInt.fromString numeratorStr
+            denominator <- BigInt.fromString denominatorStr
+            PParamRational <$> numerator % denominator
+          _ -> Left err
+    where
+    err = TypeMismatch "PParamRaional"
+
+type ProtocolParameters =
```
ngua, line 475: I know that this type is easier for us to decode, but I'm not sure if it's the most useful representation for users. If they're expecting something like Plutus' `ProtocolParameters`, returning this type will at the very least oblige them to read the source code or the Ogmios documentation

Another consideration is that Haskell developers may not know that arbitrary strings can be used as record fields in PS (and may not know how to access the fields in that case)

We can keep the easier-to-parse structure, and then convert it in a function that uses field names that correspond more closely to what Plutus gives us. This could also let us adjust some of the types of each field (for example a normal `Rational` instead of `PParamRational` and `Coin`s instead of `BigInt`s in some places). I know that requires a non-trivial amount of boilerplate, but I think it's better that we shoulder such costs instead of users

klntsky:
I checked that we didn't change the costmodels using this code:

```purescript

testGetProtocolParameters :: Aff Unit
testGetProtocolParameters = do
  pparams <- flip runQueryM (getProtocolParameters) =<<
    traceQueryConfig
  when ((pparams # unwrap >>> _.costModels) /= costModelsFixture1) do
    liftEffect $ throw "Incorrect decoding of costModels - this may fail in the future"
```

Issue #570, opened 2022-06-17
`feeTooSmall` error when submitting a tx with `scriptDataHash` set to `Nothing`


errfrom:
After moving the calculation and setting of the script integrity hash to the client, we encountered a problem where we get a `feeTooSmall` error when trying to submit a transaction with `scriptDataHash` set to `Nothing`. 

The current workaround is to explicitly set `scriptIntegrityHash` to `SNothing` on the server:
```Haskell
withScriptIntegrityHash :: C.Tx C.AlonzoEra -> C.Tx C.AlonzoEra
withScriptIntegrityHash transaction@(C.Tx txBodyAlonzo keyWits) =
  case txBodyAlonzo of
    Shelley.ShelleyTxBody era txBody scr sData aux val ->
      case Tx.scriptIntegrityHash txBody of
        SNothing ->
          let txBody' = txBody {Tx.scriptIntegrityHash = SNothing}
               shelleyB = Shelley.ShelleyTxBody era txBody' scr sData aux val
           in C.Tx shelleyB keyWits
        _ -> transaction
```

Steps to reproduce:

1. [Server -> `Api.Handlers` -> `estimateTxFees`] Replace this line 
```Haskell
fee = estimateFee pparams witCount (withScriptIntegrityHash tx')
```
with 
```Haskell
fee = estimateFee pparams witCount tx'
```
2. Run `Examples.Pkh2Pkh`

klntsky:
How can we reproduce it? I suspect it might be fixed by #866

Issue #449, opened 2022-05-24
Deprecate signTxBytes


klntsky:
We want to be able to remove signTxBytes. For that we need to be sure that tx serialization/deserialization roundtrips for finalized transactions.

After the change we should be able to modify the finalized tx as CTL data and serialize it back.

ngua:
Closed by #573

PR #573, opened 2022-06-20
Remove `finalizeTx` and `signTxBytes` and all related types

### LINKED ISSUE #273: Get CSL to work with datums, script integrity hashes, etc... and deprecate server endpoint
I've started working on this while we're still on CSL v10.x since this hopefully won't complicate the CSL upgrade too much

I've added a finalization function for the frontend and I've started testing the examples (checked means it's working):

- [x] `Pkh2Pkh`
- [x] `AlwaysMints`
- [x] `AlwaysSucceeds`

ngua:
Closes #273 and #449. Unfortunately there is still this issue: #570, but we have a plan for that

### Pre-review checklist


src/Transaction.purs
```diff
@@ -44,29 +48,60 @@ derive instance Eq ModifyTxError
 instance Show ModifyTxError where
   show = genericShow
 
--- | Set the `Transaction` body's script data hash. NOTE: Must include all of
+-- | Sets the script integrity hash and attaches redeemers, for use after
+-- | reindexing
+finalizeTransaction
```
ngua, line 53: Should we have an `UnattachedReindexedTx` or something for this? `UnbalancedUnattachedTx` doesn't work because the transaction might be balanced

bladyjoker (commented):
@ngua since this is a very deep change, I feel I should test Cardax contracts with this branch to double check if we missed something?

ngua:
Now that the hard-coded `Costmdls` have been removed, I need to find a solution for that. I think it might be good to have a `Ref (Maybe ProtocolParameters)` inside the `ContractConfig`

klntsky:
@ngua should we care about Costmdls changes during the life of our runtime?

ngua:
@klntsky Not really. My other idea was to use CSL's `default_cost_models` function, although this will need revising once we support `PlutusV2` scripts

src/Transaction.purs
klntsky: Good idea

ngua:
I see that we have #612, so I'll leave that work for another PR

Issue #479, opened 2022-06-01
Parse ogmios, ODC and ctl-server errors


klntsky:
As for now, we ignore "fault" field in ogmios response, which results in parse errors like `(AtKey "SubmitSuccess" MissingValue)`. API consumers asked us to handle errors more gracefully a number of times.

What to do:

For ODC:

Check `fault` field of response.

For Ogmios:

Handle "fault" field in response.

For ctl-server:

Make sure that all exceptions are handled and the output is actually JSON, not something like

```
The Plutus script evaluation failed: An error has occurred:  User error:
The provided Plutus code called 'error'.
Script debugging logs: 
```

Since we don't have designated types for errors, let's have a "catch all" case and return just the error message as JSON.

klntsky:
Closed by #586 #584

PR #586, opened 2022-06-21
Use catchAll in server

klntsky:
Part of #479

### Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ SHELL := bash
 
 ps-sources := $(shell fd -epurs)
 nix-sources := $(shell fd -enix --exclude='spago*')
+hs-sources := $(shell fd . './server/src' './server/exe' -ehs)
 ps-entrypoint := Examples.AlwaysSucceeds
 ps-bundle = spago bundle-module -m ${ps-entrypoint} --to output.js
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -17,10 +18,14 @@ run-build:
 	@${ps-bundle} && BROWSER_RUNTIME=1 webpack --mode=production
 
 check-format:
-	@purs-tidy check ${ps-sources} && nixpkgs-fmt --check ${nix-sources}
+	@purs-tidy check ${ps-sources}
+	nixpkgs-fmt --check ${nix-sources}
+	fourmolu -m check -o -XTypeApplications -o -XImportQualifiedPost ${hs-sources}
```
bladyjoker, line 23: I think the `at` symbol is to quiet the command, perhaps you want to do that as well for all of them?

server/src/Api.hs
```diff
@@ -115,6 +114,8 @@ appServer env = hoistServer api appHandler server
             throwError err400 {errBody = LC8.pack ih}
           OtherDecodeError str ->
             throwError err400 {errBody = LC8.pack str}
+        handleError (ErrorCall err) =
+          throwError err400 {errBody = LC8.pack $ show err}
```
bladyjoker, line 118: Curious why we're not returning Json errors by simply having an instance for `CtlServerError`?

bladyjoker (approved):
I didn't check, but if we don't have a test triggering this new particular error condition maybe we could add one?

server/src/Api.hs
klntsky: These are only for the users (they are not recoverable), thus they don't have to be in a machine-readable format

bladyjoker: That's ok, a few cents tho...

Generally speaking, you don't make logs and error messages machine readable just because they are recoverable, you make it so monitoring and analytics stack can parse them out easily. That's why a combination of syslog+machine readable format for logs and error messages are a good thing to introduce early on.

For instance, CTL server is just one runtime component in our infrastructure, we generally want to introduce log collection and analysis for all these services, this whole process is predicated on the assumption that collected information can be easily parsed.

PR #603, opened 2022-06-23
Better handling of first-time WS connections + Fix reconnecting behavior

### LINKED ISSUE #591: Handle first-time connection error (websockets)
In `mkOgmiosWebSocket'` we set the error handler only in `_onWsConnect`, that is, after the first connection is successful.
We need to throw an error to the user when the first connection attempt is not successful, BUT we should not throw the error on subsequent failing attempts (because we use `reconnecting-websocket` specifically to deal with network problems)

klntsky:
Closes #591 

### Pre-review checklist


bladyjoker (approved):
This looks good! I think it'd be possible to test this by just connecting to a non-existent IP:Port?

PR #604, opened 2022-06-23
Implement waitUntilSlot

### LINKED ISSUE #601: Implement `waitUntilSlot`
We can repeatedly get chain tip via ogmios until it's >= required slot number

The function should go to `QueryM` and `Contract.Chain` modules

klntsky:
Closes #601

### Pre-review checklist


klntsky:
This is obviously not ideal because we are querying too often without even trying to calculate next query time.

Initially I wanted to approximate time for the required slot and query slightly before that, but without knowing much about guarantees that cardano provides for slot times, I can't choose an appropriate time estimation algorithm.

Needs investigation.

bladyjoker (approved):
Good stuff!

src/QueryM.purs
```diff
@@ -252,6 +254,15 @@ getChainTip = ogmiosChainTipToTip <$> mkOgmiosRequest Ogmios.queryChainTipCall
     Ogmios.CtChainPoint { slot, hash } -> Chain.Tip $ wrap
       { slot, blockHeaderHash: wrap $ unwrap hash }
 
+waitUntilSlot :: AbsSlot -> QueryM Chain.Tip
+waitUntilSlot requiredSlot =
+  getChainTip >>= case _ of
+    tip@(Chain.Tip (Chain.ChainTip { slot }))
+      | slot >= requiredSlot -> pure tip
+    _ -> do
+      liftAff $ delay (wrap 1000.0)
```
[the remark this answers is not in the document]
ngua: I agree that using a `Slot`/`POSIXTime` conversion here would be ideal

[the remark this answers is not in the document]
klntsky: Slot <-> posix time conversion gives estimate times, it's not guaranteed that at some specific time ogmios will return specific chain tip, right? So the question is, how do we correct for that without adding extra waiting time.

I was thinking about something like:

```
waitUntilSlot requiredSlot = 
IF requiredSlot - slot < N
THEN start querying chain tip each second
ELSE
  delay (slotToPosixTime (requiredSlot - slot - N))  
  waitUntilSlot
```

where N is arbitrary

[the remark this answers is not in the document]
bladyjoker: I believe slotToPosixTime deals with absolute time in terms of PosixTime and Slots, perhaps just using [slotLength](https://github.com/Plutonomicon/cardano-transaction-lib/blob/8b1b1062d57db11efc7aa5e5ddbe305f8798c55c/src/Types/Interval.purs#L902) would be better in this scenario as we're trying to determine `how long to wait in terms of milliseconds`.

However, you're right, slot duration to posix duration is generally an estimate, so the issue still remains. I agree with using `N` as you suggested, perhaps naming it differently `timePadding`? I don't know, I suck at naming x)

Also consider `s/requiredSlot/futureSlot` and `s/slot/currentSlot` and their difference `slotsToWait`

test/AffInterface.purs
```diff
@@ -127,6 +130,15 @@ testGetChainTip :: Aff Unit
 testGetChainTip = do
   flip runQueryM (void getChainTip) =<< traceQueryConfig
 
+testWaitUntilSlot :: Aff Unit
```
adamczykm, line 133: It's not immediately clear what this test does. Could you add a docstring?

src/QueryM.purs
```diff
@@ -252,6 +254,15 @@ getChainTip = ogmiosChainTipToTip <$> mkOgmiosRequest Ogmios.queryChainTipCall
     Ogmios.CtChainPoint { slot, hash } -> Chain.Tip $ wrap
       { slot, blockHeaderHash: wrap $ unwrap hash }
 
+waitUntilSlot :: AbsSlot -> QueryM Chain.Tip
```
adamczykm, line 257: Can we have more commentary on this function?

adamczykm: Also it could contain a note about the current excessive querying so that we don't forget about it eventually

adamczykm (commented):
Left some comments - mostly about documenting the code.

src/QueryM/WaitUntilSlot.purs
```diff
@@ -0,0 +1,158 @@
...
+      <> show (unwrap result)
+      <> "ms"
+  pure result
+  where
+  nonNegative :: Number -> Number
+  nonNegative n
+    | n < 0.0 = 0.0
+    | otherwise = n
+
+posixTimeToSeconds :: POSIXTime -> QueryM Seconds
```
vvtran, line 153: Does this computation need to be in `QueryM`, can we keep it pure?

vvtran (commented):
@klntsky  I may have missed out on development to this previously, I wonder if we can do this without relying too heavily on Ogmios e.g.
Use `Instant` to get the current posix time, we convert the input slot (absSlot in this case) to posix time with `slotToPosixTime` (this will be exact), and we can just delay for `slotToPosixTime slot - instantTime` (assuming it's positive), we don't need to query ChainTIp do we? The return type could just be `Unit` instead of chaintip?

I guess we'd have to add a warning to not go too far into the future, as the `EraSummary` may change?

Edit:  this runs the risk of fragmenting our logic, there will be a reliance that our current JS POSIX time is "better" than any current slot functionality provided by Ogmios.

klntsky:
@vvtran 
> The return type could just be Unit instead of chaintip?

Most of the time we return a slot value that is higher than requested, so I think it may be useful for the applications to check if we are actually not very far in the future from the requested slot - without the need to call `getChainTip` again.

klntsky:
There's some problem with ogmios related to the hardfork: the chain tip slot does not increase after 62510369

ngua:
@jy14898 Ogmios is probably adding endpoints that expose a time interpreter in the next release

ngua (approved):

Issue #565, opened 2022-06-17
Upgrade to CSL v11.0.0-beta.1


ngua:
Support has been merged for Babbage-era transactions: https://github.com/Emurgo/cardano-serialization-lib/pull/463

PR #609, opened 2022-06-24
Upgrade to CSL v11.0.0-beta.1

errfrom:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/565

Changes:
1. Rewrote `Plutus.Conversion.Address` code using CSL functionality.
 Some conversion functions now return the result without the `Maybe` context:
 ```PureScript 
 fromPlutusAddress :: Csl.NetworkId -> Plutus.Address -> Csl.Address
 fromPlutusAddressWithNetworkTag :: Plutus.AddressWithNetworkTag -> Csl.Address
 fromPlutusTxOutput :: NetworkId -> Plutus.TransactionOutput -> Cardano.TransactionOutput
 fromPlutusTxUnspentOutput :: NetworkId -> Plutus.TransactionUnspentOutput -> Cardano.TransactionUnspentOutput
 fromPlutusUtxoM :: NetworkId -> Plutus.UtxoM -> Cardano.UtxoM
 ```
2. Changed the underlying type of `Slot`, `TransactionIndex` and `CertificateIndex` to `BigNum` and utilized new CSL functions that work with `SlotBigNum` instead of the deprecated `Slot32` (`u32`).

3. Removed protocol param update setters for the "decentralisation constant" (`set_d`) and the "extra entropy" (`set_extra_entropy`), since these two parameters are removed in the Babbage HF.

### Pre-review checklist


ngua:
@errfrom You should open this PR against the `vasil` branch

src/Types/BigNum.purs
```diff
@@ -0,0 +1,117 @@
...
+fromString :: String -> Maybe BigNum
+fromString = _fromString maybeFfiHelper
+
+fromStringUnsafe :: String -> BigNum
+fromStringUnsafe str = unsafePartial fromJust $ fromString str
+
+foreign import toString :: BigNum -> String
+
+maxValue :: BigNum
+maxValue = fromStringUnsafe "18446744073709551615"
```
vvtran, line 117: I'm curious about this value in comparison to https://ogmios.dev/api/ if you look under `QueryResponse[eraSummaries]` to see `Integer[ 0 .. 18446744073709552000 ]` which is i.e. a difference of `385`. I wonder if we can increase this and coincide it with `AbsSlot` essentially.

test/Plutus/Time.purs
```diff
@@ -192,15 +191,15 @@ suite = do
...
       toFromAesonTest "PosixTimeBeforeSystemStart" $ PosixTimeBeforeSystemStart
         posixTimeFixture
       toFromAesonTest "StartTimeGreaterThanTime" $ StartTimeGreaterThanTime
         absTimeFixture
       toFromAesonTest "EndSlotLessThanSlotOrModNonZero" $
         EndSlotLessThanSlotOrModNonZero absSlotFixture modTimeFixture
-      toFromAesonTest "CannotConvertAbsSlotToSlot" $ CannotConvertAbsSlotToSlot
-        absSlotFixture
+      -- toFromAesonTest "CannotConvertAbsSlotToSlot" $ CannotConvertAbsSlotToSlot
+      --   absSlotFixture
```
vvtran, line 202: What's the reason for removing these tests?

vvtran (commented):
We may want a separate PR for this, but given we're upgrading `Slot` to `BigNum`, do we also want to deprecate `AbsSlot` and conversions to and from? Essentially replacing all occurrences of `AbsSlot` by `Slot`? I added `AbsSlot` originally to deal with the precision of Ogmios `eraSummaries` call which `Slot` couldn't handle (whilst retaining `Slot UInt`) but we probably want to get rid of it sooner rather than later so prevent any further usage/dependencies. Also see my point on on the `maxValue`.

src/Types/BigNum.purs
errfrom: @vvtran `BigNum` has Rust `u64` as underlying type, which max value is `18446744073709551615` (https://doc.rust-lang.org/std/primitive.u64.html). For instance, ```BigNum.maxValue `BigNum.add` BigNum.one``` yields `Nothing`. I wonder if we could just ignore this difference and deprecate `AbsSlot` despite of this?

test/Plutus/Time.purs
errfrom: @vvtran I had originally planned to use `BigInt` for `Slot` and removed these tests because the conversion could never fail. Restored them, thanks for pointing this out.

src/Serialization/Address.purs
```diff
@@ -100,11 +107,12 @@ import Serialization.Hash (Ed25519KeyHash, ScriptHash)
 import Serialization.Types (Bip32PublicKey)
 import ToData (class ToData, toData)
 import Types.Aliases (Bech32String, Base58String)
+import Types.BigNum (BigNum)
 import Types.ByteArray (ByteArray)
 import Types.CborBytes (CborBytes)
 import Types.PlutusData (PlutusData(Bytes))
 
-newtype Slot = Slot UInt
+newtype Slot = Slot BigNum
```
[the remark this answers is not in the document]
errfrom: @vvtran It would be nice to deprecate `AbsSlot`, but I have some doubts regarding that small max value difference between `BigNum` and Ogmios Integer (see my comment on the max value).

src/Types/BigNum.purs
vvtran: I'm inclined to think it's fine to ignore the difference. It's minor relative the size of the max slot, i.e. `18446744073709551615` is a huge slot number already. Unless there change the slotLength to do something weird, I think it's okay. I would add a comment on this though, at least under the Ogmios query?
Any thoughts @ngua @klntsky ?

src/Serialization/Address.purs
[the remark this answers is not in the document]
vvtran: @errfrom  I replied to the other post

ngua:
I'd prefer not to hold this up on the `AbsSlot` question -- can we open another issue for that instead @vvtran @errfrom?

vvtran:
I think that's reasonable, I'll open an issue @ngua

src/Types/BigNum.purs
vvtran: https://github.com/Plutonomicon/cardano-transaction-lib/issues/632

vvtran:
@errfrom  do let me know when it's ready for the review again as this is a priority, importantly, we should  check the examples still work for Alonzo testnet.

errfrom:
@vvtran All examples except `AlwaysSucceeds` work, on which I get a `feeToSmall` error, investigating it now

src/Types/BigNum.purs
klntsky: I agree, let's ignore the difference and add a comment

server/src/Api/Handlers.hs
```diff
@@ -86,7 +86,7 @@ estimateTxFees FeesRequest {count, tx} = do
               -- be able to handle a possible integer overflow
               feeBytes =
                 bytesNeeded (fee + bytesNeeded fee * feePerByte)
-           in fee + feeBytes * feePerByte
+           in fee + feeBytes * feePerByte + 2 * feePerByte -- FIXME:
```
klntsky, line 89: Is there an issue?

src/BalanceTx.purs
```diff
@@ -437,36 +452,36 @@ balanceTx unattachedTx@(UnattachedUnbalancedTx { unbalancedTx: t }) = do
...
-    let
-      unattachedTx'' = unsignedTx ^. _transaction'
-        /\ unsignedTx ^. _redeemersTxIns
-      -- Sort inputs at the very end so it behaves as a Set.
-      sortedUnsignedTx = fst unattachedTx'' # _body <<< _inputs %~ Array.sort
-    -- Logs final balanced tx and returns it
-    logTx "Post-balancing Tx " allUtxos sortedUnsignedTx
-    except $ Right (unattachedTx'' # _1 .~ sortedUnsignedTx)
+    -- Sort inputs at the very end so it behaves as a Set:
+    let sortedUnsignedTx = unsignedTx # _body' <<< _inputs %~ Array.sort
```
klntsky, line 469: Is it possible to use Set instead?

ngua (approved):
Thank you for taking care of this @errfrom

server/src/Api/Handlers.hs
ngua: I'll open one and add a link here

src/BalanceTx.purs
vvtran: I've been thinking about this for a long time, I will create an issue for this and can open discussion.

PR #610, opened 2022-06-24
get rid of needing docker for launching ctl-runtime-deps

marijanp:
Closes # .
In this PR I introduce a `startPostgres` and a `startPlutipServer` function, which enable us to get rid of the need to manually start a `plutip-runtime` and `plutip-server.
One is now able to just run the test without any other steps being required. A developer can just
```
nix develop
spago test -m Plutip.Server
```

### Pre-review checklist

### DIFF
--- a/src/Plutip/Server.purs
+++ b/src/Plutip/Server.purs
@@ -22,6 +22,7 @@ import Contract.Monad (Contract, ContractConfig(ContractConfig), runContract)
 import Control.Monad.Error.Class (withResource)
 import Control.Monad.Except (runExcept)
 import Control.Monad.Reader (ReaderT(ReaderT), runReaderT)
+import Data.Array (head)
 import Data.Bifunctor (lmap)
 import Data.Either (Either(Right, Left), either)
 import Data.Foldable (intercalate)
@@ -30,31 +31,43 @@ import Data.Int as Int
 import Data.Maybe (Maybe(Just, Nothing))
 import Data.Newtype (wrap)
 import Data.Posix.Signal (Signal(SIGINT))
+import Data.String.Gen (genAlphaString)
 import Data.UInt as UInt
 import Data.YAML.Foreign.Decode (parseYAMLToJson)
 import Effect (Effect)
-import Effect.Aff (Aff)
+import Effect.Aff (Aff, Milliseconds(..), delay)
 import Effect.Aff.Class (liftAff)
 import Effect.Class (liftEffect)
+import Effect.Console (log)
 import Effect.Exception (throw)
 import Helpers (fromJustEff, fromRightEff)
-import Node.ChildProcess (ChildProcess, defaultSpawnOptions, kill, spawn)
+import Node.Buffer as Bf
+import Node.ChildProcess
+  ( ChildProcess
+  , defaultExecSyncOptions
+  , defaultSpawnOptions
+  , execSync
+  , kill
+  , spawn
+  )
 import Node.Encoding as Encoding
 import Node.FS.Aff (writeTextFile)
 import Node.FS.Sync (readTextFile)
 import Node.Path (concat, dirname)
 import Plutip.Types
-  ( ClusterStartupRequest(ClusterStartupRequest)
+  ( ClusterStartupParameters
+  , ClusterStartupRequest(ClusterStartupRequest)
   , FilePath
   , PlutipConfig
   , StartClusterResponse(ClusterStartupSuccess, ClusterStartupFailure)
   , StopClusterRequest(StopClusterRequest)
   , StopClusterResponse
-  , ClusterStartupParameters
+  , PostgresConfig
   )
 import Plutip.Utils (tmpdir)
 import QueryM (ClientError(..))
 import QueryM as QueryM
+import Test.QuickCheck.Gen (randomSample)
 import Types.UsedTxOuts (newUsedTxOuts)
 
 type PlutipM (r :: Row Type) (a :: Type) = ReaderT PlutipConfig (Contract r) a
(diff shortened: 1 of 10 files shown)


src/Plutip/Server.purs
```diff
@@ -30,31 +31,43 @@ import Data.Int as Int
 import Data.Maybe (Maybe(Just, Nothing))
 import Data.Newtype (wrap)
 import Data.Posix.Signal (Signal(SIGINT))
+import Data.String.Gen (genAlphaString)
 import Data.UInt as UInt
 import Data.YAML.Foreign.Decode (parseYAMLToJson)
 import Effect (Effect)
-import Effect.Aff (Aff)
+import Effect.Aff (Aff, Milliseconds(..), delay)
```
klntsky, line 38: Please use explicit constructor imports,

src/Plutip/Server.purs
```diff
@@ -162,8 +181,60 @@ startOgmios cfg params = liftEffect $ spawn "ogmios" ogmiosArgs
...
+startPlutipServer cfg = do
+  p <- liftEffect $ spawn "plutip-server" [ "-p", UInt.toString cfg.port ]
+    defaultSpawnOptions { detached = true }
+  delay $ Milliseconds 100.0
+  pure p
+
+startPostgresServer
+  :: PostgresConfig -> ClusterStartupParameters -> Aff ChildProcess
+startPostgresServer pgConfig _ = do
+  randomStr <- liftEffect $ fromJustEff "hello" =<< head <$> randomSample
```
klntsky, line 197: Let's use `QueryM.UniqueId` for that. genAlphaString gives no guarantees about string length, which can cause problems.

src/Plutip/Server.purs
```diff
@@ -162,8 +181,60 @@ startOgmios cfg params = liftEffect $ spawn "ogmios" ogmiosArgs
...
+    genAlphaString
+  tmpDir <- liftEffect tmpdir
+  let
+    workingDir = tmpDir <> "/" <> randomStr
+    databaseDir = workingDir <> "/postgres/data"
+  liftEffect $ void $ execSync ("initdb " <> databaseDir) defaultExecSyncOptions
```
klntsky, line 203: is `initdb` from the same nix package as `psql`?

src/Plutip/Server.purs
```diff
@@ -162,8 +181,60 @@ startOgmios cfg params = liftEffect $ spawn "ogmios" ogmiosArgs
...
+    , databaseDir
+    , "-p"
+    , UInt.toString pgConfig.port
+    , "-h"
+    , pgConfig.host
+    , "-k"
+    , workingDir <> "/postgres"
+    ]
+    defaultSpawnOptions
+  delay (Milliseconds 3000.0)
```
klntsky, line 215: I'd like to use [ShareStream](https://pursuit.purescript.org/packages/purescript-node-child-process/9.0.0/docs/Node.ChildProcess#v:ShareStream) on the process stdout to catch the moment it starts listening on port, if that's possible. Please add a TODO in comments and create a new issue (the TODO should include a link to it).

klntsky (changes requested):

klntsky:
Overall - great, this is really making us deliver sooner!

src/Plutip/Server.purs
marijanp: yes, everything is available through `pkgs.postgresql`

src/Plutip/Server.purs
```diff
@@ -162,8 +189,70 @@ startOgmios cfg params = liftEffect $ spawn "ogmios" ogmiosArgs
...
+    defaultSpawnOptions
+  void $ recovering defaultRetryPolicy ([ \_ _ -> pure true ])
+    $ const
+    $ liftEffect
+    $ execSync
+        ( "psql -h " <> pgConfig.host <> " -p " <> UInt.toString pgConfig.port
+            <> " -d postgres"
+        )
+        defaultExecSyncOptions
+  -- liftEffect $ log =<< Bf.toString Encoding.UTF8  =<< execSync ("createuser --createdb --superuser " <> pgConfig.user <> " -h '" <> pgConfig.host <> "' -p " <> UInt.toString pgConfig.port) defaultExecSyncOptions
```
klntsky, line 234: dead comment

test/Plutip.purs
```diff
@@ -55,21 +58,22 @@ config =
...
       { port: UInt.fromInt 8081
-      , host: "localhost"
+      , host: "127.0.0.1"
       , secure: false
       }
-  , ogmiosDatumCachePostgresConfig:
-      { port: UInt.fromInt 5432
+  , postgresConfig:
+      { host: "127.0.0.1"
+      , port: UInt.fromInt 5432
```
klntsky, line 76: Please change the port to something else. It conflicts with ctl-runtime by default and I get a password prompt.

klntsky (changes requested):

test/Plutip.purs
marijanp: did that

klntsky:
Let's silence `psql` output.

src/Plutip/Server.purs
```diff
@@ -162,8 +189,69 @@ startOgmios cfg params = liftEffect $ spawn "ogmios" ogmiosArgs
...
+
+startPlutipServer :: PlutipConfig -> Aff ChildProcess
+startPlutipServer cfg = do
+  p <- liftEffect $ spawn "plutip-server" [ "-p", UInt.toString cfg.port ]
+    defaultSpawnOptions { detached = true }
+  void
+    $ recovering defaultRetryPolicy
+        ([ \_ _ -> pure true ])
+    $ const
+    $ stopPlutipCluster cfg
```
klntsky, line 203: I don't understand why this call is needed here. `plutip-server` can only see it's own clusters. If it hasn't been started, it will now stop.

klntsky (changes requested):

src/Plutip/Server.purs
```diff
@@ -162,8 +189,69 @@ startOgmios cfg params = liftEffect $ spawn "ogmios" ogmiosArgs
...
+    defaultSpawnOptions
+  void $ recovering defaultRetryPolicy ([ \_ _ -> pure true ])
+    $ const
+    $ liftEffect
+    $ execSync
+        ( "psql -h " <> pgConfig.host <> " -p " <> UInt.toString pgConfig.port
+            <> " -d postgres"
+        )
+        defaultExecSyncOptions
+  liftEffect $ log =<< Bf.toString Encoding.UTF8 =<< execSync
```
klntsky, line 234: Let's use async versions of NodeJS API calls where possible. We should not block the interpreter if we can, since useful work can happen in background (query dispatching, etc)

src/Plutip/Server.purs
marijanp: I don't use it to stop the cluster. I use it to make sure that the server is listening.

src/Plutip/Server.purs
marijanp: I understand your concerns, but in this case we really need to execute the postgres related commands synchronized. I dont want the creation of a user to occur before `postgres-server` is executed. And that is the case for all the calls of execSync.

PR #642, opened 2022-06-30
Vasil Ogmios parsing testing and fixes

jy14898:
This PR introduces testing Ogmios response Aeson parsing for all queries, using our own fixtures and also fixtures pulled from Ogmios's own repo via nix.

To be compliant with the Ogmios fixtures, I made the following changes:
 - `CostModel` is now an `Array Int` (from `Array UInt`)
 - ~`EraSummaryParameters`'s `safeZone` is optional (`Maybe`)~
   - On `Nothing`/`null`, a default value of zero is given
 - ~The `plutus:v1` `CostModel` is optional (some fixtures only mention `plutus:v2`, which will be useful to test against)~
   - These fixtures are ignored for now 

Note: No verification of the fixtures fitting to the schemas are made (via Medea for eg), the assumption being that fixtures are pulled from the actual Ogmios service or by their official fixtures

~Once merged we can create a new issue for adding Ogmios's response fixtures for tx submission and evaluation~ This has been included in the PR


test/Plutus/Time.purs
```diff
@@ -93,8 +93,8 @@ mkEpochLength = EpochLength <<< BigInt.fromInt
 mkSlotLength :: Int -> SlotLength
 mkSlotLength = SlotLength <<< BigInt.fromInt
 
-mkSafeZone :: Int -> SafeZone
-mkSafeZone = SafeZone <<< BigInt.fromInt
+mkSafeZone :: Int -> Maybe SafeZone
+mkSafeZone = Just <<< SafeZone <<< BigInt.fromInt
```
klntsky, line 97: Why `Maybe`?

klntsky (dismissed):

test/Plutus/Time.purs
jy14898: Ogmios says safeZone can be null/nothing. To accept fixtures which used this I had to make it Maybe. Idk if there's a difference between 0 and Nothing

flake.nix
```diff
@@ -27,6 +27,13 @@
       flake = false;
     };
 
+
+    # for testing the ogmios parser
+    ogmios-fixtures = {
```
ngua, line 32: This is a redundant input. We can just use `ogmios`

flake.nix
```diff
@@ -27,6 +27,13 @@
...
+      url = "github:CardanoSolutions/ogmios?ref=master";
```
ngua, line 33: We also want the fixtures to match the version of Ogmios we're using, no? Our fork also has these fixtures

flake.nix
```diff
@@ -368,6 +375,7 @@
           project = pkgs.purescriptProject {
             inherit src pkgs;
             projectName = "cardano-transaction-lib";
+            ogmios = inputs.ogmios-fixtures;
```
ngua, line 378: Please add `ogmios-fixtures` to the CTL overlay instead, i.e.
```
      overlay = system: with inputs; (prev: final: {
        # ...
        ogmios = ogmios.packages.${system}."ogmios:exe:ogmios";
        ogmios-fixtures = ogmios;
        # ...
      });
```

nix/default.nix
```diff
@@ -1,6 +1,7 @@
 { pkgs, system }:
 { src
 , projectName
+, ogmios
```
ngua, line 4: See my other comment about this

nix/default.nix
```diff
@@ -63,10 +64,12 @@ let
         shellHook =
           let
             nodeModules = mkNodeModules { };
+            ogmiosFixtures = buildOgmiosFixtures { };
```
ngua, line 67: See my other comments about this

nix/default.nix
```diff
@@ -63,10 +64,12 @@ let
...
           in
           ''
             export NODE_PATH="${nodeModules}/lib/node_modules"
             export PATH="${nodeModules}/bin:$PATH"
+            export OGMIOS_FIXTURES="${ogmiosFixtures}"
```
ngua, line 72: Please remove this and add it to CTL's own `devShell`. Currently, _all_ library users will need to build the Ogmios fixtures even though they probably have no need for it (and even if they do, you're making it available with `buildOgmiosFixtures` below)

You can override the `shellHook` of our `devShell` to do this (the `shellHook` argument won't work since we need the `purescriptProject` to get the `buildOgmiosFixtures`), i.e.

```
          devShell = project.devShell.overrideAttrs (oas: {
            shellHook = oas.shellHook +
              ''
                export OGMIOS_FIXTURES="${buildOgmiosFixtures {}}"
              '';
          });
```

(The `devShell` is currently on line 433 of the flake)

Alternately, you could move the definition of `buildOgmiosFixtures` to outside this module, as it's not a project-wide tool that library users will likely find useful and doesn't depend on anything else in this module (except `projectName`, but that's not truly necessary)

nix/default.nix
```diff
@@ -139,6 +142,7 @@ let
         # sandbox, so we can just use node instead
         # (idea taken from `plutus-playground-client`)
         checkPhase = ''
+          export OGMIOS_FIXTURES="${buildOgmiosFixtures { }}"
```
ngua, line 145: Again, this should be removed as it requires users to build the fixtures for every test derivation created with this library. This is only useful for our own unit test, so we need to override the `checkPhase` instead of that single derivation

Starting at line 427 of the flake, you can change `ctl-unit-test` to:

```
            ctl-unit-test = (project.runPursTest {
              testMain = "Test.Unit";
              sources = [ "src" "test" "fixtures" ];
            }).overrideAttrs(oas: {
              checkPhase = ''       
                export OGMIOS_FIXTURES="${project.buildOgmiosFixtures { }}"
              '' + oas.checkPhase;
            });
```

nix/default.nix
```diff
@@ -251,10 +255,43 @@ let
       '';
     };
 
+  buildOgmiosFixtures =
```
ngua, line 258: As I mentioned in another comment, this can just be moved to the flake or another module since it's only useful for CTL itself most likely

flake.nix
```diff
@@ -400,6 +408,8 @@
             };
 
             docs = project.buildSearchablePursDocs;
+
+            ogmios-fixtures = project.buildOgmiosFixtures { };
```
ngua, line 412: Do we need this here? `buildOgmiosFixtures` is being created in several different places

ngua (changes requested):
I haven't had a chance yet to look at the Purescript changes very closely, but the Nix part requires several changes

I'm concerned about the changes to the cost models in the `ProtocolParameters`. If we don't have a `PlutusV1` entry, CTL basically won't work (for most use cases at least) -- we don't support `PlutusV2` quite yet. So instead of changing that now, why don't we ensure that any fixture pulled from Ogmios contains that entry (with `jq`)?

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,145 @@
+module Test.GenerateOgmiosFixtures where
```
vvtran, line 1: explicit exports.

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,145 @@
...
+
+import Prelude
+
+import Aeson (class DecodeAeson, class EncodeAeson, Aeson, stringifyAeson)
+import Contract.Monad (ListenerSet)
+import Control.Parallel (parTraverse)
+import Data.Either (Either(..))
```
vvtran, line 8: I won't list all of them but explicit import for constructors.

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,145 @@
...
+import QueryM.ServerConfig (ServerConfig, mkWsUrl)
+import Type.Prelude (Proxy(..))
+import Types.MultiMap as MultiMap
+import Data.Map as Map
+
+-- A simple websocket for testing
+-- TODO Generalize websocket constructors using type classes traversing rows
+-- to factor mk{Ogmios{DatumCache,},}WebSocket into a single implementation
+mkWebSocket
+  :: forall a b
```
vvtran, line 47: `:: forall (a :: Type) (b :: Type)`

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,145 @@
...
+  ws <- _mkWebSocket (logger Debug) $ mkWsUrl serverCfg
+  let
+    sendRequest = _wsSend ws (logString lvl Debug)
+    onError = do
+      logString lvl Debug "WS error occured, resending requests"
+      Ref.read pendingRequests >>= traverse_ sendRequest
+  _onWsConnect ws do
+    _wsWatch ws (logger Debug) onError
+    _onWsMessage ws (logger Debug) $ defaultMessageListener lvl md
+    _ <- _onWsError ws (logger Error) $ const onError
```
vvtran, line 68: Could use `void`

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,145 @@
...
+    cb $ Right $ WebSocket ws
+      (mkListenerSet dispatchMap pendingRequests)
+  pure $ \err -> cb $ Left $ err
+  where
+  logger :: LogLevel -> String -> Effect Unit
+  logger = logString lvl
+
+mkWebSocketAff
+  :: forall a b
```
vvtran, line 77: `:: forall (a :: Type) (b :: Type)`

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,145 @@
...
+   . DecodeAeson b
+  => Show b
+  => LogLevel
+  -> ServerConfig
+  -> Aff (WebSocket (ListenerSet a b))
+mkWebSocketAff lvl = makeAff <<< map (map (Canceler <<< (map liftEffect))) <<<
```
vvtran, line 83: Optional ofc but redundant parenths. `map (map (Canceler <<< map liftEffect))`

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,145 @@
...
+  mkWebSocket lvl
+
+foreign import md5 :: String -> String
+
+data Query = Query (JsonWspCall Unit Aeson) String
+
+mkQuery :: forall query. EncodeAeson query => query -> String -> Query
```
vvtran, line 90: `query :: Type`

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,145 @@
...
+        resp' = stringifyAeson resp
+        respMd5 = md5 resp'
+        fp = concat
+          [ "fixtures"
+          , "test"
+          , "ogmios"
+          , query <> "-" <> respMd5 <> ".json"
+          ]
+      writeTextFile UTF8 fp resp'
+    -- TODO Remove by properly shutting down the websocket
```
vvtran, line 144: Can we create an issue for this?

test/OgmiosAeson.purs
```diff
@@ -0,0 +1,103 @@
+module Test.OgmiosAeson where
```
vvtran, line 1: explict exports.

test/OgmiosAeson.purs
```diff
@@ -0,0 +1,103 @@
...
+            _ -> Nothing
+
+  for_ groupedFiles \(query /\ fps) ->
+    (if query `elem` supported then identity else skip)
+      $ test (query <> " (" <> show (length fps) <> ")")
+      $
+        for_ fps \fp -> do
+          file <- readTextFile UTF8 fp
+          let
+            handle :: forall f a. DecodeAeson a => f a -> Aff Unit
```
vvtran, line 88:
```
:: forall (f :: Type -> Type) (a :: Type)
 . DecodeAeson a
=> f a
-> Aff 
```

vvtran (commented):
Some comments.

test/Plutus/Time.purs
klntsky: I mean why `Maybe` in this function, it can be just `Int -> SafeZone` as no checks are present

flake.nix
jy14898: I kept it separate in case we did want to update our fixtures without having to take new ogmios changes. I also wasn't sure if the flake input ogmios can be treated as a directory, whereas this input is explicitly not a flake. That being said if we are going to combine them, we can move this work into our ogmios fork

flake.nix
jy14898: Mostly left over from my development of this feature, it was so I could look at the output of my fixture generating nix on its own. Happy to remove

jy14898:
@ngua WRT the protocol params and plutus:v1, I don't see it being much of a difference as to whether we fail at parsing vs some later stage, for eg when looking up PlutusV1 in ProtocolParams costModels?

I'm happy to drop those fixtures, either by checking with jq in nix or by my blacklist in purescript, but we will hit this same issue later

test/Plutus/Time.purs
jy14898: It's so it's convenient to build `EraSummaryParameters`, as that now has its `safeZone` field as a `Maybe`

jy14898:
output

vvtran (commented):
Not sure if you're getting this but one of the tests are failing for me
```
Ogmios Aeson tests
  ~ EvaluateTx (97)
  ✗ SubmitTx (14):

  SubmitTx-674441960ca1ba2de08ad4e50c9fde98.json
  An error occurred while decoding a JSON value:
  Expected value of type 'Expected Object'.

```

jy14898:
@vvtran currently waiting for the flakes dependencies to build (not sure if they're on our cachix yet I guess? I did change to the new key today)

test/GenerateOgmiosFixtures.purs
```diff
@@ -0,0 +1,148 @@
+module Test.GenerateOgmiosFixtures
```
ngua, line 1: I think this should be `Test.Ogmios.GenerateFixtures` and this file and its FFI module should be moved accordingly

test/OgmiosAeson.purs
```diff
@@ -0,0 +1,140 @@
+module Test.OgmiosAeson
```
ngua, line 1: `Test.Ogmios.Aeson`

test/OgmiosAeson.purs
```diff
@@ -0,0 +1,140 @@
...
+  ( main
+  , suite
+  ) where
+
+import Prelude
+
+import Aeson (class DecodeAeson, Aeson)
+import Aeson as A
```
ngua, line 9: Please import this as `Aeson` and update the use sites accordingly

ngua (changes requested):

test/Plutus/Time.purs
klntsky: We should definitely wrap it into maybe at the caller side.

jy14898:
Thanks, I'll spend some time investigating this oddity this evening

ngua:
@vvtran I think we should delete the `vasil` branch and start a new one for the features. This can be merged into `develop` though

flake.nix
```diff
@@ -413,10 +449,15 @@
...
-            };
+            }.overrideAttrs
+              (oas: {
+                checkPhase = exportOgmiosFixtures + oas.checkPhase;
+              });
           };
 
-          devShell = project.devShell;
+          devShell = project.devShell.overrideAttrs (oas: {
+            shellHook = oas.shellHook + exportOgmiosFixtures;
```
ngua, line 459: Since you moved `buildOgmiosFixtures` from `nix/default.nix` you don't need to do an override anymore. This can be passed directly to the `shell` argument of `purescriptProject`, so starting at line 411, it should be 

```
            shell = {
              shellHook = exportOgmiosFixtures;
              packages = [
               # ...
              ];
            };
``` 
and this can be reverted to 
```
           devShell = project.devShell;
```

ngua (commented):
Could you make this PR against `develop` now? We can remove the `vasil` branch instead

PR #646, opened 2022-06-30
Change `Show Address` add/change address to_bech32 functions

### LINKED ISSUE #623: Print Address as bech32 with correct prefix
We use bech32 without prefix in `Show Address` instance (`Serialization.Address`),  but explorers accept either addresses with `addr`  bech32 prefix or with `addr_test` prefix.

So we can't print addresses meaningfully without knowing a network id. We need a separate utility to print addresses (not `Show` instance). And we should not print bech32 in Show instance at all, because without a prefix it is useless. We need to print CborHex encoding in `Show`.

adamczykm:
Closes #623  .

### Pre-review checklist

### DIFF
--- a/src/Serialization/Address.js
+++ b/src/Serialization/Address.js
@@ -47,7 +47,14 @@ exports.addressBytes = callToBytes;
 exports.byronAddressBytes = callToBytes;
 exports.stakeCredentialToBytes = callToBytes;
 
-exports.addressBech32 = callToBech32;
+exports.addressBech32 = addr => {
+    // 0 TestnetId
+    // 1 MainnetId
+    return addr.network_id() == 0 ? addr.to_bech32("addr_test") : addr.to_bech32("addr");
+};
+exports.addressBech32_ = addr => prefix => {
+    return addr.to_bech32(prefix);
+};
 exports._addressNetworkId = toAdt => addr => {
     return toAdt(callNetworkId(addr));
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 
 ### Added
 
+- `addressBech32_` for printing bech32 repr of an `Address` with a given prefix.
 - `mkKeyWalletFromFile` helper to use `cardano-cli`-style `skey`s.
 - Single `Plutus.Conversion` module exposing all `(Type <-> Plutus Type)` conversion functions.
 - Support for using a `PrivateKey` as a `Wallet`.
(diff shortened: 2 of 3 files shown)


src/Serialization/Address.js
```diff
@@ -47,7 +47,14 @@ exports.addressBytes = callToBytes;
 exports.byronAddressBytes = callToBytes;
 exports.stakeCredentialToBytes = callToBytes;
 
-exports.addressBech32 = callToBech32;
+exports.addressBech32 = addr => {
+    // 0 TestnetId
+    // 1 MainnetId
+    return addr.network_id() == 0 ? addr.to_bech32("addr_test") : addr.to_bech32("addr");
```
klntsky, line 53: Wait I was wrong, we actually know the network id. Then we can print the address in a format acceptable by explorers. Could you please use this function?

klntsky (changes requested):

src/Serialization/Address.js
adamczykm: Im having troubles running examples and tests. Meanwhile discovered that `to_bech32` called parameterless should prepend the right prefix:
```
    pub fn to_bech32(&self, prefix: Option<String>) -> Result<String, JsError> {
        let final_prefix = match prefix {
            Some(prefix) => prefix,
            None => {
                // see CIP5 for bech32 prefix rules
                let prefix_header = match &self.0 {
                    AddrType::Reward(_) => "stake",
                    _ => "addr",
                };
                let prefix_tail = match self.network_id()? {
                    id if id == NetworkInfo::testnet().network_id() => "_test",
                    _ => "",
                };
                format!("{}{}", prefix_header, prefix_tail)
            }
        };
```

adamczykm: Will try to verify it. But that may end up waiting for tomorrow. Not sure what is your request exactly.

klntsky: It looks like the entire issue was created because of my misunderstanding (I wasn't aware that it should be OK).

It all started from this message: https://discord.com/channels/922576618424250418/940666072208662628/991274629983391786

klntsky: Could you please check that common testnet explorer indeed parse the addresses in this format, just to be sure?

adamczykm: > Im having troubles running examples and tests.
It also takes place on `develop` branch.

vvtran (dismissed):

klntsky:
> Oops, sorry, please ignore the previous approve. I think if we can print addresses knowing the network id implicitly, could that be used for Show instead of the hexstring?

Yes

ngua:
Mostly superseded by #849

klntsky:
Closing, we don't need this changes. Slightly related: https://github.com/Plutonomicon/cardano-transaction-lib/pull/849/files

PR #649, opened 2022-06-30
E2E Part 1

firefrorefiddle:
Work towards https://github.com/Plutonomicon/cardano-transaction-lib/issues/465

Can E2E-test simple examples when make run-dev is running and chrome is available

### Pre-review checklist


ngua:
I think there's a fair amount left to do on this, so I'm going to convert it to a draft for now

ngua:
@firefrorefiddle You need to regenerate `spago-packages.nix`, that's why CI is failing. `nix develop` then `spago2nix generate` (and add and commit the changed file)

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+import Effect.Aff (Aff, launchAff_, delay)
+import Effect.Class (liftEffect)
+import Effect.Console (log)
+import Effect.Ref (Ref)
+import Effect.Ref as Ref
+import Effect.Uncurried (mkEffectFn1, EffectFn1)
+import Foreign (Foreign, unsafeFromForeign)
+import Test.E2E.Feedback (testFeedbackIsTrue)
+import Test.Spec.Assertions (shouldSatisfy)
+import Toppokki as Toki
```
ngua, line 53: I don't think the few characters saved are worth it -- I'd recommend `as Toppokki` everywhere it's imported

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+
+geroInit :: Toki.Browser -> Aff Unit
+geroInit browser = do
+  page <- Toki.newPage browser
+  jQuery <- retrieveJQuery page
+  Toki.goto
+    (wrap "chrome-extension://iifeegfcfhlhhnilhfoeihllenamcfgc/index.html")
+    page
+  void $ Toki.unsafeEvaluateStringFunction jQuery page
+  delay $ wrap 1000.0
```
ngua, line 201: `delay $ wrap` is repeated a lot, why not factor that out?

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+
+{-geroSetCollateral :: Toki.Browser -> Aff Unit
+geroSetCollateral = do
+  page <- Toki.newPage browser
+  jQuery <- retrieveJQuery page
+  Toki.goto (wrap "chrome-extension://iifeegfcfhlhhnilhfoeihllenamcfgc/index.html") page
+-}
+
+geroSign :: RunningExample -> Aff Unit
+geroSign (RunningExample { browser }) = do
```
ngua, line 231: Redundant `do` here

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+  pure unit
+
+-- | Wrapper for Page so it can be used in `shouldSatisfy`, which needs 'Show'
+-- | Doesn't show anything, thus 'NoShow'
+newtype NoShowPage = NoShowPage Toki.Page
```
ngua, line 236: This comment is a bit confusing to me. By "doesn't show anything", do you mean that `Page` doesn't have a `Show` instance? Should the name be something like `ShowablePage` since it provides a `Show` instance for an otherwise un`show`able type?

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+derive instance Newtype NoShowPage _
+
+instance Show NoShowPage where
+  show _ = "Toppoki.Page"
+
+newtype Selector = Selector String
+
+derive instance Newtype Selector _
+
+newtype Action = Action String
```
ngua, line 247: This could use a doc comment

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+        Toki.unsafeEvaluateStringFunction "typeof(jQuery) !== 'undefined'" page
+    unless alreadyInjected $ void $ Toki.unsafeEvaluateStringFunction jQuery
+      page
+  pure pages
+
+reactSetValue :: Selector -> String -> Toki.Page -> Aff Unit
+reactSetValue selector value page = void
+  $ flip Toki.unsafeEvaluateStringFunction page
+  $ fold
+      [ -- https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js
```
ngua, line 328: This should be moved to a top-level doc comment I think, it's a pretty important explanation

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+  :: Selector -> String -> Toki.Page -> Effect (Promise Unit)
+
+foreign import _clickTab :: Toki.Browser -> Toki.Page -> Effect Unit
+foreign import _openPopup :: Toki.Browser -> Effect (Promise Unit)
+
+clickTab :: Toki.Browser -> Toki.Page -> Aff Unit
+clickTab b p = liftEffect $ _clickTab b p
+
+openPopup :: Toki.Browser -> Aff Unit
+openPopup b = toAffE $ _openPopup b
```
ngua, line 365:
```suggestion
openPopup = toAffE <<< _openPopup
```

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+
+foreign import _retrieveJQuery :: Toki.Page -> Effect (Promise String)
+foreign import _typeInto
+  :: Selector -> String -> Toki.Page -> Effect (Promise Unit)
+
+foreign import _clickTab :: Toki.Browser -> Toki.Page -> Effect Unit
+foreign import _openPopup :: Toki.Browser -> Effect (Promise Unit)
+
+clickTab :: Toki.Browser -> Toki.Page -> Aff Unit
+clickTab b p = liftEffect $ _clickTab b p
```
ngua, line 362:
```suggestion
clickTab b = liftEffect <<< _clickTab b
```

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+      , "nativeInputValueSetter.call(input, '" <> value <> "');"
+      , "var ev2 = new Event('input', { bubbles: true});"
+      , "input.dispatchEvent(ev2);"
+      ]
+
+reactSetValueP :: Selector -> String -> Toki.Page -> Aff Unit
+reactSetValueP selector value page = void
+  $ flip Toki.unsafeEvaluateStringFunction page
+  $ fold
+      [ -- https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js
```
ngua, line 343: This one too. You could have a "note" section somewhere and refer to it since it applies to two different top-level bindings, e.g.

```
-- Note [Nami & ReactJS]
-- ...
```

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+      page
+  pure pages
+
+reactSetValue :: Selector -> String -> Toki.Page -> Aff Unit
+reactSetValue selector value page = void
+  $ flip Toki.unsafeEvaluateStringFunction page
+  $ fold
+      [ -- https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js
+        -- Nami uses react, which complicates things a bit
+        "var input = $('" <> unwrap selector <> "').get(0);"
```
ngua, line 330: Why `var`s? We should use `let`/`const` unless there's an extremely compelling reason not to

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+  void $ doJQ (byId "user-agree") click page
+  void $ doJQ (byId "password-agree") click page
+  void $ doJQ (byId "set-up-password") click page
+  delay $ wrap 100.0
+  void $ doJQ (buttonWithText "Continue") click page
+  delay $ wrap 100.0
+  void $ doJQ (byId "done") click page
+  openPopup browser
+  delay $ wrap 50000.0
+  pure unit
```
ngua, line 221: This line is redundant

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+    delay $ wrap 100.0
+    void $ doJQ (inputType "checkbox") click page
+    void $ doJQ (buttonWithText "Connect") click page
+
+geroInit :: Toki.Browser -> Aff Unit
+geroInit browser = do
+  page <- Toki.newPage browser
+  jQuery <- retrieveJQuery page
+  Toki.goto
+    (wrap "chrome-extension://iifeegfcfhlhhnilhfoeihllenamcfgc/index.html")
```
ngua, line 198: I think the URL should be made into a constant or something. It's kind of hidden here and it might not be immediately clear what it's pointing to

test/E2E/Helpers.js
```diff
@@ -0,0 +1,34 @@
+exports._retrieveJQuery = page => () =>
```
ngua, line 1: The formatting throughout this module is difficult to read

ngua: We're going to add JS formatting/linting to CI, I'll do that soon and you can merge it in

test/E2E/Browser.purs
```diff
@@ -0,0 +1,100 @@
+module Test.E2E.Browser
```
ngua, line 1: I think all of the `Test.E2E.*` modules should be moved to a different part of the module hierarchy. Many (most?) of our users will want to use e2e testing, but the current module naming/location suggests its for CTL's internal tests

Perhaps we can have a `Contract.Test.*` namespace, with `E2E` modules therein

test/E2E/Browser.purs
```diff
@@ -0,0 +1,100 @@
...
+      [ "--disable-extensions-except=" <> extDir
+      , "--load-extension=" <> extDir
+      ] <> if mode == Headless then [ "--headless=chrome" ] else []
+  , headless: mode == Headless
+  , userDataDir: chromeUserDataDir
+  , executablePath: fromMaybe "" chromeExe
+  }
+  where
+  mode :: Mode
+  mode = if noHeadless then Visible else Headless
```
ngua, line 97:
```suggestion
  mode noHeadless
    | noHeadless = Visible 
    | otherwise = Headless
```

test/E2E/Browser.purs
```diff
@@ -0,0 +1,100 @@
...
+  , str
+  , strOption
+  , switch
+  , value
+  , info
+  , fullDesc
+  )
+import Toppokki as Toki
+
+data TestOptions = TestOptions
```
ngua, line 34:
```suggestion
newtype TestOptions = TestOptions
```

ngua: You should derive instances for this as well (`Eq`, `Newtype`, `Generic`) and add `Show` with `genericShow`

test/E2E/Browser.purs
```diff
@@ -0,0 +1,100 @@
...
+  , headless: mode == Headless
+  , userDataDir: chromeUserDataDir
+  , executablePath: fromMaybe "" chromeExe
+  }
+  where
+  mode :: Mode
+  mode = if noHeadless then Visible else Headless
+
+  extDir :: String
+  extDir = if wallet == "Gero" then geroDir else namiDir
```
ngua, line 100: If we were going to keep using a `String` argument, I'd recommend:

```suggestion
  extDir wallet
    | wallet == "Gero" = geroDir 
    | otherwise = namiDir
```

But see my other comment about that

test/E2E/Feedback.purs
```diff
@@ -0,0 +1,31 @@
...
+  ( publishTestFeedback
+  , retrieveTestFeedback
+  , testFeedbackIsTrue
+  ) where
+
+import Effect (Effect)
+import Effect.Aff (Aff)
+import Effect.Class (liftEffect)
+import Foreign (Foreign, unsafeFromForeign)
+import Prelude
```
ngua, line 14: Pleas move this to the first line of the import list, and add a space between it and the other imports

test/E2E/Browser.purs
```diff
@@ -0,0 +1,100 @@
+module Test.E2E.Browser
+  ( Mode(..)
+  , TestOptions(..)
+  , launchWithExtension
+  , parseOptions
+  ) where
+
+import Prelude
+
+import Control.Applicative (apply, (<$>))
```
ngua, line 10: These imports are redundant (they're exported from `Prelude`)

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,365 @@
...
+  findWalletPage jQuery browser >>= case _ of
+    Nothing -> do
+      delay $ wrap 100.0
+      waitForWalletPage jQuery browser
+    Just page -> pure page
+
+showOutput :: Ref (Array E2EOutput) -> Effect String
+showOutput ref =
+  let
+    show' (E2EOutput { outputType, output }) = showType outputType <> " " <>
```
ngua, line 116: I think it might be easier to read if `show'` and `showType` where moved to a `where` clause with type signatures

test/E2E/Browser.purs
```diff
@@ -0,0 +1,100 @@
...
+    ]
+  in
+    TestOptions
+      { chromeExe, namiDir, geroDir, chromeUserDataDir, noHeadless }
+
+parseOptions :: Effect TestOptions
+parseOptions = execParser $ info optParser fullDesc
+
+launchWithExtension
+  :: TestOptions -> String -> Aff Toki.Browser
```
ngua, line 83: I'd recommend not using a `String` for the `wallet` argument, since there are only two possible choices (see my other comment)

test/E2E/Browser.purs
```diff
@@ -0,0 +1,100 @@
...
+  , switch
+  , value
+  , info
+  , fullDesc
+  )
+import Toppokki as Toki
+
+data TestOptions = TestOptions
+  { chromeExe :: Maybe String
+  , namiDir :: String
```
ngua, line 36: As I also mentioned in a comment on `launchWithExtensions`, we should have a sum type representing the wallet choice

To avoid name clashes with our own `Wallet` type, you could have something like

```purescript
data WalletExt 
  = NamiExt Path
  | GeroExt Path
```

Where `Path` is some newtype around `String`s, representing the paths now held in `namiDir` and `geroDir`

This way both the extension to run and its associated directory can be held by `TestOptions` without any redundancies or non-exhaustiveness

ngua (dismissed):
I'm not sure why I didn't think of this earlier, but we can use Nix to package the extensions with chromium and put it in the CTL overlay. This has several advantages:
- we don't need to commit the extension binaries 
- we can keep things updated in one place
- it will be easier for users
I can help you with this part

Some other comments as well

test/E2E/Browser.purs
ngua: Also `TestOptions` in general needs more doc comments

ngua:
Also, we should add a variant of `runPursCheck` that can work with the e2e test

test/E2E/Helpers.purs
firefrorefiddle: It's rather the difficulty of spelling it with 2 'p's and 2 'k's. If for one haven't even heard of the toppokki (the food) before and wouldn't know how to spell it. ;-)

test/E2E/Helpers.purs
firefrorefiddle: I was thinking of that, too.

test/E2E/Helpers.purs
firefrorefiddle: It's  a Show instance with nothing to show. But I think its obsolete anyway, I will delete it.

test/E2E/Helpers.purs
firefrorefiddle: Function doesn't do anything yet. I think I should add a Pkh2Pkh with Gero, so we have something to sign with Gero.

klntsky:
Could you please provide a step-by-step guide and handle the case when binary is `chromium`. I think just putting `google-chrome` in a variable to the definitions section of the Makefile would be enough. The goal is to make using it with chromium as simple as changing one line in the Makefile.

test/E2E/Browser.purs
```diff
@@ -0,0 +1,100 @@
...
+  , geroDir :: String
+  , chromeUserDataDir :: String
+  , noHeadless :: Boolean
+  }
+
+data Mode = Headless | Visible
+
+derive instance Eq Mode
+
+optParser :: Parser TestOptions
```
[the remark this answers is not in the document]
firefrorefiddle: You mean this?  `  { chromeExe :: Maybe FilePath`

test/Examples/SignMultiple.purs
```diff
@@ -0,0 +1,26 @@
...
+  , delaySec
+  , runE2ETest
+  )
+import TestM (TestPlanM)
+
+runExample :: TestOptions -> TestPlanM Unit
+runExample options = runE2ETest "SignMultiple" options "Nami" $ \example -> do
+  namiConfirmAccess example
+  namiSign example
+  -- Wait a moment to avoid a race condition. After Nami gets confirmation,
```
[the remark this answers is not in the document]
firefrorefiddle: Maybe. There are certain things you can wait for in puppeteer, like 'navigation', 'domcontentloaded' etc., but they don't seem to be very reliable. Here https://stackoverflow.com/a/61304202 its proposed to wait until the DOM stops changing.

Makefile
```diff
@@ -6,10 +6,27 @@ SHELL := bash
 ps-sources := $(shell fd -epurs)
 nix-sources := $(shell fd -enix --exclude='spago*')
 hs-sources := $(shell fd . './server/src' './server/exe' -ehs)
-ps-entrypoint := Examples.AlwaysSucceeds
+ps-entrypoint := Examples.ByUrl
 ps-bundle = spago bundle-module -m ${ps-entrypoint} --to output.js
-
 node-ipc = $(shell docker volume inspect cardano-transaction-lib_node-ipc | jq -r '.[0].Mountpoint')
+e2e-temp-dir := $(shell mktemp -tdu e2e.XXXXXXX)
+# bump version here
```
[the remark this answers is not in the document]
firefrorefiddle: It does. No, wait. It did. Ah.

klntsky:
> Just replace google-chrome with chromium

This variable is not used in all places where it should

klntsky:
```
Invalid argument `/snap/bin/chromium'

Usage: E2ETest [--chrome-exe FILE] --nami-dir DIR --gero-dir DIR
               [--chrome-user-data DIR] [--no-headless]
```


Can be fixed by using `--chrome-exe $(call e2e-browser)"` instead of `--chrome-exe=$(call e2e-browser)"`

firefrorefiddle:
> using `./tmp` dir does not play well with `make format` - js formatter attempts to process this dir (I merged this branch with develop locally)

Yes, it's a silly place. Do you have a good idea?

src/Contract/Test/E2E.purs
```diff
@@ -0,0 +1,40 @@
...
+import Contract.Test.Examples.AlwaysMints as AlwaysMints
+import Contract.Test.Examples.AlwaysSucceeds as AlwaysSucceeds
+import Contract.Test.Examples.Datums as Datums
+import Contract.Test.Examples.MintsMultipleTokens as MintsMultipleTokens
+import Contract.Test.Examples.SignMultiple as SignMultiple
+import Test.Spec.Runner as SpecRunner
+import Test.Utils as Utils
+import TestM (TestPlanM)
+
+-- Run with `spago test --main Test.E2E`
```
klntsky, line 22:
```suggestion
-- Run with `spago test --main Contract.Test.E2E`
```

klntsky (changes requested):

klntsky:
I noticed that when I run the tests, `SignMultiple` fails with ValueNotConserved, but if I comment-out all the tests except SignMultiple, it passes. I believe the reason is that UTxO spending does not propagate fast enough to the wallet.

klntsky:
I am unable to run Gero tests:

```

Start Example Gero
Example Gero finished, check success...
... app starting
... Failed to load resource: the server responded with a status of 404 (File not found)
e2e tests
  ✗ Gero:

  false doesn't satisfy predicate
```

klntsky:
Please move `src/Contract/Test/Examples` to `test/`

`Contract.Test.E2E` should be under `test` too. And it should be renamed to something without `Contract`. `Contract.*` is our public API namespace we provide for library users.

klntsky:
I am quite sure we also don't want to export that much from `Contract.Test.Helpers`. 
`exampleUrl`, `testPassword`, `testPasswordGero` - I don't think these are needed for the users.
`Contract.Test.Helpers` should also be renamed (the name should suggest that this module is only useful for E2E tests).

firefrorefiddle:
> Please move `src/Contract/Test/Examples` to `test/`
> 
> `Contract.Test.E2E` should be under `test` too. And it should be renamed to something without `Contract`. `Contract.*` is our public API namespace we provide for library users.

@klntsky I had them there, but moved them according to @ngua 's proposal from two weeks ago (which you upvoted; further up in this conversation).

Quote:
> I think all of the Test.E2E.* modules should be moved to a different part of the module hierarchy. Many (most?) of our users will want to use e2e testing, but the current module naming/location suggests its for CTL's internal tests
>
> Perhaps we can have a Contract.Test.* namespace, with E2E modules therein

That's how I understood it.

klntsky:
> proposal from two weeks ago (which you upvoted; further up in this conversation).

I think we want to export all the machinery but not the examples themselves, as they are useless for the users - indeed that's not how the original message reads.

Also please move passwords and all functions that are only useful for us to the `test/` dir.

klntsky (changes requested):
Please move all the examples to `test/`, the main E2E module should go to `test/` too.

But `Contract.Test.Browser` should remain in `src/`.

Also why are helpers and Feedback are in `Internal` namespace? If they are not meant to be used by external users, we should remove them from `Contract` namespace.

firefrorefiddle:
@klntsky I restructured it as you requested. Is it more like you expected it now?

firefrorefiddle:
> I am unable to run Gero tests:
> 
> ```
> 
> Start Example Gero
> Example Gero finished, check success...
> ... app starting
> ... Failed to load resource: the server responded with a status of 404 (File not found)
> e2e tests
>   ✗ Gero:
> 
>   false doesn't satisfy predicate
> ```

Is that still the case for you? This one always works for me, even with a clean repo checkout.

firefrorefiddle:
> I noticed that when I run the tests, `SignMultiple` fails with ValueNotConserved, but if I comment-out all the tests except SignMultiple, it passes. I believe the reason is that UTxO spending does not propagate fast enough to the wallet.

Unfortunately, this error happens again and again with various examples. I am not sure if the solution is just to introduce rather long wait times between examples or if they perhaps point to real issue.

klntsky:
> or if they perhaps point to real issue.

I am quite sure they don't, given all the symptoms.

klntsky:
> Is that still the case for you?

I was seeing this again just now. But then I tried once more after removing `./tmp` and it worked.

test/E2E/Helpers.purs
```diff
@@ -0,0 +1,382 @@
+module Test.E2E.Helpers
```
klntsky, line 1: Sorry for not communicating it clear, I think we want helpers to be in `src`, but not the functions that are only usable for us, like our passwords.

klntsky, line 1: Module name should match directory hierarchy, regardless of where the module is located (`src` or `test`). PS does not enforce this, but we should.

src/Contract/Test/E2E.purs
```diff
@@ -0,0 +1,16 @@
+{- | This module is intended to be used for running custom E2E-tests -}
+module Contract.Test.Examples
+  ( module E2EHelpers
+  ) where
+
+import Test.E2E.Helpers
```
klntsky, line 6: We should never import anything from `test/` in `src/*` modules.

firefrorefiddle:
> > or if they perhaps point to real issue.
> 
> I am quite sure they don't, given all the symptoms.

Will we run into test errors if more than one person runs the tests simultaneously? Just wondering because at the moment I am getting loads of fails and before everything worked. Everyone is using the same wallet after all, so the chances for double spending seems pretty high to me.

Perhaps it was a mistake to have one wallet for everyone.

klntsky (changes requested):
Sorry but it's still not quite it.

Please split Helpers module in two, moving out functions that are specific to our use case I mentioned above. One module (with public-facing interface) should go to `src/Contract`, the other should go to `test/` (it's fine if it just contains URL & password fixtures)

klntsky:
> Will we run into test errors if more than one person runs the tests simultaneously?

This certainly may be the case.

klntsky:
> Perhaps it was a mistake to have one wallet for everyone.

No, that is fine at this stage.

https://github.com/Plutonomicon/cardano-transaction-lib/issues/783

test/E2E/Helpers.purs
firefrorefiddle: But it does, no?

klntsky: It should be `test/Test/E2E/Helpers.purs`, because the first directory is a source directory (see `spago.dhall`, so it does not count).

klntsky: I think it's better to adjust the module name and not the filepath

firefrorefiddle: Yeah, sorry for being dumb. ;-) I tried again.

klntsky: @firefrorefiddle you are right. Then leave it as is, I will fix it for the whole project.

klntsky (changes requested):
Sorry for bothering again, I think we should reorganize modules with Plutip in mind.

So everything E2E goes to `Contract.Test.E2E*` and Plutip testing will go to `Contract.Test.Plutip*`.

test/E2E/Helpers.purs
firefrorefiddle: @klntsky I can do it later if you like. But it will mean adjusting the filepath rather than the module name because the `Test.`-prefix is what prevents name clashes with the tested modules.

src/Contract/Test/Helpers.purs
```diff
@@ -0,0 +1,335 @@
+{- | This module is intended to be used for running custom E2E-tests -}
```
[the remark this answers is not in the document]
firefrorefiddle: > However, I am not sure about the CLI parser. How reusable it is? I think it does provide some value in case the users want to make their own binary that works exactly the same.

I feel the same about it. I tend to leave it exported, because it might be useful.

klntsky (approved):
Thank you for this enormous work!

Issue #530, opened 2022-06-10
Add a JS linter


klntsky:
Let's enforce uniform style in FFI JS code.

The Ask:

1. Choose a JS linter
2. Add it to npm dependencies and make sure to update the lockfile
3. Add a call to linter to `Makefile`: both to `format` and to `check-format`.
4. Add a similar check to `formatting-check` in `flake.nix`

PR #685, opened 2022-07-05
Add reference inputs (de)serialization + tests

### LINKED ISSUE #680: Vasil features: add support for reference inputs
Follow these changes in CSL: https://github.com/Emurgo/cardano-serialization-lib/pull/461

Modify our (de)serialization and domain types accordingly

klntsky:
Closes #680 

### Pre-review checklist

### DIFF
--- a/src/Cardano/Types/Transaction.purs
+++ b/src/Cardano/Types/Transaction.purs
@@ -198,6 +198,7 @@ newtype TxBody = TxBody
   , auxiliaryDataHash :: Maybe AuxiliaryDataHash
   , validityStartInterval :: Maybe Slot
   , mint :: Maybe Mint
+  , referenceInputs :: Maybe (Array TransactionInput)
   , scriptDataHash :: Maybe ScriptDataHash
   , collateral :: Maybe (Array TransactionInput)
   , requiredSigners :: Maybe (Array RequiredSigner)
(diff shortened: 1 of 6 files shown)


src/Deserialization/Transaction.js
```diff
@@ -59,6 +59,9 @@ exports._txBodyAuxiliaryDataHash = maybeGetter("auxiliary_data_hash");
 exports._txBodyValidityStartInterval = maybeGetter("validity_start_interval_bignum");
 // multiassets(): Mint | void;
 exports._txBodyMultiAssets = maybeGetter("multiassets");
+exports._txBodyReferenceInputs = maybe => containerhelper =>
+    body => body.reference_inputs() ?
```
ngua, line 63: I think this formatting is a bit hard to read, but that might just be me

ngua: Speaking of which we should probably add `prettier` or something and format all of the JS (along with a linter). Unrelated to this PR of course

ngua (approved):
Small comment about JS formatting but otherwise looks good

src/Deserialization/Transaction.js
klntsky: https://github.com/Plutonomicon/cardano-transaction-lib/issues/530

src/Cardano/Types/Transaction.purs
```diff
@@ -198,6 +198,7 @@ newtype TxBody = TxBody
   , auxiliaryDataHash :: Maybe AuxiliaryDataHash
   , validityStartInterval :: Maybe Slot
   , mint :: Maybe Mint
+  , referenceInputs :: Maybe (Array TransactionInput)
```
vvtran, line 201: Should we use `Set` instead? 
```
 /**
   * @param {TransactionInputs} reference_inputs
   */
  set_reference_inputs(reference_inputs: TransactionInputs): void;
```
There's probably an enforced ordering.

vvtran: I'm not actually certain on this admittedly.

klntsky: I'd prefer to preserve the ordering, it's safer to be closer to CTL than not. We don't have `Set` as API requirement, right?

PR #701, opened 2022-07-07
remove node_modules from .gitignore

### LINKED ISSUE #662: Remove `node_modules` from `.gitignore`
A seemingly common issue is developers of CTL use `npm` commands that install the `node_modules` folder. This is forbidden as the dependencies are actually managed by nix. Eg. `npm install PACKAGE` was used instead of `npm install --package-lock-only PACKAGE`. One way we can communicate this to users (other than in the docs) is by removing `node_modules` from the git ignore, making it clearer it's not an artifact of our build process.

Geometer1729:
Closes #662  .

### Pre-review checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,8 +12,6 @@ result
 result-*
 .envrc
 .direnv
-/node_modules
-/node_modules/
 .node
 .node-cfg
 /server/dist-newstyle


Geometer1729:
I don't think it makes sense to mention this in the change-log as it doesn't impact anyone importing ctl.

Geometer1729:
The idea is that if you generate `node_modules` it's clear from git status that this is not intended. Moving `package.json` into a sub directory breaks `npm run-dev` too.

ngua (approved):

PR #703, opened 2022-07-07
Improve explicit imports/exports check

### LINKED ISSUE #700: Explicit exports check improvement
https://github.com/Plutonomicon/cardano-transaction-lib/pull/693 introduced add explicit exports (and imports?) check, but it needs to be improved

Firstly, it doesn't work with nix-os (/bin/bash isn't found), and secondly, purs-tidy turns `Ty (..)` into `Ty(..)`, which isn't detected by the grep I think

klntsky:
Closes #700

(Strange I didn't notice it from the start)

### Pre-review checklist

- [ ] All Purescript imports are explicit, including constructors


Makefile
```diff
@@ -18,11 +18,11 @@ run-build:
 	@${ps-bundle} && BROWSER_RUNTIME=1 webpack --mode=production
 
 check-explicit-exports: $(eval SHELL:=/bin/bash)
```
jy14898, line 20: Is this actually necessary, given that we have `SHELL := bash` earlier?

ngua: No

vvtran (commented):
Out of curiosity, what's the motivation for explicit exports? Imports make sense as it can be unclear where an import comes from (unless you have a plugin, HLS or whatever) but exports should always be from the module you're looking at and pretty clear.

klntsky:
> Out of curiosity, what's the motivation for explicit exports?

I wanted to make this check automatic for a long time, but it'd be very time-consuming to add a grep command that respects only imports and not exports.

I think that all code style requirements should either be automated or not exist. This is the least effort thing I can do to move the situation closer to the ideal.

klntsky:
I agree with your point that it's obvious where the exports come from, but it still improves readability a bit (you can see everything that's exported without the need to scroll down a module).

jy14898:
In the future we could look into using https://github.com/natefaubion/purescript-language-cst-parser to re-write this check. I don't see it taking longer than a days work, and imo it's better than forking purs-tidy and having to keep up with their changes.

firefrorefiddle (approved):

Issue #645, opened 2022-06-30
Upgrade Cardano/Plutus libraries and endpoints


ngua:
Once #379 is complete, we need to fully support Babbage-era transactions. This requires adjusting the server endpoints after upgrading all of the libraries

klntsky:
@ngua should this be closed?

ngua:
Yes @klntsky

PR #707, opened 2022-07-08
Improve `KeyWallet` collateral selection and setting logic

errfrom:
An attempt to improve collateral selection logic for `KeyWallet`. The idea is to utilize the new Babbage collateral return feature, which will allow us to **1:** consider utxos with non-ADA assets as collateral input and **2:** not to spend the entire collateral value in case of script failure.

~~Currently it's not possible to test the functionality introduced in this PR, since we need to update the Haskell Server to support the new Babbage fields when estimating fees (blocked by https://github.com/Plutonomicon/cardano-transaction-lib/issues/645). @ngua~~ 

**Update:**

I implemented an algorithm that attempts to select the most optimal combination of utxos to use as collateral by constructing all possible combinations of utxos and then applying the following considerations: 

1. There is a protocol parameter for the maximum number of collateral inputs (`maxCollateralInputs`).
2. We would like collateral inputs to have a total of at least 5 Ada (`minRequiredCollateral`).
3. We prefer the utxo combinations that require the lowest utxo min ada value for the corresponding collateral output. Thus, we try to keep `totalCollateral` high, given that it is calculated as `sum_value(collateralInputs) - value(collateralOutput)`.
4. Also, if two utxo combinations correspond to return outputs with the same utxo min ada value, we prefer the one with fewer inputs. 

 https://github.com/Plutonomicon/cardano-transaction-lib/blob/0947f662e61d18ebd2213cedf9b15d75c2de394b/src/BalanceTx/Collateral.purs#L175-L192

### Pre-review checklist


src/BalanceTx/Collateral.purs
```diff
@@ -0,0 +1,221 @@
...
+data CollateralReturnError
+  = CollateralReturnError String
+  | CollateralReturnMinAdaValueCalcError
+
+derive instance Generic CollateralReturnError _
+
+instance Show CollateralReturnError where
+  show = genericShow
+
+addTxCollateralReturn
```
klntsky, line 81: Please add some comments to all non-trivial functions in this module

src/BalanceTx/Collateral.purs
```diff
@@ -0,0 +1,221 @@
...
+      -- If two candidate utxo combinations correspond to return outputs with
+      -- the same utxo min ada value, order them by the number of
+      -- collateral inputs:
+      EQ -> on compare (List.length <<< Tuple.fst <<< unwrap) lhs rhs
+      ordering -> ordering
+
+mkCollateralCandidate
+  :: List TransactionUnspentOutput /\ Maybe BigInt -> Maybe CollateralCandidate
+mkCollateralCandidate (_ /\ Nothing) = Nothing
+mkCollateralCandidate (unspentOutputs /\ Just val) =
```
klntsky, line 172: Use `$>`

src/BalanceTx/Collateral.purs
```diff
@@ -0,0 +1,221 @@
...
+adaValue' init = add init <<< adaValue
+
+nonAdaAsset :: TransactionUnspentOutput -> NonAdaAsset
+nonAdaAsset =
+  Value.getNonAdaAsset <<< _.amount <<< unwrap <<< _.output <<< unwrap
+
+subsequences :: forall (a :: Type). List a -> List (List a)
+subsequences Nil = Cons Nil Nil
+subsequences (Cons x xs) =
+  map (Cons x) (subsequences xs) <> subsequences xs
```
klntsky, line 216: Use `let subs = subsequences xs` to avoid calling it twice.

src/BalanceTx/BalanceTx.purs
```diff
@@ -435,6 +441,46 @@ _redeemersTxIns = lens' \(UnattachedUnbalancedTx rec@{ redeemersTxIns }) ->
...
+      <<< note (GetWalletCollateralError' CouldNotGetCollateral)
+selectCollateral (Just (KeyWallet _)) utxos = do
+  -- TODO: Combine with getWalletCollateral and supply with fee estimate
+  -- https://github.com/Plutonomicon/cardano-transaction-lib/issues/510
+  maxCollateralInputs <- Collateral.getMaxCollateralInputs
+  availableUtxos <- filterLockedUtxos utxos
+  Collateral.selectCollateral maxCollateralInputs availableUtxos <#>
+    map (Just <<< Array.fromFoldable)
+      <<< note (GetWalletCollateralError' CouldNotGetCollateral)
+selectCollateral _ _ = pure (Right Nothing)
```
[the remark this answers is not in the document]
errfrom: I opened a separate issue for this: https://github.com/Plutonomicon/cardano-transaction-lib/issues/786

src/BalanceTx/BalanceTx.purs
```diff
@@ -435,6 +441,46 @@ _redeemersTxIns = lens' \(UnattachedUnbalancedTx rec@{ redeemersTxIns }) ->
...
+  :: Transaction
+  -> Utxo
+  -> Address
+  -> QueryM (Either BalanceTxError Transaction)
+setCollateral transaction utxos ownAddr = runExceptT do
+  wallet <- asks _.wallet
+  mCollateral <- ExceptT $ selectCollateral wallet utxos
+  case mCollateral /\ wallet of
+    Nothing /\ _ ->
+      pure transaction
```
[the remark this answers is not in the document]
errfrom: I opened a separate issue for this: https://github.com/Plutonomicon/cardano-transaction-lib/issues/786

src/BalanceTx/Collateral.purs
```diff
@@ -0,0 +1,252 @@
...
+-- |   (`minRequiredCollateral`).
+-- |
+-- |   3. We prefer utxo combinations that require the lowest utxo min ada
+-- |   value for the corresponding collateral output, thus maintaining a
+-- |   sufficient `totalCollateral`.
+-- |
+-- |   4. If two utxo combinations correspond to return outputs with the same
+-- |   utxo min ada value, we prefer the one with fewer inputs.
+-- |
+selectCollateral
```
klntsky, line 203: My only concern here is time complexity. It's not uncommon to have hunderds of UTxOs. Could you please check the algorithm by faking the inputs (provide a list with increasing length). We need to know how many UTxOs can we handle before the performance becomes a problem (say before we get freeze time in range of seconds).

errfrom: @klntsky Yes, that's a valid point. My initial assumption was that we could sacrifice a bit of performance in favor of correctness. I'll add unit tests for the algorithm then, and think towards possible optimization options if performance becomes an issue.

src/BalanceTx/Collateral.purs
```diff
@@ -218,6 +224,7 @@ selectCollateral maxCollateralInputs =
     -- with the number of utxos <= `maxCollateralInputs`:
     <<< combinations maxCollateralInputs
     <<< map asTxUnspentOutput
+    <<< List.take maxCandidateUtxos
```
klntsky, line 227: I don't think it's acceptable. It's a regression even, because now it's possible that we don't include the only Ada UTxOs and fail because of this limit.

Instead, let's do this:

- use only utxos with ada (and possibly tokens), sorted by amount, in decreasing order 
- in case there's not a single one candidate after we cut 10 first entries in this list, that means we have a lot of ada-only dust, so we can't combine them anyways and we can fail

klntsky (changes requested):

src/BalanceTx/Collateral.purs
errfrom: @klntsky It's a good idea to sort utxos by amount in decreasing order before taking the first 10, but there is no need to impose any restrictions on whether an utxo should contain any tokens or not, since we add collateral return output afterwards

klntsky: It will be beneficial in terms of Tx size, right?

errfrom: @klntsky The algorithm will attempt to select the combination with fewer tokens, thus keeping the size of tx relatively small given all the constraints

errfrom:
@klntsky https://github.com/Plutonomicon/cardano-transaction-lib/pull/845

klntsky (approved):

PR #711, opened 2022-07-11
Add `runPlutipTest`

ngua:
Also fixes a problem with flags passed to ODC

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -156,6 +156,7 @@
         purescriptProject = import ./nix { inherit system; pkgs = prev; };
         buildCtlRuntime = buildCtlRuntime system;
         launchCtlRuntime = launchCtlRuntime system;
+        ctl-server = self.packages.${system}."ctl-server:exe:ctl-server";
         inherit cardano-configurations;
       });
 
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -459,24 +460,22 @@
...
+            {
+              ctl-unit-test = project.runPursTest {
+                name = "ctl-unit-test";
+                testMain = "CTL.Test.Unit";
+                env = { OGMIOS_FIXTURES = "${ogmiosFixtures}"; };
+              };
+              ctl-plutip-test = project.runPlutipTest {
+                name = "ctl-plutip-test";
+                testMain = "Test.Plutip";
+                env = { OGMIOS_FIXTURES = "${ogmiosFixtures}"; };
```
klntsky, line 476: I think we don't need ogmios response fixtures for plutip test.

klntsky (changes requested):

flake.nix
ngua: We do, running it without having the fixtures in the env will throw an error since they're used to create the test tree

klntsky (approved):

PR #712, opened 2022-07-11
Fix overlay, other improvements

### LINKED ISSUE #706: Make the flake overlay actually use the passed nixpkgs instance instead of using own instance

ngua:
Closes #706 and adds some other small Nix improvements (e.g. using `writeShellApplication` instead of `writeShellScriptBin` with overrides)


ngua:
Note: Ogmios itself wasn't upgraded, I just removed `inputs.cardano-node.flake = false` from the flake for our Ogmios fork so we could use its `packages` output

jy14898 (approved):

yusdacra:
Maybe you could base on the commit from https://github.com/yusdacra/cardano-transaction-lib/commits/fix-nix-overlay ? It fixes the issue we were experiencing on our end.

ngua:
Thanks @yusdacra, I've applied some of the same changes as in that commit

PR #716, opened 2022-07-12
Fix overlay (was #712)

### LINKED ISSUE #706: Make the flake overlay actually use the passed nixpkgs instance instead of using own instance

ngua:
Closes #706

Was #712, but Hydra wasn't building the branch, trying again


ngua:
Hmmm Hydra still isn't working, closing this

PR #718, opened 2022-07-12
Add env var to nix/default.nix to configure only writing package.json and its lock

jy14898:
### Pre-review checklist


ngua (commented):
I don't know if this should go in every `devShell` -- theoretically CTL users could use `npm` directly. I agree that we shouldn't support it for developing CTL itself though. We can put it in the flake though, in the `shellHook` argument for `purescriptProject.shell`

ngua:
Another option would be to add a boolean field to `purescriptProject.shell`, perhaps `packageLockOnly`, that would enable this, along with a comment explaining its purpose

jy14898:
Are they not in the same situation as CTL developers, where the nix shell is providing a conflicting/out of date node_modules?

jy14898:
I'm happy to put it in our override for now, and we can poll on monday

ngua:
That would be their choice to use `npm` directly or not. With this change we would be making that decision for them, which I don't think is ideal

The best solution then would be to add a flag as I mentioned in my second comment, that is by default set to true. Then users could opt into using `npm` and would be more aware of the possible issues that might come up

firefrorefiddle:
Anyone who has node_modules can not use `make format` anymore, which is required as per the pre-review checklist. So is there really a choice to use npm directly?

jy14898:
> Anyone who has node_modules can not use `make format` anymore, which is required as per the pre-review checklist. So is there really a choice to use npm directly?

The checklist is only for CTL development, but you are right in that `make format` becomes broken

ngua:
@firefrorefiddle That only applies to developing CTL itself and we all seem to agree that the environment variable should be set for us. The question is whether users should always have the same variable set

nix/default.nix
```diff
@@ -19,6 +19,9 @@
   # Autogenerated Nix from `spago2nix generate`
 , spagoPackages ? "${src}/spago-packages.nix"
   # Configuration that will be used to generate a `devShell` for the project
+  # shell.
+  # `shell.packageLockOnly`: When set to true, the `node_modules` folder will
+  # not be built when `npm` is used inside the shell.
 , shell ? { }
 , ...
 }:
```
jy14898, line 27: Is this a good place to place the documentation, or should I move it to the `shellFor` definition? @ngua

ngua: No, I'd move it to `shellFor`

nix/default.nix
```diff
@@ -84,6 +87,7 @@ let
         shellHook = ''
           export NODE_PATH="${nodeModules}/lib/node_modules"
           export PATH="${nodeModules}/bin:$PATH"
+          export NPM_CONFIG_PACKAGE_LOCK_ONLY="${if packageLockOnly then "true" else "false"}"
```
ngua, line 90: Since this only takes effect when set to `true`, why not just leave it unset if `packageLockOnly` is `false`? You could do

```
${pkgs.lib.optionalString packageLockOnly "export NPM_CONFIG_PACKAGE_LOCK_ONLY=true"}
```

If you want to set it either way, you can also do

```
export NPM_CONFIG_PACKAGE_LOCK_ONLY="${pkgs.lib.boolToString packageLockOnly}"
```

jy14898: Thanks, completely forgot about `optionalString`

nix/default.nix
jy14898: Seeing as none of the other parameters in `shellFor` are documented there, I moved it the comment to the `ctl-as-a-dependency` docs alongside the others

ngua (approved):

PR #729, opened 2022-07-13
Ardana/overlay fix

### LINKED ISSUE #706: Make the flake overlay actually use the passed nixpkgs instance instead of using own instance

Geometer1729:
Closes #706  .
### Pre-review checklist


flake.nix
```diff
@@ -149,10 +149,10 @@
             inputs.ogmios-datum-cache.defaultPackage.${system};
           ogmios = ogmios.packages.${system}."ogmios:exe:ogmios";
           ogmios-fixtures = ogmios;
-          purescriptProject = import ./nix { inherit system; pkgs = final; };
+          cardano-cli = cardano-node-exe.packages.${system}.cardano-cli;
```
ngua, line 152: This input no longer exists -- did you forget to update the lockfile?

ngua (dismissed):

ngua (approved):

flake.nix
Geometer1729: Thanks and sorry about that. My understanding of the nix is fairly limited I was mostly just working from a commit which works and one which doesn't and pulling over the differences.

PR #741, opened 2022-07-15
Implement `awaitTxConfirmed`

### LINKED ISSUE #739: Implement `awaitTxConfirmed`
Recently, there have been some reports of transactions getting silently dropped despite apparent submission success. The most likely cause for this is that the tx will no longer be valid (as specified by its validity intervals) before landing on a block

One way to help mitigate this is to implement `awaitTxConfirmed` even though we don't have support for full transaction fetching yet (#30). The basic idea is

- add a handler for ODC's [`GetTxByHash`](https://github.com/mlabs-haskell/ogmios-datum-cache#gettxbyhash) and add it to the relevant websocket initialization code
- ignore the return value for now (that is, don't bother finishing #575, the handler can return `Aeson` for now without parsing it)
- add `awaitTxConfirmed`, which should do the following:
  - calls `submit`
  - calls the `GetTxByHash` handler using the tx hash returned from the previous step
  - if the tx is returned by ODC, return the tx hash and exit, otherwise
  - wait one slot and perform the last step again
 
Note: The intended behavior for txs that are never confirmed is an infinite loop. This is consistent with Plutus' version of the same `Contract` effect. Users can, however, set a timeout on their call (e.g. consistent with the validity interval) and retry submission until a confirmation is obtained

Related: #30

klntsky:
Closes #739 

### Pre-review checklist

### DIFF
--- a/examples/Pkh2Pkh.purs
+++ b/examples/Pkh2Pkh.purs
@@ -24,7 +24,7 @@ import Contract.Monad
   , runContract_
   )
 import Contract.ScriptLookups as Lookups
-import Contract.Transaction (balanceAndSignTx, submit)
+import Contract.Transaction (awaitTxConfirmed, balanceAndSignTx, submit)
 import Contract.TxConstraints as Constraints
 import Contract.Value as Value
 import Contract.Wallet (mkNamiWalletAff)
@@ -62,4 +62,5 @@ main = launchAff_ $ do
       liftedM "Failed to balance/sign tx" $ balanceAndSignTx ubTx
     txId <- submit bsTx
     logInfo' $ "Tx ID: " <> show txId
-
+    awaitTxConfirmed txId
+    logInfo' $ "Tx sumbitted successfully!"
--- a/examples/Pkh2Pkh.purs
+++ b/examples/Pkh2Pkh.purs
@@ -25,7 +25,7 @@ import Contract.Monad
   )
 import Contract.ScriptLookups as Lookups
 import Contract.Transaction
-  ( awaitTxConfirmedWithTimeout
+  ( awaitTxConfirmed
   , balanceAndSignTx
   , submit
   )
@@ -33,7 +33,6 @@ import Contract.TxConstraints as Constraints
 import Contract.Value as Value
 import Contract.Wallet (mkNamiWalletAff)
 import Data.BigInt as BigInt
-import Data.Time.Duration (Seconds(Seconds))
 
 main :: Effect Unit
 main = launchAff_ $ do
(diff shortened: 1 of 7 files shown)


examples/Pkh2Pkh.purs
```diff
@@ -62,4 +67,5 @@ main = launchAff_ $ do
       liftedM "Failed to balance/sign tx" $ balanceAndSignTx ubTx
     txId <- submit bsTx
     logInfo' $ "Tx ID: " <> show txId
-
+    awaitTxConfirmedWithTimeout (Seconds 50.0) txId
+    logInfo' $ "Tx sumbitted successfully!"
```
jy14898, line 71:
```suggestion
    logInfo' $ "Tx submitted successfully!"
```

src/QueryM.purs
```diff
@@ -283,6 +289,30 @@ getDatumsByHashes :: Array DataHash -> QueryM (Map DataHash Datum)
...
+  go timeoutTime = do
+    isTxFound <- unwrap <$> mkDatumCacheRequest DcWsp.getTxByHash _.getTxByHash
+      txHash
+    nowMs <- getNowMs
+    when (nowMs >= timeoutTime) do
+      liftEffect $ throw $
+        "awaitTxConfirmedWithTimeout: timeout exceeded, Transaction not \
+        \confirmed"
+    liftAff $ delay $ wrap 1000.0
+    if isTxFound then pure unit else go timeoutTime
```
jy14898, line 314: The docstring for `Contract.Transaction.awaitTxConfirmed` implies it doesn't have a timeout. I think either it should be updated to say it has a default timeout, or make `awaitTxConfirmedWithTimeout` accept `Maybe Seconds`?

jy14898 (changes requested):

src/QueryM.purs
klntsky: `infinity` is a valid `Number`

ngua:
@rmgaray mentioned on Slack that this is working as intended with the SN codebase

src/QueryM.purs
```diff
@@ -283,6 +290,30 @@ getDatumsByHashes :: Array DataHash -> QueryM (Map DataHash Datum)
 getDatumsByHashes hashes = unwrap <$> do
   mkDatumCacheRequest DcWsp.getDatumsByHashesCall _.getDatumsByHashes hashes
 
+awaitTxConfirmed :: TxHash -> QueryM Unit
+awaitTxConfirmed = awaitTxConfirmedWithTimeout (Seconds infinity)
+
+awaitTxConfirmedWithTimeout :: Seconds -> TxHash -> QueryM Unit
```
ngua, line 296: I don't want to hold this PR up for this, but ideally this would take a number of `Slot`s to wait instead of seconds

src/QueryM/DatumCacheWsp.purs
```diff
@@ -115,10 +118,32 @@ instance DecodeAeson GetDatumsByHashesR where
         r
         [ "DatumsFound", "value" ]
 
+newtype GetTxByHashR = GetTxByHashR Boolean
```
ngua, line 121: I think it might make sense to link to the `getTxById` issue here, since this will need to change when that's implemented

ngua (commented):
There's a note in the documentation that we don't support `awaitTxConfirmed` which we need to remove

ngua (approved):

PR #758, opened 2022-07-20
Bump ODC

ngua:
Bumps ODC to version where deserialization issue is fixed (https://github.com/mlabs-haskell/ogmios-datum-cache/pull/109)


klntsky (approved):

klntsky:
Do we need this for `babbage` as well?

PR #763, opened 2022-07-21
Only include the necessary cost models during transaction finalization

jy14898:
Fixes being able to run scripts (well, some of the examples now) on the `babbage` branch

### Pre-review checklist

### DIFF
--- a/src/Serialization.purs
+++ b/src/Serialization.purs
@@ -832,6 +832,7 @@ convertCostmdls (T.Costmdls cs) = do
   costmdls <- newCostmdls
   -- TODO Check the traversal order of the costmodels map isn't important for
   --      mixed language transactions
+  --      https://github.com/Plutonomicon/cardano-transaction-lib/issues/764
   forWithIndex_ cs \language costModel -> do
     language' <- case language of
       S.PlutusV1 -> newPlutusV1
--- a/src/Serialization.purs
+++ b/src/Serialization.purs
@@ -830,9 +830,6 @@ convertValue val = do
 convertCostmdls :: T.Costmdls -> Effect Costmdls
 convertCostmdls (T.Costmdls cs) = do
   costmdls <- newCostmdls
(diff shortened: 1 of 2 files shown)


src/Serialization.purs
```diff
@@ -830,15 +830,15 @@ convertValue val = do
...
-      costModel <-
-        convertCostModel <=< fromJustEff
-          ("`" <> show lang <> "` not found in `Costmdls`")
-          $ Map.lookup lang cs
-      costmdlsSetCostModel costmdls lang' costModel
-  loadCostModel S.PlutusV1 =<< newPlutusV1
-  loadCostModel S.PlutusV2 =<< newPlutusV2
+  -- TODO Check the traversal order of the costmodels map isn't important for
+  --      mixed language transactions
+  --      https://github.com/Plutonomicon/cardano-transaction-lib/issues/764
```
klntsky, line 835: As long as I'm not missing something obvious, the cost models go to different keys, so why would there be any problem? I think this TODO is not needed, as well as the issue

jy14898: I'm not sure how costmodels is serialized when hashing as part of the transaction, which might make things not line up when hashes are being checked if order is important?

klntsky: OK but what is the reason you think the order may not be fixed?

jy14898: I don't have any information for or against ordering, and I'm not sure whose responsibility the order is, ours or CSL? I don't mind dropping the TODO if you'd like, I thought it didn't hurt to be more rigorous and also to include a hint if the issue does arise later

klntsky: > whose responsibility the order is, ours or CSL?

CSL's. Costmdls is a wrapper over `std::collections::BTreeMap` (rust) and it's a sorted map.

jy14898: OK great, I'll drop the todo then :+1:

klntsky (approved):
Great, thanks!

PR #765, opened 2022-07-21
Properly implement CIP-25 V2 metadata

### LINKED ISSUE #752: CIP 25 Metadata not following spec
There's two issues that I can see:

**Issue 1**: When CTL adds metadata to transactions with `setTxMetadata` it seems to add `0x` before byte strings (I'm guessing this is just the cardano metadata standard for byte strings). I think this might be an issue with [CIP25](https://cips.cardano.org/cips/cip25/) and CTL's support for it: the spec requires the `asset_name` and `policy_id` to be "utf-8 encoded and in text format" for version 1, and CTL includes them as raw bytes (and CTL uses version 1 by default because the `version` property is not included). An example of metadata created by CTL using `setTxMetadata` with `Cip25Metadata`:

```json
[
  {
    "721": {
      "0xda90b3ed7f25fcb4f43794e568d8397afe05de82399d78e27a6c0fe8": {
        "0xabcdef": {
          "description": [
            "Allium",
            "Seabug Testing"
          ],
          "files": [],
          "image": [
            "ipfs://k2cwuee3arxg398hwxx6c0iferxitu126xntuzg8t765oo020h5y6npn"
          ],
          "name": "0xabcdef"
        }
      }
    }
  }
]
```

I think this would be valid if `"version": 2` were added because version 2 uses raw bytes for `policy_id` and `asset_name`. For reference this is how cardanoscan renders that metadata:

2022-07-18-110712_807x400_scrot

**Issue 2**: The `name` field is supposed to be a string but is also included as raw bytes (with the `0x` prefix) by CTL (it seems to just use the token name). See the previous example.

klntsky:
Closes #752 

### Pre-review checklist


src/Metadata/Cip25/Cip25String.purs
```diff
@@ -0,0 +1,145 @@
...
+unCip25String :: Cip25String -> String
+unCip25String (Cip25String str) = str
+
+-- | A smart constructor for `Cip25String`
+mkCip25String :: String -> Maybe Cip25String
+mkCip25String str
+  | byteLength (wrap (encodeUtf8 str)) <= 64 = Just $ Cip25String str
+  | otherwise = Nothing
+
+takeCip25String :: String -> Maybe { init :: Cip25String, rest :: Maybe String }
```
jy14898, line 72: I appreciate the work put in to generalize with `forwardSearch` etc, however, I feel like this would be more simply implemented by repeatedly `Data.String.CodePoints.uncons` and applying `Data.String.CodePoints.fromCodePointArray` until failure to `mkCip25String`? Am I missing something?

klntsky: This is basically no different from `forwardSearch`, except that the step is set to 1.

I don't understand how CodePoints are related.

> CodePoint is an Int bounded between 0 and 0x10FFFF

jy14898: For the same reason you're using them here, we iterate using codepoints such that we don't break the encoding

jy14898: Also, I hadn't seen that step is being divided in the final case - this feels like a premature optimisation to me though

jy14898: Actually it seems like a mistake to me that you're using CodePoints in your implementation, right?

jy14898: Maybe it's not, to me this isn't straight forward to understand

jy14898: I think I understand it now, 16 point points might be exactly 64 bytes, but if it's not, it'll always be less, so we attempt take 24 codepoints (why this amount?) and try and fill up the rest of the remaining bytes with smaller and smaller amounts until it completes.

klntsky: (64-16) / 2 = 24

klntsky: > For the same reason you're using them here, we iterate using codepoints such that we don't break the encoding

Yes, but we don't need to access actual code points as numbers.

klntsky: > this feels like a premature optimisation to me though

It's making an algorithm that is quadratic in time to be `O(n * log(n))`.

jy14898: > Yes, but we don't need to access actual code points as numbers.
The number repr wouldn't be accessed, it's just so we never split a character incorrectly

klntsky: Then it does not matter if we use code point array slicing or string slicing

jy14898: Yeah, the suggestion wasn't from a correctness perspective, just an alternate implementation, sorry that wasn't clear. I think on average your implementation will have slightly better performance, it's unclear to me if that performance benefit is worth the extra code.

It's also worth bringing up, what are we supposed to do with combining characters/diacritics?

klntsky: I think that if the following property holds, we are safe:

```
forall ix str. 
  encodeUtf8 ((String.splitAt ix str).head) <> encodeUtf8 ((String.splitAt ix str).tail) == 
    encodeUtf8 str
```

jy14898: > It's making an algorithm that is quadratic in time to be `O(n * log(n))`

the `n` here is pretty small though, plus it's trivial to turn the quadratic algorithm into linear by using Data.String.CodeUnits.length to get how many bytes are in the codepoint (represented as a string).

I don't think what you've written is `O(n * log(n))` anyway, due to invoking `encodeUtf8` at each test?

If we really care about performance we can do better though, using `TextDecoder` we can take 64 bytes and decode them, count how many replacement characters at the end (the only place decoding can fail if the input was valid, and only up to 3), drop them, and use that information to aid in decoding the next. I'll create a mockup so you can decide for yourself

src/Metadata/Cip25/Cip25String.purs
```diff
@@ -0,0 +1,145 @@
...
+  decodeAeson = decodeAeson >=> mkCip25String >>> note
+    (TypeMismatch "Cip25String")
+
+unCip25String :: Cip25String -> String
+unCip25String (Cip25String str) = str
+
+-- | A smart constructor for `Cip25String`
+mkCip25String :: String -> Maybe Cip25String
+mkCip25String str
+  | byteLength (wrap (encodeUtf8 str)) <= 64 = Just $ Cip25String str
```
jy14898, line 69:
```suggestion
  | Data.String.CodeUnits.length str <= 64 = Just $ Cip25String str
```

jy14898: Nvm, CodeUnits.length isn't the same as bytes

src/Metadata/Cip25/Cip25String.purs
jy14898: I think this is about as fast as it gets in terms of big o, but also most operations are delegated to node/the browser, and it's in tail call form. Needs more testing, but what do you think?

https://github.com/Plutonomicon/cardano-transaction-lib/blob/df914937c0b36903f98150c1d150f5b34d20126f/src/Split64.purs#L26-L49

jy14898: Just to be clear: I'm happy with your current implementation, but maybe with a comment explaining the approach

jy14898: https://github.com/Plutonomicon/cardano-transaction-lib/blob/1c4afec40278ee21a62c20cee7f95f246ea07256/src/Split64.purs#L26-L27

@klntsky It's probably better we specify it using escape characters, it's a standard character used to replace invalid data in unicode streams. I seem to remember seeing a specification somewhere that TextDecoder will replace it, but there's at least some reference here https://infra.spec.whatwg.org/#javascript-string-convert.

src/Metadata/Cip25/Cip25String.purs
```diff
@@ -0,0 +1,146 @@
...
+      , rest:
+          let
+            rest = String.drop ix str
+          in
+            if rest == "" then Nothing else Just rest
+      }
+
+forwardSearch
+  :: forall a
+   . { step :: Int, minBound :: Int, maxBound :: Int, isTrue :: Int -> Maybe a }
```
jy14898, line 101: Can we rename `isTrue` to `parseN` or something?

src/Metadata/Cip25/V2.purs
```diff
@@ -0,0 +1,348 @@
...
+  [ "name" /\ toData entry.name
+  , "image" /\ toData entry.image
+  ]
+    <> mbMediaType
+    <> mbDescription
+    <> mbFiles
+  where
+  mbFiles = case entry.files of
+    [] -> []
+    files -> [ "files" /\ toData files ]
```
jy14898, line 176: I think this case can be done the same as the others, right?

src/Metadata/Cip25/V2.purs
```diff
@@ -0,0 +1,348 @@
...
+
+derive instance Generic Cip25Metadata _
+derive instance Newtype Cip25Metadata _
+derive instance Eq Cip25Metadata
+
+instance Show Cip25Metadata where
+  show = genericShow
+
+instance MetadataType Cip25Metadata where
+  metadataLabel _ = wrap (BigInt.fromInt 721)
```
jy14898, line 224: Shame about the unsafePartial, up to you
```suggestion
  metadataLabel _ = wrap (unsafePartial $ fromJust $ BigInt.fromString nftMetadataLabel)
```

src/Metadata/Cip25/V2.purs
```diff
@@ -0,0 +1,348 @@
...
+
+groupEntries
+  :: Array Cip25MetadataEntry -> Array (NonEmptyArray Cip25MetadataEntry)
+groupEntries =
+  groupBy \(Cip25MetadataEntry a) (Cip25MetadataEntry b) ->
+    a.policyId == b.policyId
```
jy14898, line 230:
```suggestion
  group `on` (unwrap >>> _.policyId)
```

src/Metadata/Cip25/V2.purs
```diff
@@ -0,0 +1,348 @@
...
+      $ for (Map.toUnfoldable mp1)
+          \(key /\ assets) ->
+            if key == toMetadata "version" then
+              -- top-level version tag
+              ( if assets == toMetadata Cip25V2 then pure Nothing
+                else Nothing -- fail if version does not match
+              )
+            else Just -- key is policyId
+
+              case assets of
```
jy14898, line 258: I think the comment made the formatting odd here, is it possible to move it elsewhere?

src/Metadata/Cip25/Cip25String.purs
```diff
@@ -0,0 +1,146 @@
...
+      { init: cip25String
+      , rest:
+          let
+            rest = String.drop ix str
+          in
+            if rest == "" then Nothing else Just rest
+      }
+
+forwardSearch
+  :: forall a
```
jy14898, line 100: Kind annotation missing, first of a couple

src/Metadata/Cip25/V2.purs
klntsky: It's Array instead of Maybe, so no

jy14898 (approved):

PR #770, opened 2022-07-22
Update ogmios-datum-cache (to the version with darwin build)

aske:
### Pre-review checklist

- [ ] All code has been formatted using our config (`make format` for Purescript, `nixpkgs-fmt` for Nix)
- [ ] All Purescript imports are explicit, including constructors
- [ ] **All** existing examples have been run locally against a fully-synced testnet node
- [ ] The integration and unit tests have been run (`npm run test`) locally
- [ ] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`)

PR #773, opened 2022-07-22
Update ogmios-datum-cache (to the version with darwin build)

klntsky:
Replacement of https://github.com/Plutonomicon/cardano-transaction-lib/pull/770

### Pre-review checklist


firefrorefiddle (commented):
I have no opinion about the upgrade to ODC in general. However, I'm not sure what's the intended purpose is, and if the original problem will be solved.

I made some mixed experiences with M1 mac a couple months ago. When ODC builds Darwin binaries, they don't run in Docker/Arion, because Docker expects Linux binaries. So if anyone would like to use `ctl-runtime` on the Mac, then Nix should be convinced to build Docker images for Linux.

ngua (approved):

PR #775, opened 2022-07-22
Add balanceTxWithOwnAddress and balanceTxsWithOwnAddress

### LINKED ISSUE #746: Add variant of `balanceTx` with `Address` as parameter
Quoting myself from Slack:

> The only way we could possibly support that is to change `balanceTx` internally. We could have
> ```
> balanceTx'
>   :: UnattachedUnbalancedTx
>   -> Maybe Address
>   -> QueryM (Either BalanceTxError FinalizedTransaction)
> ```
> 
> 
> And then `balanceTx` would be `balanceTx' Nothing`, and we could add `balanceTxWith` or something that could take a `PaymentPubKeyHash` and `StakePubKeyHash` that get converted to an `Address` before calling `balanceTx'`. And I just checked, we'd only have to update a small part of the current `balanceTx` -- `getWalletAddress` only gets called once and then the `Address` is passed as an explicit parameter elsewhere
> 
> Of course this would be quite different from Plutus. But since there's no new version of `plutus-apps` on the horizon, I think we're safe to move even further away from Plutus than we already have

This would be useful to at least one project, and perhaps more

klntsky:
Closes #746 

### Pre-review checklist

### DIFF
--- a/src/BalanceTx/BalanceTx.purs
+++ b/src/BalanceTx/BalanceTx.purs
@@ -46,7 +46,7 @@ module BalanceTx
   , UtxosAtError(CouldNotGetUtxos)
   , UtxoMinAdaValueCalcError(UtxoMinAdaValueCalcError)
   , balanceTx
-  , balanceTxWithOwnAddress
+  , balanceTxWithAddress
   ) where
 
 import Prelude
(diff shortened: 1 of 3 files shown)


src/BalanceTx/BalanceTx.purs
```diff
@@ -475,22 +476,22 @@ addTxCollateral utxos transaction =
...
 -- multisig but we don't have such functionality ATM.
 
--- | Balances an unbalanced transaction. For submitting a tx via Nami, the
--- | utxo set shouldn't include the collateral which is vital for balancing.
--- | In particular, the transaction inputs must not include the collateral.
-balanceTx
-  :: UnattachedUnbalancedTx
+-- | Like `balanceTx`, but allows to provide an address that is treated like
+-- | user's own (while `balanceTx` gets it from the wallet).
+balanceTxWithOwnAddress
```
ngua, line 481: I think the implementation is fine, but the name might be a bit confusing. `withOwnAddress` makes me think of the attached wallet's address, not an external user-provided address. Maybe just `withAddress` would be clearer?

ngua (approved):

PR #776, opened 2022-07-23
Get Plutip integration working in Babbage era

ngua:
Fixes Plutip integration to work in Babbage era. I also needed to change a few things to get the script loading working with both Webpack and Node. Anyway, `Test.Plutip` now works! I also merged `develop` in so the branch will be up to date


package.json
```diff
@@ -35,7 +35,7 @@
     "buffer": "6.0.3",
     "html-webpack-plugin": "5.5.0",
     "webpack": "5.67.0",
-    "webpack-cli": "4.9.2",
+    "webpack-cli": "4.10",
```
klntsky, line 38: Let's make a new `plutip-test` entry in `package.json` (we should also run it as part of `npm run test`)

klntsky (approved):

PR #795, opened 2022-07-27
Add faq.md

klntsky:


ngua (commented):
I might suggest changing the document name/title to `faq.md`

klntsky:
OK sounds less frigtening

firefrorefiddle:
> I might suggest changing the document name/title to `faq.md`

Then the contents should be rephrased as questions.

jy14898 (approved):
Approving as the comments have been addressed

PR #802, opened 2022-07-28
BalanceTx: Prevent non-Ada change from having zero-valued assets

errfrom:
This PR fixes an issue when zero-valued non-Ada assets were added to the non-Ada change output, thus causing errors when estimating fees and calculating utxo min ada value for outputs containing those assets.

The problem was that we used the `minus` function from `Cardano.Types.Value` to build `nonAdaChange`, which doesn't guard against zero-valued non-Ada assets in `Value`.

For now I added a new `minusNonZero` function to solve the problem, but I wonder if it would be better to update the `Semigroup` instance of `Value` to not allow zero-valued non-Ada assets after merging two `Value`s. 
What do you think @ngua @klntsky? Do we have any cases where current behavior can be useful? 

### Pre-review checklist


klntsky (changes requested):
> if it would be better to update the Semigroup instance of Value to not allow zero-valued non-Ada assets after merging two Values.

Yes, I think it should be done this way.

examples/KeyWallet/MintsAndSendsToken.js
```diff
@@ -0,0 +1 @@
+exports.alwaysMints = require("Scripts/always-mints.plutus");
```
ngua, line 1: This won't work in a nodejs environment, right? I think you need to use the `fs.readSync` trick in the other examples. Otherwise you could import `Examples.AlwaysMints.alwaysMintsPolicy`

examples/KeyWallet/MintsAndSendsToken.purs
```diff
@@ -0,0 +1,60 @@
+-- | This module demonstrates how the `Contract` interface can be used to build,
+-- | balance, and submit a smart-contract transaction. It creates a transaction
+-- | that mints a token using the `AlwaysMints` policy and sends it along with
+-- | the selected amount to the specified address.
+module Examples.KeyWallet.MintsAndSendsToken where
```
ngua, line 5: Missing export list

ngua (changes requested):
Were you planning on modifying the `Semigroup` instance as mentioned above?

errfrom:
@ngua I've already updated the construction and merging behavior for `Value` in this commit https://github.com/Plutonomicon/cardano-transaction-lib/pull/802/commits/4ce7261a62c58e9ffb7ecc6302d9f68c0e565f00, although it didn't include modifying the `Semigroup` instance directly

errfrom:
@klntsky @ngua I'm not sure if we should wait for the results of external testing to merge this PR. I would suggest opening subsequent PRs in case some new related problems are discovered.

klntsky (approved):

ngua (approved):

PR #829, opened 2022-08-01
Add getTxByHash

### LINKED ISSUE #824: Add getTxByHash
From [Ogmios changelog](https://github.com/CardanoSolutions/ogmios/blob/master/CHANGELOG.md#530---2022-05-07)

> Transaction JSON objects from all eras now contains an extra field raw, which represents the raw serialized transaction (CBOR) as a base64-encoded text string. This is the case of the chain-sync protocol, but also for the tx-monitor protocol. The field is however absent in the ogmios.v1:compact mode. See https://github.com/CardanoSolutions/ogmios/issues/190.

That means we can now deserialize cborbase64 with CSL and implement getTxByHash without the need to decode Json.

klntsky:
Closes #824

### Pre-review checklist


src/Base64.js
```diff
@@ -0,0 +1,13 @@
...
+
+exports.toByteArray = base64.toByteArray;
+
+exports._decodeBase64 = maybe => str => {
+  try {
+    return maybe.just(base64.toByteArray(str));
+  } catch (_) {
+    return maybe.nothing;
+  }
+};
```
jy14898, line 13: I don't think we need to add a new library for this - we already have a shim for nodes buffer set up in webpack to allow browsers to use it. We can use `node-buffer` which is already a dependency and https://pursuit.purescript.org/packages/purescript-node-buffer/8.0.0/docs/Node.Encoding#v:Base64

jy14898 (changes requested):

src/Base64.js
klntsky: I don't get how to use that package. Is it only interfaces? There isn't a single instance of [MutableBuffer](https://pursuit.purescript.org/packages/purescript-node-buffer/8.0.0/docs/Node.Buffer.Class#t:MutableBuffer) class.

klntsky: base64-js is 150 lines of pure JS, it's not a big deal. As a bonus it would be easier to switch away from webpack.

jy14898: Well, we don't have to use that purescript library, but IMO it's still better to use what we already have. We're already doing something very similar over in `TypedArray`, so we'd at least be being consistent:
https://github.com/Plutonomicon/cardano-transaction-lib/blob/6c018d43ddfef771bfd586e885ecb0cc7ebd8421/src/Types/ByteArray.js#L33

I think webpack only saves us from explicitly requiring buffer, we could move away from webpack and still keep the shim.

...

Just had a look into it, the `Buffer` shim's b64 conversion is implemented using `base64-js` anyway. I'm still slightly inclined to prefer using one interface for conversions but I don't think it matters, up to you

klntsky: When using Buffer directly, we lose validity checks for base64.

```
> new Uint8Array(Buffer.from('#$#$!(*#)(*$#)*', 'base64'))
Uint8Array(0) []
```

klntsky: > I'm still slightly inclined to prefer using one interface for conversions

Also note that we don't use `MutableBuffer` interface elsewhere. We only depend on `node-buffer` to pass an `Encoding` value to `readTextFile` in one place.

jy14898 (dismissed):

jy14898 (approved):
Approved given CI passes

PR #838, opened 2022-08-03
Add staking keys support for plutip integration

### LINKED ISSUE #659: Add staking keys to plutip-server
For now plutip-server does not provide stake keys, because they are not available in plutip. However, they are in BPI:

https://github.com/mlabs-haskell/bot-plutus-interface/blob/master/src/BotPlutusInterface/Config.hs#L147
https://github.com/mlabs-haskell/plutip/blob/master/src/Test/Plutip/Internal/BotPlutusInterface/Run.hs#L130

We can extend plutip.

### LINKED ISSUE #823: Add support for staking keys with plutip
From slack:

> I don't see a way to access the private payment keys returned from plutip to use with mkKeyWallet

There's no way really.
I think that instead of exposing them, we should add a way to specify staking keys in plutip wallet spec. Something like:
`distr = withStakeKey sk1 [ BigInt.fromInt 1000000, BigInt.fromInt 1000000 ]`

We can change the type class machinery to support that very easily.

rynoV:
Closes #823 .

### Pre-review checklist


src/Plutip/Types.purs
```diff
@@ -36,26 +31,47 @@ import Aeson
...
+import Effect.Class.Console (log)
 import QueryM.ServerConfig (ServerConfig)
 import Serialization (privateKeyFromBytes)
 import Serialization.Types (PrivateKey)
 import Types.ByteArray (hexToByteArray)
 import Types.RawBytes (RawBytes(RawBytes))
 import Wallet.Key
   ( KeyWallet
-  , PrivatePaymentKey(PrivatePaymentKey)
+  , PrivatePaymentKey(..)
```
klntsky, line 73: We use explicit imports and exports, including constructors. Please undo these changes here and above

src/Plutip/Types.purs
```diff
@@ -197,27 +216,138 @@ instance DecodeAeson StopClusterResponse where
...
+
+instance UtxoDistribution InitialUTxOWithStakeKey KeyWallet where
+  encodeDistribution (InitialUTxOWithStakeKey _ amounts) = [ amounts ]
+  decodeWallets ourKey (InitialUTxOWithStakeKey stake _) [ key ] =
+    decodeWallet ourKey key (Just stake)
+  decodeWallets _ _ _ = wrongNumberPrivateKeysError
+
+instance
+  UtxoDistribution restSpec restWallets =>
+  UtxoDistribution (InitialUTxOWithStakeKey /\ restSpec)
```
klntsky, line 251: Shouldn't we generalize these two recursive instances?

```
  UtxoDistribution headSpec headWallet =>
  UtxoDistribution restSpec restWallets =>
  UtxoDistribution (headSpec /\ restSpec) (headWallet /\ restWallets) where
```

and remove `encodeDistributionRecurse`

src/Plutip/Types.purs
```diff
@@ -197,27 +216,138 @@ instance DecodeAeson StopClusterResponse where
...
+decodeWalletsRecurse ourKey distr rest = Array.uncons >>> case _ of
+  Nothing -> wrongNumberPrivateKeysError
+  Just { head, tail } -> do
+    wallet <- decodeWallets ourKey distr [ head ]
+    Tuple wallet <$> decodeWallets ourKey rest tail
+
+-- | Get a wallet for a single private key response, and if a stake
+-- | key is provided for it, transfer the funds allocated by plutip to
+-- | the address combining the payment key and the stake key
+decodeWallet
```
klntsky, line 286: I think we should use one contract to fund for all addresses at once. We can add multiple `mustPayToPubKeyAddress` constraints.

This will require more code, but it gives linear performance improvement which is important for tests.

klntsky (changes requested):

src/Plutip/Server.purs
```diff
@@ -95,26 +100,40 @@ withPlutipContractEnv
...
   withPlutipServer =
     withResource (startPlutipServer plutipCfg) stopChildProcess <<< const
 
-  withPlutipCluster :: (ClusterStartupParameters -> Aff a) -> Aff a
-  withPlutipCluster cc = withResource (startPlutipCluster plutipCfg distr)
+  withPlutipCluster
+    :: ((PrivatePaymentKey /\ ClusterStartupParameters) -> Aff a) -> Aff a
+  withPlutipCluster cc = withResource
+    ( startPlutipCluster plutipCfg
+        ( [ BigInt.fromInt 1_000_000_000, BigInt.fromInt 1_000_000_000 ] /\
```
klntsky, line 119: We should insert the extra keys elsewhere. And we should calculate our amounts based on requested amounts.

src/Plutip/Types.purs
```diff
@@ -197,27 +216,138 @@ instance DecodeAeson StopClusterResponse where
...
+  :: PrivatePaymentKey
+  -> PrivateKeyResponse
+  -> Maybe PrivateStakeKey
+  -> Contract () KeyWallet
+decodeWallet ourKey (PrivateKeyResponse key) stake = do
+  let
```
klntsky, line 292: Move these bindings under `when` (PS is trict)

src/Plutip/Types.purs
rynoV: Sounds like we should just move the transferring logic out of `decodeWallets` into a separate function we call once after decoding the wallets.

I'm not super familiar with transaction size limits, I wonder if we have to worry about making the single transaction too big

src/Plutip/Types.purs
rynoV: Is there an automated way to do this? Emacs is switching to implicit, I'll have to check if it can do the opposite

src/Plutip/Server.purs
```diff
@@ -143,12 +154,22 @@ withPlutipContractEnv plutipCfg distr cont = do
...
-        " wallets from private keys. Please report as bug."
-    Just wallets -> cc wallets
+  withWallets
+    :: ContractEnv ()
+    -> PrivatePaymentKey
+    -> ClusterStartupParameters
+    -> (wallets -> Aff a)
+    -> Aff a
+  withWallets env ourKey response cc = do
+    wallets <- runContractInEnv env do
```
rynoV, line 164: I wonder if we should change the log level before we run this, so the user doesn't get flooded with our traces

src/Plutip/Types.purs
```diff
@@ -197,27 +231,116 @@ instance DecodeAeson StopClusterResponse where
 -- | distribution for wallets.
 -- | Number of wallets in distribution specification matches the number of
 -- | wallets provided to the user.
-class UtxoDistribution distr wallets | distr -> wallets, wallets -> distr where
+class UtxoDistribution distr wallets | distr -> wallets where
   encodeDistribution :: distr -> Array (Array UtxoAmount)
-  decodeWallets :: Array PrivateKeyResponse -> Maybe wallets
+  decodeWallets :: distr -> Array PrivateKeyResponse -> Maybe wallets
+  keyWallets :: Proxy distr -> wallets -> Array KeyWallet
```
rynoV, line 237: Couldn't find a way to get this to type check without the proxy. Giving type annotations where the method is used like `keyWallets :: blah` didn't work

src/Plutip/Types.purs
```diff
@@ -197,27 +231,116 @@ instance DecodeAeson StopClusterResponse where
...
+type WalletInfo =
+  { utxos :: Utxo
+  , payPkh :: PaymentPubKeyHash
+  , stakePkh :: StakePubKeyHash
+  , wallet :: KeyWallet
+  }
+
+-- | For each wallet which includes a stake key, transfer the value of
+-- | the utxos at its enterprise address to its base address.
+transferFundsFromEnterpriseToBase
```
rynoV, line 283: Maybe this would fit better in a different file

src/Plutip/Types.purs
```diff
@@ -197,27 +231,116 @@ instance DecodeAeson StopClusterResponse where
...
+  -> Contract r Unit
+transferFundsFromEnterpriseToBase ourKey wallets = do
+  -- Get all utxos and key hashes at all wallets containing a stake key
+  walletsInfo <- foldM addStakeKeyWalletInfo mempty wallets
+  unless (null walletsInfo) do
+    let ourWallet = privateKeysToKeyWallet ourKey Nothing
+    ourAddr <- liftedM "Could not get our address"
+      $ withKeyWallet ourWallet getWalletAddress
+    ourUtxos <- liftedM "Could not find our utxos" $ utxosAt ourAddr
+    ourPkh <- withKeyWallet ourWallet $ liftedM "Could not get our payment pkh"
```
rynoV, line 296: Should make the order of `withKeyWallet` and `liftedM` consistent

src/Plutip/Server.purs
klntsky: We should provide a `customLogger` that logs nothing.

src/Plutip/Types.purs
klntsky: a Proxy is OK

src/Plutip/Types.purs
klntsky: Let's move all `UtxoDistribution` machinery to `Plutip.UtxoDistribution`

src/Plutip/Server.purs
```diff
@@ -209,7 +240,33 @@ startPlutipCluster cfg utxoDistribution = do
...
+ourInitialUtxos :: InitialUTxODistribution -> InitialUTxO
+ourInitialUtxos utxoDistribution =
+  let
+    numUtxos = foldr (length >>> add) 0 utxoDistribution
+  in
+    [ -- For each utxo, include an excessive amount of ada to cover
+      -- the transaction fees of including that utxo. Also make sure
+      -- we don't request a 0 ada utxo
+      BigInt.fromInt 1_000_000_000 +
+        (BigInt.fromInt numUtxos * BigInt.fromInt 1_000_000_000)
```
klntsky, line 266: No, we should actually sum up the amounts and add extra on top of it.

klntsky (changes requested):

test/Plutip.purs
```diff
@@ -136,10 +133,9 @@ suite = do
       withResource (startPlutipServer config)
         (stopChildProcessWithPort config.port) $ const do
         startRes <- startPlutipCluster config unit
-        startRes `shouldSatisfy` case _ of
```
klntsky, line 139: Why did you remove the assertion?

test/Plutip.purs
```diff
@@ -136,10 +133,9 @@ suite = do
...
-          ClusterStartupSuccess _ -> true
-          _ -> false
-        liftEffect $ Console.log $ "startPlutipCluster: " <> show startRes
+        liftEffect $ Console.log $ "startPlutipCluster: "
```
klntsky, line 136: Please remove the change to the logging on this line.

src/Plutip/Types.purs
```diff
@@ -197,27 +231,116 @@ instance DecodeAeson StopClusterResponse where
...
+          signTransaction tx
+      )
+      (unwrap signedTx)
+      walletsInfo
+    txHash <- submit (wrap signedTx')
+    awaitTxConfirmed txHash
+  where
+  constraintsForWallet :: WalletInfo -> Constraints.TxConstraints Unit Unit
+  constraintsForWallet { utxos, payPkh, stakePkh } =
+    -- It's necessary to include `mustBeSignedBy`, we get a
```
klntsky, line 319: No, this is probably a bug on our side. `mustBeSignedBy` should not be required in any case. Please create a new issue:

- Include a permanent link to this line
- add a link to the issue in the comments here. Use `TODO` keyword

src/Plutip/Server.purs
```diff
@@ -209,7 +240,33 @@ startPlutipCluster cfg utxoDistribution = do
...
+    , -- In case a collateral utxo is needed
```
klntsky, line 267: It is not, because we don't use scripts with this wallet. Please remove the second UTxO.

test/Plutip.purs
rynoV: Because what it was testing is now tested just by the function not throwing an exception

test/Plutip.purs
rynoV: I can't exactly because `PrivatePaymentKey` doesn't have a show instance, I can remove the `<ourKey> /\` part though

klntsky: Yes, please remove the `<ourkey>`

src/Plutip/Server.purs
```diff
@@ -179,14 +205,24 @@ configCheck cfg = do
   printServiceEntry (port /\ service) =
     "- " <> service <> " (port: " <> show (UInt.toInt port) <> ")\n"
 
+-- | Start the plutip cluster, initializing the state with the given
+-- | utxo distribution. Also initializes an extra payment key (aka
+-- | `ourKey`) with some utxos for use with further plutip
+-- | setup. `ourKey` has funds proportional to the number of utxos in
```
klntsky, line 211:
```suggestion
-- | setup. `ourKey` has funds proportional to the total amount of utxos in
```

src/Plutip/UtxoDistribution.purs
```diff
@@ -0,0 +1,175 @@
...
+  UtxoDistribution (headSpec /\ restSpec) (headWallets /\ restWallets) where
+  encodeDistribution (distr /\ rest) =
+    encodeDistribution distr <> encodeDistribution rest
+  decodeWallets (distr /\ rest) =
+    Array.uncons >=>
+      \{ head, tail } -> do
+        wallet <- decodeWallets distr [ head ]
+        Tuple wallet <$> decodeWallets rest tail
+  keyWallets _ (headWallets /\ restWallets) =
+    (keyWallets (Proxy :: Proxy headSpec) headWallets)
```
klntsky, line 97: No need for parentheses

test/Plutip.purs
```diff
@@ -160,80 +180,79 @@ suite = do
...
+          [ BigInt.fromInt 1_000_000_000
+          , BigInt.fromInt 2_000_000_000
+          ]
+
         distribution :: InitialUTxO /\ InitialUTxO
-        distribution =
+        distribution = aliceUtxos /\ bobUtxos
+      withPlutipContractEnv config distribution \env (alice /\ bob) -> do
+        runContractInEnv env do
+          assertCorrectDistribution [ alice /\ aliceUtxos, bob /\ bobUtxos ]
```
klntsky, line 223: Very good testing

test/Plutip.purs
```diff
@@ -410,3 +465,95 @@ mkCurrencySymbol mintingPolicy = do
...
+    utxos <- liftedM "Could not get wallet utxos" $ map unwrap <$> utxosAt addr
+    assertContract "Incorrect distribution of utxos" $
+      checkDistr utxos expectedAmounts
+  where
+  -- Idea here is to iterate through the expected amounts and remove
+  -- one matching utxo from the utxo set if found, otherwise return
+  -- false. Once we've gone through all expected amounts, if all of
+  -- them have been found in the utxo set, we expect there to be no
+  -- utxos remaining
+  checkDistr :: Utxo -> InitialUTxO -> Boolean
```
klntsky, line 526: Could you please also rename `InitialUTxO` to `InitialUTxOs`. I think this is just a mistake on my side.

klntsky (changes requested):

src/Plutip/UtxoDistribution.purs
```diff
@@ -0,0 +1,175 @@
...
+-- | distribution for wallets.
+-- | Number of wallets in distribution specification matches the number of
+-- | wallets provided to the user.
+class UtxoDistribution distr wallets | distr -> wallets where
+  encodeDistribution :: distr -> Array (Array UtxoAmount)
+  decodeWallets :: distr -> Array PrivateKeyResponse -> Maybe wallets
+  keyWallets :: Proxy distr -> wallets -> Array KeyWallet
+
+instance UtxoDistribution Unit Unit where
+  encodeDistribution _ = []
```
rynoV, line 66: @klntsky Should this line be this instead:
```suggestion
  encodeDistribution _ = [[]]
```
Currently when the `Unit` instance is used it can break things because a `PrivateKeyResponse` is not returned for `unit`. For example `distr = unit /\ aliceUtxos` will only request one private key, and this messes with the wallet decoding. The change above fixes that though

rynoV: I'm just unsure what the purpose of the unit instance was. I assume it was to represent an empty wallet

rynoV: Actually no that doesn't make sense because the wallet is also `Unit`

klntsky: The idea is to not ask for any wallet and not use it, in cases when we are just using queries.

rynoV: Would it be okay to make this change and just ignore the returned private key? Because currently we're able to keep the recursion simple by assuming there will be a private key for each distribution, but the assumption is violated by this instance. I think the alternatives would be quite messy

klntsky: I don't see a problem with it, the wallet will just be `unit`.

rynoV: Ended up going with a different solution the fixed this as well as another bug, see https://github.com/Plutonomicon/cardano-transaction-lib/pull/838/commits/bea82b5772e432d6847eb9a95640c107854cd38c

src/Plutip/UtxoDistribution.purs
```diff
@@ -0,0 +1,184 @@
...
+  , privateKeysToKeyWallet
+  )
+
+-- | A type class that implements a type-safe interface for specifying UTXO
+-- | distribution for wallets.
+-- | Number of wallets in distribution specification matches the number of
+-- | wallets provided to the user.
+class UtxoDistribution distr wallets | distr -> wallets where
+  encodeDistribution :: distr -> Array (Array UtxoAmount)
+  decodeWallets :: distr -> Array PrivateKeyResponse -> Maybe wallets
```
rynoV, line 63: Should probably add a check in the `decodeWallets` definitions that the final `PrivateKeyResponse` array is empty

test/Plutip.purs
```diff
@@ -388,6 +402,64 @@ suite = do
           logInfo' "Try to spend locked values"
           AlwaysSucceeds.spendFromAlwaysSucceeds vhash validator txId
 
+-- TODO: use this function in the other tests
+checkUtxoDistribution
```
rynoV, line 406: This file is starting to get big, might be a good idea to break it up

test/Plutip.purs
```diff
@@ -331,35 +362,27 @@ suite = do
...
-          ubTx1 <- liftedE $ Lookups.mkUnbalancedTx lookups constraints
-          ubTx2 <- liftedE $ Lookups.mkUnbalancedTx lookups constraints
-
-          withBalancedAndSignedTxs [ ubTx1, ubTx2 ] $ \txs -> do
-            locked <- getLockedInputs
-            logInfo' $ "Locked inputs inside bracket (should be nonempty): " <>
-              show
-                locked
-            traverse_ submitAndLog txs
+    -- TODO: not sure what's causing this to fail
```
rynoV, line 369: From slack

> (me) I added [this test](https://github.com/Plutonomicon/cardano-transaction-lib/blob/c3ad6a7c648c791c2adcefbdc0ea21731d9816b1/test/Plutip.purs#L358) for running the sign multiple contract with a stake key, but it ends up failing [here](https://github.com/Plutonomicon/cardano-transaction-lib/blob/c3ad6a7c648c791c2adcefbdc0ea21731d9816b1/test/Plutip.purs#L417) with "locked inputs map is not empty", while the test above it succeeds and the only difference is it doesn't use a stake key. I'm not sure why using a staking key is causing this, it seems unrelated to me, do you have any ideas?

> (@klntsky) So these locked inputs remain from the transactions that distribute balances, meaning our assertion is wrong.
I think we can be sure we are not going to try spending these already spent UTxOs, so feel free to clear locked inputs Ref manually  after sending the tx that distributes funds. See getLockedInputs - we want to set cache to mempty

rynoV: @klntsky If the assertion is wrong and the cache is as expected, could I just remove the assertion?

klntsky: @rynoV 

I think we should not remove the assertion and instead clear locked inputs. Because distribution of funds is transparent to the user, and since we await the tx, we can be sure the state of these utxos is settled. This locked inputs mechanism is designed to prevent double spending attempts in some scenarios. But there is nothing we should guard against in this case.

test/Plutip/UtxoDistribution.purs
```diff
@@ -0,0 +1,214 @@
...
+  -> wallet
+  -> Contract r Unit
+checkUtxoDistribution distr wallets = do
+  let
+    walletsArray = keyWallets (Proxy :: Proxy distr) wallets
+    walletUtxos = encodeDistribution distr
+  for_ walletsArray assertUtxosAtPlutipWalletAddress
+  assertCorrectDistribution $ zip walletsArray walletUtxos
+
+-- TODO: minimum value of 1 ada is hardcoded, tests become flaky below
```
klntsky, line 88: We have a policy that each TODO should link to an open issue, Could you please add a link to https://github.com/Plutonomicon/cardano-transaction-lib/issues/857 in the comments?

test/Plutip/UtxoDistribution.purs
```diff
@@ -0,0 +1,214 @@
...
+        ]
+
+data ArbitraryUtxoDistr
+  = UDUnit
+  | UDInitialUtxos InitialUTxOs
+  | UDInitialUtxosWithStake InitialUTxOsWithStakeKey
+  | UDTuple ArbitraryUtxoDistr ArbitraryUtxoDistr
+
+instance Show ArbitraryUtxoDistr where
+  show = case _ of
```
klntsky, line 116: We want `Show` instances to be as close to code as possible. Please include literal constructor names in the strings. If you want to pretty-print the type, use a standalone function.

test/Plutip/UtxoDistribution.purs
```diff
@@ -0,0 +1,214 @@
...
+  assertCorrectDistribution $ zip walletsArray walletUtxos
+
+-- TODO: minimum value of 1 ada is hardcoded, tests become flaky below
+-- that value. Ideally this shouldn't be hardcoded
+genInitialUtxo :: Gen InitialUTxOs
+genInitialUtxo = map (BigInt.fromInt >>> (_ * BigInt.fromInt 1_000_000))
+  <$> arrayOf (chooseInt 1 1000)
+
+instance Arbitrary ArbitraryUtxoDistr where
+  arbitrary = fix \_ -> frequency <<< wrap $
```
klntsky, line 95: Please make this instance respect `size` parameter that is implicitly available in `Gen` monad. The simplest way is to use `Test.QuickCheck.Gen.sized` and `resize`.

test/Plutip/UtxoDistribution.purs
```diff
@@ -0,0 +1,214 @@
...
+            ( UDInitialUtxosWithStake <$>
+                ( InitialUTxOsWithStakeKey
+                    <$> (pure privateStakeKey)
+                    <*> genInitialUtxo
+                )
+            )
+        , 4.0 /\ (UDTuple <$> arbitrary <*> arbitrary)
+        ]
+
+data ArbitraryUtxoDistr
```
klntsky, line 109: Should this type be exposed to the users in `Contract.Test.Plutip` namespace?

klntsky (changes requested):

test/Plutip/UtxoDistribution.purs
rynoV: I guess it could be useful if someone wanted to do some kind of property testing of their contracts, like maybe checking that it never throws an exception.

There's a couple others that I felt could be in the testing interface, like `assertContract`, `assertNoUtxosAtAddress`, `assertNoUtxosAtEnterpriseAddress`

rynoV: There's also the fact that it uses the same staking key for every stake key wallet, I'm not sure if that's something that should be put in the public interface. Also not really sure how easy it would be to generate a random staking key

klntsky: I agree, let's keep it private

src/Plutip/Server.purs
```diff
@@ -143,12 +157,24 @@ withPlutipContractEnv plutipCfg distr cont = do
...
-    Just wallets -> cc wallets
+  withWallets
+    :: ContractEnv ()
+    -> PrivatePaymentKey
+    -> ClusterStartupParameters
+    -> (wallets -> Aff a)
+    -> Aff a
+  withWallets env ourKey response cc = do
+    wallets <- runContractInEnv
+      (over wrap (_ { config { customLogger = Just (const $ pure unit) } }) env)
```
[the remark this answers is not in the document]
klntsky: Why do you think so? I tried logging a string and it suppressed it.

klntsky (approved):

src/Plutip/Server.purs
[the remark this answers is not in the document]
rynoV: I still see the debug ogmios request/response logs from this contract in the tests. Maybe those can't be suppressed?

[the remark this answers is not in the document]
klntsky: I see, we get protocol parameters before we fully initialize the contract. That's why the logger is not the same, see
https://github.com/Plutonomicon/cardano-transaction-lib/blob/37dd169e9f060af09de259a2891fa4a4f2a3cfe6/src/QueryM.purs#L676

It's our problem, let's not block on that.

https://github.com/Plutonomicon/cardano-transaction-lib/issues/897

klntsky (changes requested):
>  ✓︎ runPlutipContract: stake key transfers with random distribution: ((((stake + [] /\ []) /\ [fromString "120000000"]) /\ (((([] /\ (stake + [] /\ [])) /\ ((stake + [] /\ stake + []) /\ ([] /\ []))) /\ [fromString "223000000"]) /\ [fromString "635000000",fromString "720000000"])) /\ [fromString "415000000",fromString "911000000",fromString "275000000",fromString "694000000"])


Can we remove `fromString` and just print the number?

errfrom (approved):

klntsky (approved):

PR #817, opened 2022-08-01
Various Nix things

ngua:
Closes #760, closes #769, closes #762, closes #796, closes #816, closes #445, closes #840

### Pre-review checklist

- [x] All code has been formatted using our config (`make format` for Purescript, `nixpkgs-fmt` for Nix)
- [ ] All Purescript imports are explicit, including constructors
- [ ] The integration and unit tests have been run (`npm run test`) locally
- [x] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

PR #854, opened 2022-08-05
Make sure `haskell-nix.overlay` has been applied

ngua:
Fixes an issue introduced in #817. In order to avoid instantiating our own `nixpkgs` (and using our own version of haskell.nix, which can cause painful issues for users), we add `ctl-server` to `overlays.runtime` using the `hsProjectFor` function. If users aren't using haskell.nix themselves, `haskell-nix` will be missing from the package set and `ctl-server` can't be built. This fixes it by checking if that attribute is present and if not applying the overlay

I also fixed the template `welcomeText` so it gets formatted correctly

### DIFF
--- a/templates/ctl-scaffold/README.md
+++ b/templates/ctl-scaffold/README.md
@@ -4,10 +4,14 @@ Welcome to your new CTL project!
 
 To enter the Nix environment and start working on it, run `nix develop`
 
-Please also see
+Please also see our
 
-- Our documentation at https://github.com/Plutonomicon/cardano-transaction-lib/tree/develop/doc
-- Our Pursuit docs at https://plutonomicon.github.io/cardano-transaction-lib/
-- Our Discord server https://discord.gg/c8kZWxzJ
+- [Documentation](https://github.com/Plutonomicon/cardano-transaction-lib/tree/develop/doc)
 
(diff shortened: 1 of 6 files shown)


ngua:
We also need to check for `libsodium-vrf` I guess

templates/ctl-scaffold/README.md
```diff
@@ -4,10 +4,14 @@ Welcome to your new CTL project!
...
-Please also see
+Please also see our
 
-- Our documentation at https://github.com/Plutonomicon/cardano-transaction-lib/tree/develop/doc
-- Our Pursuit docs at https://plutonomicon.github.io/cardano-transaction-lib/
-- Our Discord server https://discord.gg/c8kZWxzJ
+- [Documentation](https://github.com/Plutonomicon/cardano-transaction-lib/tree/develop/doc)
 
-If you encounter problems and/or want to report a bug, you can open an issue at https://github.com/Plutonomicon/cardano-transaction-lib/issues. Please search for existing issues before!
+- [Pursuit docs](https://plutonomicon.github.io/cardano-transaction-lib/)
```
klntsky, line 11:
```suggestion
- [Generated docs](https://plutonomicon.github.io/cardano-transaction-lib/)
```

Pursuit is not used there in any way.

klntsky (dismissed):

PR #859, opened 2022-08-05
make plutip use WalletSpec instead of KeyWallet

Geometer1729:
Closes # .

Downstream we want to test our CLI with plutip. The cli reads wallets from files so we need plutip to provide wallets in a type that's still possible to serialize into a file.

### Pre-review checklist


klntsky (changes requested):
Add a method to get private keys from KeyWallet instead of breaking the API. I was also thinking about exposing public keys as well, to not call `publicKeyFromPrivateKey` each time we ask for a wallet address.

PR #866, opened 2022-08-09
Fee calculation using CSL

### LINKED ISSUE #851: Switch to CSL for fee calculation

klntsky:
Closes #851

### Pre-review checklist


CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - The [`ctl-scaffold`](https://github.com/mlabs-haskell/ctl-scaffold) repository has been archived and deprecated and its contents moved to `templates.ctl-scaffold` in the CTL flake ([#760](https://github.com/Plutonomicon/cardano-transaction-lib/issues/760)).
 - The CTL `overlay` output has been deprecated and replaced by `overlays.purescript` and `overlays.runtime` ([#796](https://github.com/Plutonomicon/cardano-transaction-lib/issues/796)).
 - `buildCtlRuntime` and `launchCtlRuntime` now take an `extraServices` argument to add `docker-compose` services to the resulting Arion expression ([#769](https://github.com/Plutonomicon/cardano-transaction-lib/issues/769)).
+- Use `cardano-serialization-lib` for fee calculation, instead of server-side code.
```
ngua, line 71: I think the fee endpoint should go under `### Removed`

ngua (commented):
Very happy to see this change!

The docs should be updated to reflect that CTL server is only required now to use `applyArgs`

There are some other things we should do as well I think, but we can leave them for other PRs:
- `runPlutipTest` should accept a flag if CTL server should be included
- Spawning the services for Plutip should be configurable in some way (I'm not sure the best way to do this though, environment variables?)
- `ctl-server` should be removed from `overlays.runtime` and put into its own overlay

jy14898:
Should we update https://github.com/CardanoSolutions/ogmios/issues/225 about our intentions? Specifically, CTL would no longer benefit (or would it?) from having Ogmios estimate fees. Maybe just the following isn't true

> For the sever, yes. Either I or someone from the CTL team could work on that part.

CHANGELOG.md
klntsky: I disagree, It's not a part of our public API, only we ourselves use it. No point in bothering users with this hidden change.

jy14898 (approved):

ngua (approved):

PR #869, opened 2022-08-10
Add NativeScripts

### LINKED ISSUE #791: Support native scripts
Time and signature based NativeScript validators are currently supported in a strict sense - we are able to create arbitrary transactions. But it's not very convenient for the users.

We need to:

- Expose `NativeScript.hash` as part of our public API
- Create an example that locks and spends funds
- See if we can alter our Constraints interface for even more convenience (we are free to diverge from Plutus here)

klntsky:
Basing this on #866 because it only works with CSL fee calculation.

Closes #791

### Pre-review checklist


src/Scripts.purs
```diff
@@ -82,6 +84,10 @@ validatorHashEnterpriseAddress network valHash =
       , paymentCred: scriptHashCredential (unwrap valHash)
       }
 
+nativeScriptHashEnterpriseAddress :: NetworkId -> NativeScriptHash -> Address
```
jy14898, line 87: (basically something similar to what is done above)
```suggestion
-- | Converts a Plutus-style `NativeScriptHash` to an `Address` as an
-- | `EnterpriseAddress`.
nativeScriptHashEnterpriseAddress :: NetworkId -> NativeScriptHash -> Address
```

src/Transaction.purs
```diff
@@ -130,6 +132,12 @@ attachPlutusScripts ps tx@(Transaction { witnessSet: ws }) = do
   updateTxWithWitnesses tx
     =<< convertWitnessesWith ws (Serialization.WitnessSet.setPlutusScripts ps')
 
+attachNativeScript
```
jy14898, line 135: Once again docs can be straightforwardly added as in my previous comment

test/Plutip.purs
```diff
@@ -251,6 +269,103 @@ suite = do
...
+              , BigInt.fromInt 2_000_000_000
+              ]
+      runPlutipContract config distribution \(alice /\ bob /\ charlie /\ dan) ->
+        do
+          alicePaymentPKH <- liftedM "Unable to get Alice's PKH" $
+            coerce <$> withKeyWallet alice ownPaymentPubKeyHash
+          bobPaymentPKH <- liftedM "Unable to get Bob's PKH" $
+            coerce <$> withKeyWallet bob ownPaymentPubKeyHash
+          charliePaymentPKH <- liftedM "Unable to get Charlie's PKH" $
+            map (unwrap <<< unwrap) <$> withKeyWallet charlie
```
jy14898, line 298: Can be coerced like the others

jy14898 (commented):
Looking good AFAICT

ngua (changes requested):
Can you also update the documentation to note the differences between Plutus and CTL wrt native scripts please?

ngua (commented):
Should we add an example that can be run in the browser as well?

src/Scripts.purs
klntsky: There is no plutus-style NativeScriptHash, because plutus does not deal with native scripts.

src/Transaction.purs
klntsky: This function is internal and the purpose is obvious

klntsky:
Converting to draft, because without https://github.com/Plutonomicon/cardano-transaction-lib/issues/871 this API is mostly unusable.

klntsky:
> Should we add an example that can be run in the browser as well?

We need multiple parties. Too hard to implement with wallets.

src/NativeScripts.purs
```diff
@@ -0,0 +1,91 @@
...
+
+subsetsOfLength :: forall a. Int -> Array a -> Array (Array a)
+subsetsOfLength n =
+  List.fromFoldable >>> sublists n >>> List.toUnfoldable >>> map
+    List.toUnfoldable
+
+sublists :: forall a. Int -> List a -> List (List a)
+sublists n xs = List.take (List.length xs - n + 1) $ sublists' n xs
+  where
+  sublists' _ Nil = Cons Nil Nil
```
ngua, line 63: I know it's obvious, but for the same of consistency please add a type signature

src/NativeScripts.purs
```diff
@@ -0,0 +1,91 @@
...
+instance Show NativeScriptHash where
+  show (NativeScriptHash sh) = "(NativeScriptHash " <> show sh <> ")"
+
+nativeScriptHash :: NativeScript -> Maybe NativeScriptHash
+nativeScriptHash ns = wrap <<< Hashing.nativeScriptHash <$> convertNativeScript
+  ns
+
+-- | `SetChoice` is an internal type representing internal state of
+-- | `getMaximumSigners` algorithm.
+type SetChoice a = Array (Set a)
```
ngua, line 47: Missing kind signatures here and several places below

ngua (changes requested):

ngua (approved):

Issue #398, opened 2022-05-10
Fix error recovery mode for SubmitTx


klntsky:
As part of #233 we implemented a simple approach to error recovery with the use of `reconnecting-websocket`. However, there's one special endpoint for which simple replaying is not going to work. We need to check if a transaction has been sent before resubmitting it (otherwise UTXOs may be consumed by TX and we'll get an error when trying to send it for the second time).

How to address this problem: maintain request queue as a list of transactions instead of a list of `String` payloads. Before retrying, use https://ogmios.dev/mini-protocols/local-tx-monitor/ to see if a transaction is in local node mempool.

klntsky:
The reason why we are delaying this: the chance to observe the error is low.

PR #870, opened 2022-08-10
Use Ogmios Tx Monitor to improve error recovery logic for SubmitTx

errfrom:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/398

### Pre-review checklist


src/QueryM.purs
```diff
@@ -652,29 +744,53 @@ mkOgmiosWebSocket' lvl serverCfg continue = do
...
+          label <> ": " <> show value <> " TxHash: " <> show txHash
+    withMempoolSnapshot ogmiosWs lvl case _ of
+      Nothing ->
+        liftEffect $ traverse_ sendRequest submitPendingRequests
+      Just ms -> do
+        let (pr' :: Array _) = Map.toUnfoldable submitPendingRequests
+        for_ pr' \(listenerId /\ requestBody /\ txHash /\ _) -> do
+          txInMempool <- mempoolSnapshotHasTxAff ogmiosWs lvl ms txHash
+          logger "Tx in the mempool" txInMempool txHash
+          retrySubmitTxRef <- liftEffect $ Ref.new false
```
firefrorefiddle, line 779: That looks very imperative.

firefrorefiddle (commented):
Looks good to me.

test/AffInterface.purs
```diff
@@ -145,8 +145,9 @@ testGetEraSummaries = do
 
 testSubmitTxFailure :: Aff Unit
 testSubmitTxFailure = do
+  let bytes = hexToByteArrayUnsafe "00"
   runQueryM testnetTraceQueryConfig
-    (void $ submitTxOgmios (wrap $ hexToByteArrayUnsafe "00"))
+    (void $ submitTxOgmios bytes (wrap bytes))
```
jy14898, line 150: Is it just a coincidence that the hash of `"00"` is itself? Reads like a bug at first. Maybe a comment is necessary. Or does it not matter given it's a failure case?

errfrom: @jy14898 It does not matter what value we provide for the tx hash in this test, since the transaction itself is invalid

src/QueryM/Ogmios.purs
```diff
@@ -211,39 +208,79 @@ queryUtxoCall = mkOgmiosCallType
...
-  , version: "1.0"
-  , servicename: "ogmios"
-  }
+mkOgmiosCallType = flip
+  ( mkCallType
+      { "type": "jsonwsp/request"
+      , version: "1.0"
+      , servicename: "ogmios"
+      }
+  )
```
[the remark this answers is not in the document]
errfrom: @jy14898 Hm why do you think that using `flip` here is ugly?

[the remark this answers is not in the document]
jy14898: Just the parenthesis actually, but you could use `$`. Not super important either way

src/QueryM.purs
```diff
@@ -652,29 +748,76 @@ mkOgmiosWebSocket' lvl serverCfg continue = do
...
+    logger "Tx in the mempool" txInMempool txHash
+    retrySubmitTxRef <- liftEffect $ Ref.new false
+    unless txInMempool do
+      -- Check if the transaction was included in the block:
+      txConfirmed <- checkTxByHashAff datumCacheWs lvl txHash
+      logger "Tx confirmed" txConfirmed txHash
+      unless txConfirmed $ liftEffect do
+        Ref.write true retrySubmitTxRef
+        sendRequest (requestBody /\ unit)
+    retrySubmitTx <- liftEffect $ Ref.read retrySubmitTxRef
```
jy14898, line 807: A suggestion for @firefrorefiddle's criticism 
```suggestion
    retrySubmitTx <-
      if txInMempool then do
        -- Check if the transaction was included in the block:
        txConfirmed <- checkTxByHashAff datumCacheWs lvl txHash
        logger "Tx confirmed" txConfirmed txHash
        unless txConfirmed $ liftEffect do
          sendRequest (requestBody /\ unit)
        pure (not txConfirmed)
      else pure false
```

jy14898: Actually it's probably better to use `MonadPlus`'s guard

errfrom: @jy14898 Thanks for this suggestion!

jy14898:
@errfrom Is it possible to write a test for this, one that kills the websocket?

CHANGELOG.md
```diff
@@ -89,6 +89,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Bug with collateral selection for `KeyWallet` when signing multiple transactions ([#709](https://github.com/Plutonomicon/cardano-transaction-lib/pull/709))
 - Bug when zero-valued non-Ada assets were added to the non-Ada change output ([#802](https://github.com/Plutonomicon/cardano-transaction-lib/pull/802))
 - Properly implemented CIP-25 V2 metadata. Now there's no need to split arbitrary-length strings manually to fit them in 64 PlutusData bytes (CTL handles that). A new `Cip25String` type has been introduced (a smart constructor ensures that byte representation fits 64 bytes, as required by the spec). Additionally, a new `Metadata.Cip25.Common.Cip25TokenName` wrapper over `TokenName` is added to ensure proper encoding of `asset_name`s. There are still some minor differences from the spec:
+- Error recovery logic for `SubmitTx` if the ws connection is dropped ([#870](https://github.com/Plutonomicon/cardano-transaction-lib/pull/870))
```
klntsky, line 92:
```suggestion
- Error recovery logic for `SubmitTx` if the WebSocket connection is dropped ([#870](https://github.com/Plutonomicon/cardano-transaction-lib/pull/870))
```

errfrom:
@jy14898 Currently it's possible to test these changes by uncommenting the following lines:
https://github.com/Plutonomicon/cardano-transaction-lib/blob/031ce4a201327f169cabb7630eb028a9c2a8f257/src/QueryM.purs#L1089-L1091

However, I have not yet figured out how to approach writing tests for this

klntsky (changes requested):
How is it going to work with multiple users of the same Ogmios instance? The state is global, right?

errfrom:
@klntsky It should work just fine, because the `_onWsConnect` handler will only be invoked once in case of a failure, so the code responsible for resending `SubmitTx` requests will work with a single mempool snapshot

errfrom:
@klntsky If you are referring to the case where multiple users would connect to Ogmios and use its Tx Monitor protocol, so for each user there will be a separate internal state maintained by Ogmios (AFAIK)

klntsky:
> so for each user there will be a separate internal state maintained by Ogmios (AFAIK)

What is the user identity Ogmios uses to tie the state? IP?

errfrom:
@klntsky IP + port number. For each new connection, Ogmios initializes all services and handles the state separately. This is what I understood from reading the source code.
https://github.com/CardanoSolutions/ogmios/blob/a33342ba76b8424b7c2ff67f10e9a09bf81a9ebe/server/src/Ogmios/App/Server/WebSocket.hs#L154-L164

klntsky:
What can happen at worst in case there are multiple browser tabs (that share the state)?

CHANGELOG.md
```diff
@@ -89,6 +89,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Bug with collateral selection for `KeyWallet` when signing multiple transactions ([#709](https://github.com/Plutonomicon/cardano-transaction-lib/pull/709))
 - Bug when zero-valued non-Ada assets were added to the non-Ada change output ([#802](https://github.com/Plutonomicon/cardano-transaction-lib/pull/802))
 - Properly implemented CIP-25 V2 metadata. Now there's no need to split arbitrary-length strings manually to fit them in 64 PlutusData bytes (CTL handles that). A new `Cip25String` type has been introduced (a smart constructor ensures that byte representation fits 64 bytes, as required by the spec). Additionally, a new `Metadata.Cip25.Common.Cip25TokenName` wrapper over `TokenName` is added to ensure proper encoding of `asset_name`s. There are still some minor differences from the spec:
+- Error recovery logic for `SubmitTx` if the WebSocket connection is dropped ([#870](https://github.com/Plutonomicon/cardano-transaction-lib/pull/870))
```
ngua, line 92: This needs to be moved, the nested list below here belongs to the previous item (CIP-25 changes)

src/Hashing.purs
```diff
@@ -54,3 +58,9 @@ datumHash =
 plutusScriptHash :: PlutusScript -> Aff (Maybe ScriptHash)
 plutusScriptHash =
   map (scriptHashFromBytes <<< wrap) <<< Promise.toAffE <<< hashPlutusScript
+
+-- | Calculates the hash of the transaction by applying `blake2b256Hash` to
+-- | the cbor-encoded transaction body.
+transactionHash :: Serialization.Transaction -> Aff TransactionHash
```
ngua, line 64: I think there are a few other places where we've used `toBytes <<< asOneOf <<< ...` for `Transaction`s before. Maybe we can replace those now?

CHANGELOG.md
```diff
@@ -89,6 +89,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
```
ngua, line 93: It might be worth it to note that `Contract.Hashing.transactionHash` is now available. That was a blind spot in the API

src/Hashing.purs
errfrom: There doesn't seem to be any other places in the codebase where we need transaction hash calculation

errfrom:
@klntsky In the worst case scenario (when two clients attempt to resend their pending submit requests), one of them may potentially be blocked on `AwaitAcquire` request until a new mempool snapshot is available, but the odds of this happening are quite low.

klntsky (approved):

PR #888, opened 2022-08-11
Calculate NativeScript maximum required signers number automatically

### LINKED ISSUE #871: Calculate number of required signers given a NativeScript
With Native, or Multisignature Scripts introduced in #869, it is impossible to know which pubkey owners will add their signatures to the final Tx, for example, in case of `NOfK`-type multisig script. We should add calculation of the maximum possible number to CTL (given a particular script).

klntsky:
Based on #869 

Closes #871

### Pre-review checklist

- [ ] ~~The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included~~ No visible changes

### DIFF
--- a/src/NativeScripts.purs
+++ b/src/NativeScripts.purs
@@ -50,10 +50,7 @@ anyChoice :: forall a. Ord a => SetChoice a -> SetChoice a -> SetChoice a
 anyChoice as bs = Array.nub $ as <> bs
 
 allChoice :: forall a. Ord a => SetChoice a -> SetChoice a -> SetChoice a
-allChoice as bs = do
-  a <- as
-  b <- bs
-  pure $ a <> b
+allChoice as bs = (<>) <$> as <*> bs
 
 subsetsOfLength :: forall a. Int -> Array a -> Array (Array a)
 subsetsOfLength n =
--- a/src/NativeScripts.purs
+++ b/src/NativeScripts.purs
@@ -44,7 +44,7 @@ nativeScriptHash ns = wrap <<< Hashing.nativeScriptHash <$> convertNativeScript
 
 -- | `SetChoice` is an internal type representing internal state of
 -- | `getMaximumSigners` algorithm.
-type SetChoice a = Array { set :: Set a, size :: Int }
+type SetChoice a = Array (Set a)
 
 anyChoice :: forall a. Ord a => SetChoice a -> SetChoice a -> SetChoice a
 anyChoice as bs = Array.nub $ as <> bs
@@ -53,9 +53,7 @@ allChoice :: forall a. Ord a => SetChoice a -> SetChoice a -> SetChoice a
 allChoice as bs = do
   a <- as
   b <- bs
-  let
-    set = a.set <> b.set
-  pure { size: Set.size set, set }
+  pure $ a <> b
 
 subsetsOfLength :: forall a. Int -> Array a -> Array (Array a)
 subsetsOfLength n =
@@ -76,13 +74,13 @@ sublists n xs = List.take (List.length xs - n + 1) $ sublists' n xs
 -- | of already known signers to be ignored in this function.
 getMaximumSigners :: Set Ed25519KeyHash -> NativeScript -> Int
 getMaximumSigners alreadyCounted =
-  sizes >>> maximumBy (compare `on` _.size) >>> map _.size >>> fromMaybe 0
+  sizes >>> maximumBy (compare `on` Set.size) >>> map Set.size >>> fromMaybe 0
   where
   sizes :: NativeScript -> SetChoice Ed25519KeyHash
   sizes = case _ of
     ScriptPubkey kh
       | Set.member kh alreadyCounted -> emptySetChoice
-      | otherwise -> [ { size: 1, set: Set.singleton kh } ]
+      | otherwise -> [ Set.singleton kh ]
     ScriptAll nss -> foldr allChoice emptySetChoice
       (sizes <$> nss)
     ScriptAny nss -> foldr anyChoice emptySetChoice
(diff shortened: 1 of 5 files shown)


src/NativeScripts.purs
```diff
@@ -21,3 +41,66 @@ instance Show NativeScriptHash where
...
+-- | twice, because that would lead to excessive fees. Hence we accept a set
+-- | of already known signers to be ignored in this function.
+getMaximumSigners :: Set Ed25519KeyHash -> NativeScript -> Int
+getMaximumSigners alreadyCounted =
+  choices >>> maximumBy (compare `on` _.choice) >>> map _.choice >>> fromMaybe 0
+  where
+  choices :: NativeScript -> SetChoice Ed25519KeyHash
+  choices = case _ of
+    ScriptPubkey kh
+      | Set.member kh alreadyCounted -> [ { choice: 0, set: Set.empty } ]
```
firefrorefiddle, line 97: Why not give a name to `{ choice: 0, set: Set.empty }`?

src/NativeScripts.purs
```diff
@@ -21,3 +41,66 @@ instance Show NativeScriptHash where
 nativeScriptHash :: NativeScript -> Maybe NativeScriptHash
 nativeScriptHash ns = wrap <<< Hashing.nativeScriptHash <$> convertNativeScript
   ns
+
+-- | `SetChoice` is an internal type representing internal state of
+-- | `getMaximumSigners` algorithm. There are a number of values (key hashes)
+-- | from which we must choose exactly `choice`. It's like a suspended `NOfK`
+-- | choice. We track the alternatives alongside their count to adjust for
+-- | possible duplicates in `allChoice`.
+type SetChoice a = Array { set :: Set a, choice :: Int }
```
firefrorefiddle, line 50: Isn't `choice` always equal to `Set.size set`?

firefrorefiddle: Because the only instance where you create a `{ set :: Set a, choice :: Int }` other then the empty and singleton variants is in `allChoice`, where you take care to maintain this invariant.

src/Serialization/MinFee.purs
```diff
@@ -44,16 +48,34 @@ calculateMinFeeCsl (ProtocolParameters pparams) txNoSigs = do
...
+    -- We want to add space for required signatures (at least one, if
+    -- none specified).
+    nRequiredSigners = tx # unwrap >>> _.body >>> unwrap >>> _.requiredSigners
+      >>> map (map unwrap >>> Array.length)
+      >>> fromMaybe 1
+
   in
     tx # _witnessSet <<< _vkeys .~ Just
-      (Array.replicate nRequiredSigners fakeVkeywitness)
+      (Array.replicate (nRequiredSigners + nsPossibleSigners) fakeVkeywitness)
```
firefrorefiddle, line 78: If `requiredSigners` is empty, but `nsPossibleSigners` isn't, will this count one too many, owing to the `fromMaybe 1` above? Is this even possible?

src/NativeScripts.purs
klntsky: Initially I had another algorithm in mind, but now I realize that the choice is really the same as `size`. I will update the comment. But I'll keep the `choice` around, because it is `O(1)`.

src/Serialization/MinFee.purs
klntsky: It depends on what's inside the script. We need to add at least one witness, but if we don't know the expected signer, we can't know if it is used in the script. So it's the best we can do.

src/NativeScripts.purs
```diff
@@ -21,3 +41,66 @@ instance Show NativeScriptHash where
...
+  b <- bs
+  pure
+    -- We want to select a.choice from a.set and b.choice from b.set,
+    -- but there are equal elements that count for both if we select them.
+    -- So to satisfy the requirement we must uncount the elements present in
+    -- both sets.
+    [ { choice: a.choice + b.choice - Set.size (Set.intersection a.set b.set)
+      , set: a.set <> b.set
+      }
+    ]
```
jy14898: I'm not sure about performance, but `Array.concat` is just `Bind.join`, thus you can just drop `Array.concat` and the final `pure`

`join (m >>= f >>> pure) ≡ m >>= f`

src/NativeScripts.purs
jy14898: You only observe `choice` (other than for creating new `choice`) right at the end. You could just calculate the sizes at the end in `getMaximumSigners` (before `maximumBy`, if you care about the cost)

jy14898: Something like:

```haskell
type SetChoice a = Array (Set a)
...
choices >>> map (\set -> { set, size: size set }) >>> maximumBy (compare `on` _.size) >>> map _.size >>> fromMaybe 0
```

klntsky: I agree.

src/NativeScripts.purs
```diff
@@ -21,3 +41,54 @@ instance Show NativeScriptHash where
...
+type SetChoice a = Array (Set a)
+
+anyChoice :: forall a. Ord a => SetChoice a -> SetChoice a -> SetChoice a
+anyChoice as bs = Array.nub $ as <> bs
+
+allChoice :: forall a. Ord a => SetChoice a -> SetChoice a -> SetChoice a
+allChoice as bs = do
+  a <- as
+  b <- bs
+  pure $ a <> b
```
jy14898, line 56: Taking a look again, this is `zipWith (<>) as bs`, right?

jy14898, line 56: `(<>) <$> as <*> bs`?

jy14898 (approved):

PR #889, opened 2022-08-11
Wallet selection by URL

### LINKED ISSUE #833: Allow wallet & example selection in examples
It's tiresome to build examples one-by-one.

From slack:

> Can we make a command to build all examples at once in Nix, and then serve them in a storybook-like interface (a page that contains "tabs" that link to other pages, each containing one example).

This has been done by @firefrorefiddle in ByURL example.

We want to:

- ~~Make it a default example in `flake.nix`~~ DONE
- ~~Add links to `examples/index.hml` for each of the example URLs~~ DONE
- Add a drop-down button with a list of wallets we support. When we select from the list, the chosen wallet should be used to run the example. That is, before each example we should await an Aff action that returns a value when a selection is made by the user and dispatches the wallet. We should use `isNamiAvailable`/`isGeroAvailable`/`isEternlAvailable` to raise an error in case a wallet does not exist.
- Wallet selection should be performed via a URL parameter

klntsky:
Closes #833 

### Pre-review checklist


firefrorefiddle (commented):
That part looks very good, but the E2E tests need to be adjusted (I can do that).

firefrorefiddle:
> That part looks very good, but the E2E tests need to be adjusted (I can do that).

https://github.com/Plutonomicon/cardano-transaction-lib/pull/890

firefrorefiddle (approved):

PR #900, opened 2022-08-15
 Add integration tests to Plutip

### LINKED ISSUE #873: Add integration tests to plutip too
We need to run `Test.AffInterface` suite in plutip environment too.. We could have caught the recent time-related bug.

jy14898:
Closes #873.

### Pre-review checklist


jy14898:
@klntsky I've made it so `posixTimeToSlot`/`slotToPosixTime` runs for both unit tests (given a current era fixture) and for integration (via `getEraSummaries` etc). I pulled it out of `AffInterface` into `Test.Types.Interval` because it doesn't belong there, but as I said it still runs for integration tests.

So to be clear: `getEraSummaries` and `getSystemStart` are called in `Test.AffInterface`, `posixTimeToSlot` and `slotToPosixTime` are tested in `Test.Unit`, and the combination of `get*` and time conversions are tested in `Test.Integration`.

It's the best we can do until we are able to generate `posixTimeToSlot` etc tests from a given `eraSummaries` and `systemStart`, I'm not sure how the current fixtures were picked for testing so I was thinking to leaving that to a separate PR if someone with the know-how can pick it up more easily

test/AffInterface.purs
```diff
@@ -53,6 +41,8 @@ addr1 :: OgmiosAddress
 addr1 =
   "addr1qyc0kwu98x23ufhsxjgs5k3h7gktn8v5682qna5amwh2juguztcrc8hjay66es67ctn0jmr9plfmlw37je2s2px4xdssgvxerq"
 
+type TestPlanM a = MoteT (Const Void) (QueryM Unit) Aff a
```
ngua, line 44: Kind signature

test/Types/Interval.purs
```diff
@@ -0,0 +1,184 @@
...
+  a
+
+suite :: TestPlanM Unit
+suite = do
+  group "Interval type" do
+    test "Inverse posixTimeToSlot >>> slotToPosixTime " $ testPosixTimeToSlot
+    test "Inverse slotToPosixTime >>> posixTimeToSlot " $ testSlotToPosixTime
+    test "PosixTimeToSlot errors" $ testPosixTimeToSlotError
+
+-- launchAff_ $ runContract testnetConfig $ getEraSummaries >>= encodeAeson >>> stringifyAeson >>> logShow
```
ngua, line 43: There's a few commented-out lines like this

test/Types/Interval.purs
```diff
@@ -0,0 +1,184 @@
...
+import Test.Spec.Assertions (shouldEqual)
+import Types.BigNum (fromInt) as BigNum
+import Types.Interval
+  ( PosixTimeToSlotError(PosixTimeBeforeSystemStart)
+  , POSIXTime(POSIXTime)
+  , posixTimeToSlot
+  , slotToPosixTime
+  )
+
+type TestPlanM a = MoteT (Const Void)
```
ngua, line 31: Kind signature

test/AffInterface.purs
ngua, line 44: Also I think it's a bit confusing to have two `TestPlanM`s. Instead of having two (or more), why not parameterize `TestM.TestM` by some `m :: Type -> Type`?

ngua, line 44: Another thing that I'd prefer if possible: can we test through the `Contract` interface instead? I don't see a reason to use `QueryM` directly (I know these tests used to use `runQueryM`, but I think they actually predate our `Contract` interface)

test/AffInterface.purs
```diff
@@ -94,207 +81,65 @@ suite = do
...
-    -- curl localhost:9999/control/fetch_blocks -X POST -d '{"slot": 54066900, "id": "6eb2542a85f375d5fd6cbc1c768707b0e9fe8be85b7b1dd42a85017a70d2623d", "datumFilter": {"address": "addr_xyz"}}' -H 'Content-Type: application/json'
-    -- ```
-    _datums <- getDatumsByHashes $ pure $ DataHash $ hexToByteArrayUnsafe
-      "f7c47c65216f7057569111d962a74de807de57e79f7efa86b4e454d42c875e4e"
-    pure unit
+testOgmiosDatumCacheGetDatumsByHashes :: QueryM Unit
+testOgmiosDatumCacheGetDatumsByHashes = do
+  -- Use this to trigger block fetching in order to actually get the datum:
+  -- ```
+  -- curl localhost:9999/control/fetch_blocks -X POST -d '{"slot": 54066900, "id": "6eb2542a85f375d5fd6cbc1c768707b0e9fe8be85b7b1dd42a85017a70d2623d", "datumFilter": {"address": "addr_xyz"}}' -H 'Content-Type: application/json'
```
ngua, line 97: We don't need this comment anymore (I know you didn't add it, but now would be a good time to remove it)

test/AffInterface.purs
```diff
@@ -94,207 +81,65 @@ suite = do
...
+  -- ```
+  _datums <- getDatumsByHashes $ pure $ DataHash $ hexToByteArrayUnsafe
```
ngua, line 99: Again, I know this wasn't introduced in your commits but stylistically I'd prefer to use `void` here, then we don't need the `pure unit` either

ngua (changes requested):
I'm generally in favor of the direction of this PR and the changes you've introduced, but I think a few things need slight revision

test/Types/Interval.purs
jy14898: I left the comments as guides of how to generate new fixtures, when EraSummaries update/to indicate where it came from. I'll make the comment clearer

test/AffInterface.purs
jy14898: I've had a look into this, for the most part it'd be straight forward, but the submit tx failure case would be a little more verbose due to the input construction. Also, I'm not sure I agree anyway - the point of these tests are to test the websocket interface (maybe we should rename the module), `QueryM` being a fairly light wrapper around them.

jy14898: Hm, perhaps I should move

```purescript
  group "Address loop" do
    test "Ogmios Address to Address & back Testnet"
      $ testFromOgmiosAddress testnet_addr1
    test "Ogmios Address to Address & back non-Testnet"
      $ testFromOgmiosAddress addr1
```

elsewhere too, these don't even use `QueryM`

test/Types/Interval.purs
```diff
@@ -0,0 +1,179 @@
...
+    "[{\"end\":{\"epoch\":74,\"slot\":1598400,\"time\":31968000},\"parameters\":{\"epochLength\":21600,\"safeZone\":4320,\"slotLength\":20},\"start\":{\"epoch\":0,\"slot\":0,\"time\":0}},{\"end\":{\"epoch\":102,\"slot\":13694400,\"time\":44064000},\"parameters\":{\"epochLength\":432000,\"safeZone\":129600,\"slotLength\":1},\"start\":{\"epoch\":74,\"slot\":1598400,\"time\":31968000}},{\"end\":{\"epoch\":112,\"slot\":18014400,\"time\":48384000},\"parameters\":{\"epochLength\":432000,\"safeZone\":129600,\"slotLength\":1},\"start\":{\"epoch\":102,\"slot\":13694400,\"time\":44064000}},{\"end\":{\"epoch\":154,\"slot\":36158400,\"time\":66528000},\"parameters\":{\"epochLength\":432000,\"safeZone\":129600,\"slotLength\":1},\"start\":{\"epoch\":112,\"slot\":18014400,\"time\":48384000}},{\"end\":{\"epoch\":215,\"slot\":62510400,\"time\":92880000},\"parameters\":{\"epochLength\":432000,\"safeZone\":129600,\"slotLength\":1},\"start\":{\"epoch\":154,\"slot\":36158400,\"time\":66528000}},{\"end\":{\"epoch\":224,\"slot\":66398400,\"time\":96768000},\"parameters\":{\"epochLength\":432000,\"safeZone\":129600,\"slotLength\":1},\"start\":{\"epoch\":215,\"slot\":62510400,\"time\":92880000}}]"
```
ngua, line 42: This should go into our `fixtures` as an actual JSON file I think

test/Types/Interval.purs
```diff
@@ -0,0 +1,179 @@
...
+testPosixTimeToSlotError eraSummaries sysStart = do
+  let
+    posixTime = mkPosixTime "1000"
+  -- Some difficulty reproducing all the errors
+  errTest eraSummaries sysStart
+    posixTime
+    (PosixTimeBeforeSystemStart posixTime)
+  where
+  errTest
+    :: forall (err :: Type)
```
ngua, line 168: Unused type variable?

test/Types/Interval.purs
ngua: Again recommending that we have a small script somewhere that does this

ngua (changes requested):

ngua (commented):
One more thing that we can do: let's remove the checklist item for running the integration tests from the PR template (it's great that we can do that now!)

ngua (approved):

PR #901, opened 2022-08-15
Exit early if the UTxO value is less than 1 ADA (plutip)

### LINKED ISSUE #857: `runPlutipContract` stalls then errors with small utxo values
If a utxo distribution provides a value lower than ~800000 lovelace (rough estimate), runPlutipContract will stall then eventually error with ClientHttpError There was a problem making the request: request failed. Reproduce with:

```purescript
  let
    distribution =
        [ BigInt.fromInt 2_000_000_000
        , BigInt.fromInt 800000
        ]
  runPlutipContract config distribution \alice -> pure unit
```

This is caused by user error (I think?), but it should probably be handled better.

klntsky:
Closes #857

### Pre-review checklist


klntsky:
Note: the actual required value is less than 1 ADA (it depends on `coinsPerUtxoByte` and minimum possible Tx size. Bit it's hard to calculate, so let's keep it this way.

jy14898 (approved):

errfrom:
I think this is acceptable as a temporary solution, but it would be much better to adapt the `adaOnlyUtxoMinAdaValue` function from `BalanceTx.UtxoMinAda` for this purpose.

errfrom (approved):

klntsky:
@errfrom the actual minimum value is lower than `adaOnlyUtxoMinAdaValue` (which is 986990). E.g. it passes with 8900000. I didn't binsearch more.

PR #902, opened 2022-08-15
Make `ctl-server` optional

### LINKED ISSUE #872: Make ctl-server optional
We should allow to NOT start `ctl-server`. 

# Changes to Contract

`applyArgs` should fail on connection error with a meaningful error message, showing how to enable `ctl-server`

# Changes to Plutip

We should add a `PlutipConfig` parameter that controls `ctl-server` availability.

# Changes to nix

`ctl-server` should be removed from `overlays.runtime` and put into its own overlay

# Changes to docs

We should make it clear that `ctl-server` is optional. Show how to use it with changes in Nix.

ngua:
Closes #872 

Marking this as a draft since I thought we might want to discuss some of these changes:
- Is this approach to making `ctl-server` optional in the `ConfigParams` and `PlutipConfig` the best way to do it?
- Do we want `tesnetNamiConfig`, etc... to hold a `Just` value by default for `ctlServerConfig`?
- Do we want to remove `ctl-server` from `overlays.runtime` entirely? I've kept it for now with a deprecation notice for backwards compat, but the logging is pretty annoying
- Should `ctl-server.enable` be set to `true` by default in the runtime derivations?

I'm definitely open to hearing your thoughts about this. Considering most of our adopters use `applyArgs`, I didn't want things to suddenly break on them

Also fixes an issue with `mkServerUrl` (there was a missing separator for the `path` when `Just`) and updates `templates.ctl-scaffold`

### Pre-review checklist


ngua:
I'll quote myself from Slack to explain why I think `ctl-server` should be in a separate overlay:
> ... it's not worth having it around by default. If a project also uses haskell.nix there can be issues with hackage.nix having the wrong revision when another haskell.nix project is in the overlay. To get around this we need to build the server from the passed package set directly instead of using `self.packages...` directly (there were several PRs concerning this, it's affected at least one project). But that means we need to apply both `haskell-nix.overlay` and `iohk-nix.overlays.crypto` (another project had an issue when they hadn't applied these)

doc/runtime.md
```diff
@@ -64,6 +67,10 @@ Here is an example that uses the `runtime` overlay to launch all of the required
           # you probably want this one too, although it's not demonstrated
           # in this example
           cardano-transaction-lib.overlays.purescript
+          # This one is optional. If you set `ctl-server.enable = true;`
+          # in the runtime config as shown below, you must enable this
+          # overlay. `ctl-server` itself is **only** required when using
+          # CTL's `applyArgs` effect
```
jy14898, line 73: Sorry if I'm misunderstanding, but this is supposed to be followed by `cardano-transaction-lib.overlays.ctl-server` I think?

jy14898:
> - Is this approach to making `ctl-server` optional in the `ConfigParams` and `PlutipConfig` the best way to do it?

It's the best we can do without making `Contract` an extensible effects monad imo

jy14898 (commented):
Took a look and nothing sticks out to me as wrong or bad

doc/runtime.md
ngua: Good catch, you're right

src/QueryM.purs
```diff
@@ -498,29 +498,42 @@ applyArgs
...
+applyArgs script args = asks (_.ctlServerConfig <<< _.config) >>= case _ of
+  Nothing ->
+    pure
+      $ Left
+      $
+        ClientOtherError
+          "The `ctl-server` service is required to call `applyArgs`. Please \
+          \provide a `Just` value in `ConfigParams.ctlServerConfig` and make \
+          \sure that the `ctl-server` service is running and available at the \
+          \provided host and port"
```
klntsky, line 510: Please add a link to the docs on how to enable it in nix.

test/Plutip.purs
```diff
@@ -106,9 +106,7 @@ main = launchAff_ do
     -- we don't want to exit because we need to clean up after failure by
     -- timeout
     defaultConfig { timeout = Just $ wrap 30_000.0, exit = false }
-    do
-      suite
-      UtxoDistribution.suite
+    $ suite *> UtxoDistribution.suite
```
klntsky, line 109: It's better to have do notation in case we add more

klntsky (changes requested):

klntsky (approved):

PR #906, opened 2022-08-15
Update slot length handling for floating points

### LINKED ISSUE #868: Slot length is truncated on private testnet (Plutip)
During `Slot`~`POSIXTime` conversion, we might be inadvertently rounding a value somewhere to the nearest second. This might cause a division by zero error when using conversions and slotting arithmetic in Plutip tests

rynoV:
Closes #868.

TODO:

- [x] Use at least https://github.com/CardanoSolutions/ogmios/commit/8121673c7f266550e7b4f3f66c6e3938a6b78899 once it is cherrypicked into our fork
- [ ] Maybe add an ogmios fixture test for the result from plutip:
```json
[{"start":{"time":0.0,"slot":0,"epoch":0},"end":{"time":0.0,"slot":0,"epoch":0},"parameters":{"epochLength":50,"slotLength":0.1,"safeZone":10}},{"start":{"time":0.0,"slot":0,"epoch":0},"end":{"time":0.0,"slot":0,"epoch":0},"parameters":{"epochLength":80,"slotLength":0.2,"safeZone":30}},{"start":{"time":0.0,"slot":0,"epoch":0},"end":{"time":0.0,"slot":0,"epoch":0},"parameters":{"epochLength":80,"slotLength":0.2,"safeZone":30}},{"start":{"time":0.0,"slot":0,"epoch":0},"end":{"time":0.0,"slot":0,"epoch":0},"parameters":{"epochLength":80,"slotLength":0.2,"safeZone":30}},{"start":{"time":0.0,"slot":0,"epoch":0},"end":{"time" :0.0,"slot":0,"epoch":0},"parameters":{"epochLength":80,"slotLength":0.2,"safeZone":30}},{"start":{"time":0.0,"slot":0,"epoch":0},"end":{"time":16.0,"slot":80,"epoch":1},"parameters":{"epochLength":80,"slotLength":0.2,"safeZone":30}}]
```

### Pre-review checklist


src/QueryM/Ogmios.purs
```diff
@@ -430,18 +445,56 @@ instance Show EraSummaryParameters where
...
     epochLength <- getField o "epochLength"
-    slotLength <- getField o "slotLength"
+    slotLength <- getField o "slotLength" >>= stringifyAeson >>> parseSlotLength
     safeZone <- fromMaybe zero <$> getField o "safeZone"
     pure $ wrap { epochLength, slotLength, safeZone }
 
+slotLengthFactor :: Int
+slotLengthFactor = 1000
+
+parseSlotLength :: String -> Either JsonDecodeError SlotLength
```
ngua, line 455: I'm not sure if I'm missing something -- but can we just store the `SlotLength` as a `Number`?

Luis-omega: About that, since we have this [https://github.com/input-output-hk/ouroboros-network/blob/69fc48e9b0b7f845e2cfb40e8daaae7ea11548bb/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/EraParams.hs#L69](https://github.com/input-output-hk/ouroboros-network/blob/69fc48e9b0b7f845e2cfb40e8daaae7ea11548bb/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/EraParams.hs#L69 ) : 

```
 When a hard fork point is still unknown, we assume that each era determines a
  "safe zone": a number of slots from the tip of the ledger in which it is
  guaranteed that the hard fork will not happen.
  > CASE (i)
  >
  > ---------------A--------------*----------?--------------------
  >                               \..../
  >                                safe
  >                                zone
  Since the hard fork will not happen in the safe zone, we can extend the use of
  the current set of era parameters past the tip into the safe zone, giving us a
  limited ability to make predictions for the future (such as converting between
  slot numbers and wallclock time).
```

And as seen in [this slack discussion](https://mlabs-corp.slack.com/archives/C02QXGRGAPK/p1658940234733219),  the `safe zone` right now is around 5 days.

If the slot length becomes incredibly large, the mechanism to determine the `safe zone` and others may begin to work with fractions of `slots`. This means we would need to handle a new unit of time that would replace `slots` in most of its use cases. So, either this won't happen or we would need to modify the code to handle the new units aniway. 

So I believe there are low changes that `SlotLenght` would become incredibly long. It maybe would grow to 10 or 100 in the following years but not much more.

Luis-omega: Also, becoming incredibly small is even less probable to happen since it makes little sense to use something below micro seconds (even using microseconds is senseless for the kind of task a `slot` represents).

Luis-omega:
This change has a big impact on the functions and types inside `Types.Interval` 

Most of the types are made to work with `BigInt` and to threaten them as milliseconds, but the change to `Number` on `SlotLenght` and `RelativeTime` make us do lots of conventions. Converting from `BigInt` to `Number` is easy but converting in the opposite direction is tricky since sometimes we would like to round up and others we would want to truncate, also, the convention involves a possible failure.  So, we have three options here : 

- Keep it as is with mixed types `BigInt` and `Number`
- Change all the types to `Number`, this would mean that we may have problems converting to `POSIXTime`  (the same kind of problems mentioned above). 
- We can keep `SlotLenght` and `RelativeTime` as `BigInt` representing `Miliseconds`. In this case, we could forbid parsing time values below `0.001` (since we get them as a `Number` and at a seconds scale). The down side is that Ogmios fixtures could include values below `0.001` and this means a more complex test. Another option is to truncate to milliseconds the value obtained from Ogmios, but in this case, we could return to the original case in which `SlotLength` could be 0 if someone use a very low value. 

I'm a fan of avoiding using `Number` since it means we can't write the code naively but the second option seems to be the more `easy to go` solution here.

src/QueryM/Ogmios.purs
```diff
@@ -413,7 +419,7 @@ instance Show Epoch where
 
 newtype EraSummaryParameters = EraSummaryParameters
   { epochLength :: EpochLength -- 0-18446744073709552000 An epoch number or length.
-  , slotLength :: SlotLength -- <= 18446744073709552000 A slot length, in seconds.
+  , slotLength :: SlotLength -- <= 9,007,199,254,740,992 A slot length, in miliseconds.
```
Luis-omega, line 422: I'm not sure about leaving the mention of the range here since it's just the maximum range of `Number` and not a Cardano/Ogmios related restriction.

Luis-omega, line 422: I think we should delete the comment about the range here since it's just the `Number` maximum rather than a Cardano/Ogmios constraint.

src/Types/Interval.purs
```diff
@@ -508,7 +499,7 @@ maxSlot = wrap BigNum.maxValue
 data SlotToPosixTimeError
   = CannotFindSlotInEraSummaries Slot
   | StartingSlotGreaterThanSlot Slot
-  | EndTimeLessThanTime AbsTime
+  | EndTimeLessThanTime Number
```
Luis-omega, line 502: This error made me think about changing `Abstime` from `BiInt` to `Number` as most of the conversions are done by `Abstime` , but I don't know it would made sense in terms of `POSIXTime`.

test/Plutus/Time.purs
```diff
@@ -55,6 +55,9 @@ slotFixture = mkSlot 34892625
 absTimeFixture :: AbsTime
 absTimeFixture = AbsTime $ BigInt.fromInt 321541237
 
+absTimeNumberFixture :: Number
```
Luis-omega, line 58: See previous comment about `AbsTime`

test/Plutus/Time.purs
```diff
@@ -180,8 +189,49 @@ eraSummariesFixture = EraSummaries
           , "safeZone": mkSafeZone 129600
           }
       }
+  , EraSummary
```
Luis-omega, line 192: The intention here is to add a fixture for fractional values but maybe it needs more.

src/QueryM/Ogmios.purs
klntsky: We should add a comment mentioning both this constant and MAX_SAFE_INTEFER

src/QueryM/WaitUntilSlot.purs
```diff
@@ -86,12 +84,12 @@ waitUntilSlot futureSlot =
   retryDelay :: Milliseconds
   retryDelay = wrap 1000.0
 
-  logLag :: BigInt -> Milliseconds -> QueryM Unit
+  logLag :: Number -> Milliseconds -> QueryM Unit
```
klntsky, line 87: Why not `SlotLength`?

src/QueryM/Ogmios.purs
Luis-omega: I will leave the MAX_SAFE_INTEGER then, the previous constant isn't there anymore in Ogmios : 

https://github.com/CardanoSolutions/ogmios/commit/8121673c7f266550e7b4f3f66c6e3938a6b78899

src/QueryM/WaitUntilSlot.purs
Luis-omega: I think it would be worse as it only has 3 calls (is defined locally to `waitUntilSlot` ) and is used always with the result of a function that returns a `Number` instead of `SlotLength`  ( the function is `getSlotLength` ) .

ngua:
Looks like the Plutip test timed out on the most recent commit?

test/Plutus/Time.purs
```diff
@@ -92,7 +98,10 @@ mkEpochLength :: Int -> EpochLength
 mkEpochLength = EpochLength <<< BigInt.fromInt
 
 mkSlotLength :: Int -> SlotLength
-mkSlotLength = SlotLength <<< BigInt.fromInt
+mkSlotLength = SlotLength <<< Int.toNumber
+
+mkSlotLength' :: Number -> SlotLength
+mkSlotLength' = SlotLength
```
ngua, line 104: Why not just use the constructor directly?

CHANGELOG.md
```diff
@@ -103,6 +104,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
     - We require a `"version": 2` tag
     - `policy_id` must be 28 bytes
     - `asset_name` is up to 32 bytes
+- Problem parsing ogmios `SlotLenght` and `RelativeTime` in era Summaries if those include non integer values. ([#906](https://github.com/Plutonomicon/cardano-transaction-lib/pull/906))
```
ngua, line 107:
```suggestion
- Problem parsing ogmios `SlotLength` and `RelativeTime` in era Summaries if those include non integer values. ([#906](https://github.com/Plutonomicon/cardano-transaction-lib/pull/906))
```

test/Utils.purs
```diff
@@ -128,7 +153,7 @@ aesonRoundTrip
   => EncodeAeson a
   => a
   -> Either JsonDecodeError a
-aesonRoundTrip = decodeAeson <<< encodeAeson
+aesonRoundTrip = aesonRoundTripWith \x -> x
```
ngua, line 156:
```suggestion
aesonRoundTrip = aesonRoundTripWith identity
```

test/Plutus/Time.purs
ngua: Maybe we could get some from Plutip?

src/Types/Interval.purs
```diff
@@ -79,13 +79,14 @@ import Aeson.Decode ((</$\>), (</*\>))
 import Aeson.Decode as Decode
 import Aeson.Encode ((>$<), (>/\<))
 import Aeson.Encode as Encode
+import Aeson.Utils (maybeToEither)
```
ngua, line 82: `note`?

ngua: I don't know why `Aeson.Utils` re-implements `note` under a different name. I'd definitely prefer if we use `note`, we use it extensively elsewhere

test/Plutus/Time.purs
Luis-omega: Do you mean in the same way we get `Ogmios` fixtures in nix or just go look in plutip repo and copy some here?

ngua: I meant we could just make some fixtures by recording responses from a node started by Plutip. However, I don't think it should hold up this PR since you've included at least one fixture. We can open another issue and add more fixtures later

ngua: I don't think we can replicate the Ogmios testing approach here. We can do that because the Ogmios repo has a lot of test vectors that we can use. I don't think there's anything similar for Plutip

ngua (approved):

Issue #864, opened 2022-08-08
Investigate `MaxListenersExceededWarning` in nodejs


klntsky:
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 exit listeners added to [process]. Use emitter.setMaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
```

epiro-ps:
Here's the log with `--trace-warnings`:
```
(node:685761) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 exit listeners added to [process]. Use emitter.setMaxListeners() to increase limit
    at _addListener (node:events:601:17)
    at process.addListener (node:events:619:10)
    at /nix/store/4f9s131s302588580cgn80ghrh9viqh6-Test.Main-compiled/Node.Process/foreign.js:13:13
    at __do (/nix/store/4f9s131s302588580cgn80ghrh9viqh6-Test.Main-compiled/Plutip.Spawn/index.js:99:11)
    at runSync (/nix/store/4f9s131s302588580cgn80ghrh9viqh6-Test.Main-compiled/Effect.Aff/foreign.js:90:20)
    at run (/nix/store/4f9s131s302588580cgn80ghrh9viqh6-Test.Main-compiled/Effect.Aff/foreign.js:328:22)
    at /nix/store/4f9s131s302588580cgn80ghrh9viqh6-Test.Main-compiled/Effect.Aff/foreign.js:348:19
    at drain (/nix/store/4f9s131s302588580cgn80ghrh9viqh6-Test.Main-compiled/Effect.Aff/foreign.js:120:9)
    at Object.enqueue (/nix/store/4f9s131s302588580cgn80ghrh9viqh6-Test.Main-compiled/Effect.Aff/foreign.js:141:11)
```

klntsky:
This happens because we add global `process` `onExit` listeners in `Plutip.Spawn.killOnExit` for each child process.

We should instead add a single listener that reads a `Ref` containing all `ChildProcess`es to be killed, and maintain that `Ref` in `killOnExit`.

klntsky:
We are OK with passing it around.

jy14898:
The story has changed a little: we currently add exit handlers to clean up directories. This may be fixed with https://github.com/Plutonomicon/cardano-transaction-lib/pull/1201

klntsky:
Does not happen anymore

Issue #627, opened 2022-06-28
Avoid `execSync` in Plutip.Server


klntsky:
We added it in https://github.com/Plutonomicon/cardano-transaction-lib/pull/610#discussion_r908728739 because there's no easy to use NodeJS API that is wrapped in `Aff`. 

We should add an utility function `execSynAff`. It should get the stdout handle, read it until it's closed, and return the combined output in `Aff`, all without blocking (it should use `spawn`).

See See https://pursuit.purescript.org/packages/purescript-node-streams/5.0.0/docs/Node.Stream#t:Readable

klntsky:
Note the new [spawnAndCollectOutput](https://github.com/Plutonomicon/cardano-transaction-lib/blob/7f9c1d88a0d704eefa7a1fd1be4ff7c3a407d4e2/src/Internal/Test/E2E/Runner.purs#L621) function in [my recent PR ](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1075/files#)- I think we should use it to avoid duplicate work, because what it does matches is exactly what is needed here.

PR #918, opened 2022-08-17
Rework Plutip process management

jy14898:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/864.
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/627.

~Originally started to address 864, this PR includes a few changes:~
  - ~`spawnAndWaitForOutput` now filters by line, via nodes `readline` module. Before, we were converting each binary chunk to utf8 and then appending them, which can cause issues with multibyte characters. We are no longer appending all data together for the filter, but supplying a single line at each check. The `filter` parameters result is now lifted to `Aff`, which allows users to combine the lines via state if they please.~
  - ~We no longer add handlers to the process to check if it's being killed, and then invoke kill on the children (this is nodes default behaviour). The reason node wasn't closing was due to references of childrens stdout streams (https://github.com/mafintosh/why-is-node-running was very useful in debugging this)~
  - ~Redesigned process killing for Plutip server launching to kill in parallel, and wait for all children to exit after doing so. Not necessary for 864 in the end, however I think there's benefits in waiting for a process to be fully killed instead of just waiting for the port to close (db state for example)~

TODO: Redescribe

### Pre-review checklist


jy14898:
Closing in favour of https://github.com/Plutonomicon/cardano-transaction-lib/pull/1154

PR #931, opened 2022-08-22
Handle inline datums when parsing OgmiosTxOut, support inline datums in constraints, add tests & examples

### LINKED ISSUE #922: Add `PlutusV2` examples on `babbage`
We should add some example contracts that contain `PlutusV2` features

https://github.com/input-output-hk/Vasil-testnet

 - [x] [Collateral return](https://github.com/Plutonomicon/cardano-transaction-lib/pull/947)
Consume a utxo whose value is larger than needed, and return the remainder as change. Also test non-ada value being returned - this might require changes elsewhere.
 - [x] [Reference input](https://github.com/Plutonomicon/cardano-transaction-lib/pull/946)
 - [x] [Reference script](https://github.com/Plutonomicon/cardano-transaction-lib/pull/946)
 - [x] [Inline datum](https://github.com/Plutonomicon/cardano-transaction-lib/pull/931)
 - [ ] (optional) Secp256k1 functions used in a script. Plutip-only at this point

### LINKED ISSUE #928: Handling inline datums when parsing TransactionOutputs from Ogmios
Ogmios seems to correctly handle inline datums, but they disappear in the CTL parsing step:
> DEBUG] 2022-08-20T19:14:51.337Z message: {"type":"jsonwsp/response","version":"1.0","servicename":"ogmios","methodname":"Query","result":[[{"txId":"425a27a6e4891ae515a6a956cc34ec450e2a65f2933b57d6f3a3acfd179d4d9e","index":1},{"address":"addr_test1wqcvsw2rdp62d79d873umw034uyer0p8k7et9y28623v0xss3tp02","value":{"coins":1340410,"assets":{}},"datumHash":null,"datum":"d8799f46616162626363466161626263634661616262636346616162626363d8799fd8799f5820af2b7df401bf2c03a25151dcf2e954482d5b8cc8ddde467e3ec721e78fc71a3cff00ff581c0f45aaf1b2959db6e5ff94dbb1f823bf257680c3c723ac2d49f97546ff","script":null}]],"reflection":"Query-sktwi1zlgl72a55ja"}

When doing logInfo' $ show utxos

> [INFO] 2022-08-20T19:14:51.374Z (fromFoldable [(Tuple (TransactionInput { index: 1u, transactionId: (TransactionHash (hexToByteArrayUnsafe "425a27a6e4891ae515a6a956cc34ec450e2a65f2933b57d6f3a3acfd179d4d9e")) }) (TransactionOutput { address: (Address { addressCredential: (ScriptCredential (ValidatorHash (ScriptHash 30c839436874a6f8ad3fa3cdb9f1af0991bc27b7b2b29147d2a2c79a))), addressStakingCredential: Nothing }), amount: (PlutusValue (Map [(Tuple (CurrencySymbol (hexToByteArrayUnsafe "")) (Map [(Tuple (TokenName (hexToRawBytesUnsafe "")) fromString "1340410")]))])), datum: NoOutputDatum, referenceScript: Nothing }))])

gege251:
Closes #928
Partially closes #922

### Todo
- [x] Fix error when spending utxos with an inline datum
- [x] Add inline datum for `ScriptOutput`
- ~[ ] Add new utxo fixtures with inline datums~ I believe this is not necessary, as we also pull in fixtures from ogmios itself, and there are examples of inline datum. The reason this slipped through is because we allow additional fields in our parsing

### Pre-review checklist

- [x] Any new API feature includes a full example or integration test run with Plutip on CI


klntsky:
@gege251 are you going to complete the TODO in this PR?

gege251:
@klntsky My current priority is to get this working on some project, so I cannot do it right now, but after that I'd happily implement `ScriptOutput` handling.
Depends on your priorities, if this PR can wait, then I'm okay with putting this on a parking lot.

gege251:
btw, I think I'm pretty close to actually figure out all my issues, so maybe I can get back to this today.

jy14898:
@gege251 I hope it's okay if I take over this PR, I'm going to add the inline datum tests and examples as per https://github.com/Plutonomicon/cardano-transaction-lib/issues/922 into this PR, as they depend on it

src/Types/ScriptLookups.purs
```diff
@@ -1058,6 +1060,25 @@ processConstraint mpsMap osMap = do
...
+      networkId <- getNetworkId
+      let amount = fromPlutusValue plutusValue
+      runExceptT do
+        let
+          txOut = TransactionOutput
+            { address: validatorHashEnterpriseAddress networkId vlh
+            , amount
+            -- TODO: save correct datum and scriptRef, should be done in
+            -- Constraints API upgrade that follows Vasil
+            -- https://github.com/Plutonomicon/cardano-transaction-lib/issues/691
```
ngua, line 1074: Wrong issue?

src/Types/TxConstraints.purs
```diff
@@ -246,6 +249,15 @@ mustPayToScript vh dt vl =
   singleton (MustPayToScript vh dt vl)
     <> singleton (MustIncludeDatum dt)
 
+mustPayToScriptInlineDatum
+  :: forall (i :: Type) (o :: Type)
+   . ValidatorHash
+  -> Datum
+  -> Value
+  -> TxConstraints i o
+mustPayToScriptInlineDatum vh dt vl =
```
ngua, line 258: I know we want to be backwards-compatible, but I think we should just update `mustPayToScript` to take an `OutputDatum` instead. It's a breaking change, but I think upgrading won't be too bad

examples/InlineDatum.purs
```diff
@@ -0,0 +1,154 @@
...
+      logInfo' $ "The id "
+        <> show txId
+        <> " does not have output locked at: "
+        <> show scriptAddress
+  where
+  hasTransactionId :: TransactionInput /\ _ -> Boolean
+  hasTransactionId (TransactionInput tx /\ _) =
+    tx.transactionId == txId
+
+buildBalanceSignAndSubmitTx
```
ngua, line 164: Oh, we need to merge `develop` in soon where this helper is exported from `Examples.Helpers`

src/Types/ScriptLookups.purs
```diff
@@ -1058,6 +1060,25 @@ processConstraint mpsMap osMap = do
         -- constraint already.
         _cpsToTxBody <<< _outputs %= Array.(:) txOut
         _valueSpentBalancesOutputs <>= provideValue amount
+    MustPayToScriptInlineDatum vlh dat plutusValue -> do
+      -- TODO Check if vlh corresponds to v1 or v2?
```
ngua, line 1064: I guess we can't check for this particular constraint. I imagine the error from the evaluate or submit endpoints would point users to the problem if they tried to use inline datums with `PlutusV1` scripts

src/Types/ScriptLookups.purs
jy14898: This issue has been used all over the place in script lookups, it either needs to be changed or reopened, I think it was accidentally closed from other changes

src/Types/ScriptLookups.purs
jy14898: Indeed I tested it at some point, it failed with a pretty good message. Maybe I'll write a test to make sure it does fail rather than submit

src/Types/TxConstraints.purs
jy14898: Unfortunately we can't use `OutputDatum`, as the corresponding constructor for the normal `mustPayToScript` only holds the hash, not the datum.

src/Types/ScriptLookups.purs
jy14898: Actually it fails when trying to *spend*, meaning the value is locked forever. Not great as you kinda realise too late

ngua: Oh that's not ideal

src/Types/TxConstraints.purs
ngua: You mean `MustPayToScript`? AFAIK it takes a `Datum` argument and then hashes it

jy14898: No I mean the following DT:
https://github.com/Plutonomicon/cardano-transaction-lib/blob/4f84fceddbe9e7aecf319e0ceed0e9313ca1813f/src/Types/OutputDatum.purs#L30
`OutputDatumHash` is what corresponds to 'normal' datum providing, like in `mustPayToScript`. We need at least the datum, not just the hash

ngua: Sorry, I still don't understand what you mean. If we change `mustPayToScript` to take an `OutputDatum` instead of a `Datum`, we can just add it to the `TransactionOutput` directly (perhaps checking to ensure that `NoOutputDatum` wasn't provided). If a user provides an `OutputDatumHash`, then there's no inline datum and it works like the current `mustPayToScript` implementation

jy14898: The current implementation of `mustPayToScript` adds both `MustPayToScript` and `MustIncludeDatum` constraints. While the former only requires the hash, the latter requires the full datum. Hence the full datum is required whether we're using inline datum or not, and `OutputDatumHash` only holds the hash

src/Types/ScriptLookups.purs
jy14898: https://github.com/cardano-foundation/CIPs/pull/310/files#diff-4ffe445f8e72a6cee701a5a989aef62750d96c6296058bb612cad8ff305b49a5R24 touches on this exact issue actually

src/Types/TxConstraints.purs
ngua: Ah right, of course. I suppose having two constraints will work, although I would like to find a way to combine them at some point. We could add an additional arguments (maybe a sum type isomorphic to `Boolean`) to indicate whether or not the datum should be inline

jy14898: I've added a sum type, it's kinda ugly in practice though, at least in our examples with the qualified constraint import

src/Types/ScriptLookups.purs
```diff
@@ -987,68 +978,37 @@ processConstraint mpsMap osMap = do
...
-        maybe (liftEither $ Right unit) (ExceptT <<< addDatum) mDatum
-        let
-          liftDatumHash
-            :: forall (e :: Type) (dh :: Type)
-             . e
-            -> Maybe dh
-            -> Either e (Maybe dh)
-          liftDatumHash e Nothing = Left e
-          liftDatumHash _ (Just x) = Right (Just x)
-        -- [DataHash Note]
```
jy14898, line 1001: I've dropped this note as it doesn't really apply anymore, due to us no longer using `Maybe Datum`, but a custom sum type.

CHANGELOG.md
```diff
@@ -60,6 +60,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Improved error handling of transaction evaluation through Ogmios. This helps with debugging during balancing, as it requires the transaction to be evaluated to calculate fees. ([#832](https://github.com/Plutonomicon/cardano-transaction-lib/pull/832))
 - Flint wallet support ([#556](https://github.com/Plutonomicon/cardano-transaction-lib/issues/556))
 - `mustNotBeValid` constraint which marks the transaction as invalid, allowing scripts to fail during balancing and for Ogmios to allow submission. ([#947](https://github.com/Plutonomicon/cardano-transaction-lib/pull/947))
+- `Lose7Ada` example, for testing collateral return. ([#947](https://github.com/Plutonomicon/cardano-transaction-lib/pull/947))
+- `PlutusV2.AlwaysSucceeds` example, for testing PlutusV2 scripts. ([#947](https://github.com/Plutonomicon/cardano-transaction-lib/pull/947))
```
jy14898, line 64: Forgot to add these in #947...

ngua (approved):

PR #947, opened 2022-08-24
Restore babbage collateral return changes, add examples and test.

### LINKED ISSUE #922: Add `PlutusV2` examples on `babbage`

jy14898:
Partial closes #922 .

### Pre-review checklist

- [x] Any new API feature includes a full example or integration test run with Plutip on CI


src/BalanceTx/Collateral.purs
```diff
@@ -5,7 +5,6 @@ module BalanceTx.Collateral
       )
```
jy14898, line 5: I moved these functions from `QueryM` down to `Effect` so that we can use `selectCollateral` directly in `privateKeysToKeyWallet`. This was necessary due to module import cycles with `QueryM`.

test/Plutip.purs
```diff
@@ -336,6 +340,98 @@ suite = do
...
+          balance <- fold <$> getWalletBalance
+          let
+            collateralLoss = Value.lovelaceValueOf $ BigInt.fromInt $ -5_000_000
+          balance `shouldEqual` (balanceBefore <> collateralLoss)
+
+    test "runPlutipContract: AlwaysFails Native Asset Collateral Return" do
+      let
+        distribution :: InitialUTxOs /\ InitialUTxOs
+        distribution =
+          [] /\ [ BigInt.fromInt 2_100_000_000 ]
```
jy14898, line 390: Originally used a separate wallet for creating the initial utxos here to keep alices numbers round, although I moved balance checking to be relative to the start so not really necessary any more

src/Types/TxConstraints.purs
```diff
@@ -320,6 +323,9 @@ mustSatisfyAnyOf =
     >>> MustSatisfyAnyOf
     >>> singleton
 
+mustNotBeValid :: forall (i :: Type) (o :: Type). TxConstraints i o
```
ngua, line 326: We really need a doc comment for this, I don't think most users are familiar with the purpose of `isValid` and why you'd ever want to set it to `false`. I saw the note in the changelog but it should also be here

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `bytesFromPrivateKey` to `Serialization`
 - Improved error handling of transaction evaluation through Ogmios. This helps with debugging during balancing, as it requires the transaction to be evaluated to calculate fees. ([#832](https://github.com/Plutonomicon/cardano-transaction-lib/pull/832))
 - Flint wallet support ([#556](https://github.com/Plutonomicon/cardano-transaction-lib/issues/556))
+- `mustNotBeValid` constraint which marks the transaction as invalid, allowing scripts to fail during balancing and for ogmios to allow submission. ([#947](https://github.com/Plutonomicon/cardano-transaction-lib/pull/947))
```
ngua, line 62:
```suggestion
- `mustNotBeValid` constraint which marks the transaction as invalid, allowing scripts to fail during balancing and for Ogmios to allow submission. ([#947](https://github.com/Plutonomicon/cardano-transaction-lib/pull/947))
```

examples/AlwaysFails.purs
```diff
@@ -0,0 +1,136 @@
+-- | Warning: This contract will permanently lock 5ada
+-- | This module demonstrates how the `Contract` interface can be used to build,
+-- | balance, and submit a failing smart-contract transaction. It creates a
+-- | transaction that pays two Ada to the `AlwaysFails` script address, and
+-- | then attempts to spend the two Ada, failing and losing the collateral.
```
klntsky, line 5: Let's rename this example to LoseCollateral (it would be annoying to actually lose it by accidentally opening the example)

src/QueryM/Utxos.purs
```diff
@@ -157,7 +157,14 @@ getWalletCollateral = do
         addr <- liftAff $ (unwrap kw).address networkId
         utxos <- utxosAt addr <#> map unwrap >>> fromMaybe Map.empty
           >>= filterLockedUtxos
-        pure $ Array.singleton <$> (unwrap kw).selectCollateral utxos
+        pparams <- asks $ _.runtime >>> _.pparams <#> unwrap
+        let
+          coinsPerUtxoByte = pparams.coinsPerUtxoByte
+          maxCollateralInputs = fromMaybe 3 $ map UInt.toInt $
```
klntsky, line 163: Is it possible that the maximum is not set, i.e. any number is allowed?

If it is possible, let's not default to 3. If it is not possible, let's fail in case it's `Nothing`, because otherwise it would be very hard to debug.

klntsky (dismissed):

examples/AlwaysFails.purs
jy14898: Yeah I didn't include it in ByUrl for this reason. Also, I just realised it actually locks 7 ada. I'll update that too

src/QueryM/Utxos.purs
jy14898: That default was taken from the original collateral return PR. I think it was done this way to limit the complexity of `selectCollateral`.

I think 3 is a reasonable number, given 5 inputs would guarantee collateral due to min ada restrictions. Collateral return reduces the impact of using larger value utxos anyway.

~I'll push the choice of 3 into the definition of `BalanceTx.Collateral.selectCollateral`, where there's already arbitrary choices for the sake of timely termination.~ This will be tackled in https://github.com/Plutonomicon/cardano-transaction-lib/issues/971 so perhaps we should do nothing here for now

klntsky: Thank you!

klntsky (approved):

src/Wallet/Key.purs
```diff
@@ -42,12 +40,15 @@ import Serialization.Address
   )
 import Serialization.Types (PrivateKey)
 
+import BalanceTx.Collateral (selectCollateral) as Collateral
```
ngua, line 43: This import could be moved

ngua (approved):
LGTM

PR #954, opened 2022-08-25
Fix runPlutipTest not adding custom buildInputs

przembot:
Closes # .

`runPlutipTest` did not pass additional `buildInputs` to `runPursTest`

### Pre-review checklist

- [x] All code has been formatted using our config (`make format`)
- [x] Any new API features or modification of existing behavior is covered as defined in our [test plan](../doc/test-plan.md)
- [x] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

PR #955, opened 2022-08-26
Fix `runPlutipTest` not adding custom `buildInputs` (was #954)

ngua:
Changes from #954

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -260,7 +260,9 @@ let
           ogmios
           ogmios-datum-cache
           plutip-server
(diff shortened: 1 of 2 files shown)


nix/default.nix
```diff
@@ -254,14 +254,16 @@ let
...
         buildInputs = with pkgs; [
           postgresql
           ogmios
           ogmios-datum-cache
           plutip-server
-        ] ++ pkgs.lib.lists.optional withCtlServer pkgs.ctl-server;
-      } // args
+        ]
+        ++ (pkgs.lib.lists.optional withCtlServer pkgs.ctl-server)
+        ++ args.buildInputs;
```
ngua, line 265: Just realized this isn't going to work if no `buildInputs` are passed

PR #956, opened 2022-08-26
Expose the compiled project from `purescriptProject`

ngua:
It would be nice if our users could access the project including the compiled `output` (to simplify creating custom derivations, etc...), so I've added a `compiled` attribute to what `purescriptProject` returns. I've also added the generated `node_modules` to the return value as well

Not sure if `compiled` is the best name. I'm open to other suggestions. I didn't want to call it `project` (its internal name) because I think most people currently do `project = pkgs.purescriptProject { ... }` and `project.project` looks odd


jy14898:
> Not sure if compiled is the best name. I'm open to other suggestions.

`compiled` is fine, I'd either go with that or `built` to follow the naming elsewhere

klntsky (approved):

klntsky:
Can we have a note in the docs as well?

ngua:
@klntsky I'll add one

PR #944, opened 2022-08-24
Rory/preprod network

ngua:
Closes #923. A few things are broken now because `preprod` is still in Alonzo

### Pre-review checklist

- [ ] All code has been formatted using our config (`make format`)
- [ ] Any new API feature includes a full example or integration test run with Plutip on CI
- [ ] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

PR #958, opened 2022-08-26
Use valid firstBlock config for preprod

klntsky:
To drop the postgresql state, this should be executed in `psql`:

```
$ psql --host localhost --port 5432 -U ctxlib -d ctxlib
password: ctxlib

# update last_block set slot=5702409, hash='f737e8f756b2ece3e361cf252d122cff54562897532d83ba75ab679a7fd859d1'; 
```


ngua:
This should be merged into #944 I think

ngua (approved):

PR #959, opened 2022-08-26
Add CIP-30 mock using KeyWallet

### LINKED ISSUE #784: Mock CIP-30 using `KeyWallet`s in the browser
Since we have #649 , we can think about how to deliver CIP-30 testing with Plutip in the browser using `KeyWallet`s. The idea is to inject a piece of JS that contains both a private key and some glue code that can talk with the rest of CTL as defined in CIP-30.

This is worse than testing with real browser extensions in terms of coverage, but better than testing with NodeJS+Plutip (because we would also test CIP-30 interface and everything that can go wrong in the browser, and also we would not have to provide a public testnet pre-funded wallet).

I think this is a good middle ground that will make possible running Plutip and Chromium in CI.

klntsky:
Closes #784

### Pre-review checklist

- [x] Any new API features or modification of existing behavior is covered as defined in our [test plan](../doc/test-plan.md)

### DIFF
--- a/src/Wallet/Cip30Mock.purs
+++ b/src/Wallet/Cip30Mock.purs
@@ -2,28 +2,71 @@ module Wallet.Cip30Mock where
 
 import Prelude
 
-import Control.Monad.Error.Class (liftMaybe)
+import Contract.Monad (Contract, ContractEnv, wrapContract)
+import Control.Monad.Error.Class (liftMaybe, throwError, try)
 import Control.Monad.Reader (ask)
+import Control.Monad.Reader.Class (local)
 import Control.Promise (Promise, fromAff)
-import Data.Either (hush)
+import Data.Either (Either(Left, Right), hush)
 import Data.Foldable (fold)
+import Data.Lens ((.~))
+import Data.Lens.Common (simple)
+import Data.Lens.Iso.Newtype (_Newtype)
+import Data.Lens.Record (prop)
 import Data.Map as Map
-import Data.Maybe (Maybe)
+import Data.Maybe (Maybe(Just))
 import Data.Newtype (unwrap)
 import Deserialization.Transaction (deserializeTransaction)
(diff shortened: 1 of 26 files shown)


ngua:
I think this should be merged into #944 instead of `develop`, since we're still blocked on light wallet support

src/Wallet/Cip30Mock.purs
```diff
@@ -0,0 +1,131 @@
...
+  wallet <- liftAff mkWalletAff
+  let
+    setUpdatedWallet :: ContractEnv r -> ContractEnv r
+    setUpdatedWallet =
+      simple _Newtype <<< prop (Proxy :: Proxy "runtime") <<< prop
+        (Proxy :: Proxy "wallet") .~
+        (Just wallet)
+  res <- try $ local setUpdatedWallet contract
+  liftEffect deleteMock
+  case res of
```
ngua, line 71: Stylistic thing: my preference would be for `either` instead of an explicit pattern match

ngua: Actually we can just use `Helpers.liftEither`

src/Wallet/Cip30Mock.purs
```diff
@@ -0,0 +1,131 @@
...
+    Left err -> throwError err
+    Right r -> pure r
+  where
+  mkWalletAff = case mock of
```
ngua, line 75: Missing type signature

src/Wallet/Cip30Mock.purs
```diff
@@ -0,0 +1,131 @@
...
+    MockFlint -> mkFlintWalletAff
+    MockGero -> mkGeroWalletAff
+    MockNami -> mkNamiWalletAff
+  mockString = case mock of
```
ngua, line 79: Here too

ngua (approved):

PR #965, opened 2022-08-29
Terminate websockets instead of closing

klntsky:
### Pre-review checklist

- [x] Any new API features or modification of existing behavior is covered as defined in our [test plan](../doc/test-plan.md)

### DIFF
--- a/src/JsWebSocket.js
+++ b/src/JsWebSocket.js
@@ -65,4 +65,4 @@ exports._wsReconnect = ws => () => {
   ws.reconnect();
 };
 
-exports._wsClose = ws => () => ws.close();
+exports._wsTerminate = ws => () => ws.terminate();
--- a/src/JsWebSocket.purs
+++ b/src/JsWebSocket.purs
@@ -9,7 +9,7 @@ module JsWebSocket
   , _onWsMessage
   , _wsSend
   , _wsReconnect
-  , _wsClose
+  , _wsTerminate
   ) where
 
 import Prelude
@@ -57,4 +57,4 @@ foreign import _wsSend
 foreign import _wsReconnect
   :: JsWebSocket -> Effect Unit
 
-foreign import _wsClose :: JsWebSocket -> Effect Unit
+foreign import _wsTerminate :: JsWebSocket -> Effect Unit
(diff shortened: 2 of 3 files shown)


src/JsWebSocket.js
```diff
@@ -65,4 +65,10 @@ exports._wsReconnect = ws => () => {
   ws.reconnect();
 };
 
-exports._wsClose = ws => () => ws.close();
+exports._wsTerminate = ws => () => {
+  if (typeof BROWSER_RUNTIME == "undefined" || !BROWSER_RUNTIME) {
+    ws.close();
```
ngua, line 70: Shouldn't this be the other way around? IIRC Node's `ws` has both `terminate` and `close` methods, but the browser only has `close`

ngua (commented):
What's the motivation for this change?

src/JsWebSocket.js
klntsky: You are right. It's strange that the error does not appear in the console.

klntsky:
The motivation was to prevent websockets reconnect attempts after the end of contract execution.

jy14898:
Why do we even reconnect on error on the PS side, isn't that the whole purpose for using `reconnecting-websockets`?

jy14898:
> @jy14898 reconnecting-websockets can only reconnect, it does not replay the unfinished requests (because it can't know which requests have received a response and which haven't).

Resubmission is handled by the onConnect handler though, which gets invoked whether we call reconnect or not, because the ws is reconnecting
https://github.com/Plutonomicon/cardano-transaction-lib/blob/014e2b0119a16bd5ca936ac28d26ffeea8e9db35/src/Internal/QueryM.purs#L828-L833

klntsky:
@jy14898 if you want to re-check this could you please try the following:

1. remove the manual reconnect
2. start some long-running Contract (e.g. AlwaysSucceeds example)
3. stop the runtime
4. see the connection errors
5. start the runtime again
6. the contract should end successfully

I think it is not going to work, but I don't remember the exact reason.

jy14898:
> @jy14898 if you want to re-check this could you please try the following:
> 
>     1. remove the manual reconnect
>     2. start some long-running Contract (e.g. AlwaysSucceeds example)
>     3. stop the runtime
>     4. see the connection errors
>     5. start the runtime again
>     6. the contract should end successfully
> 
> 
> I think it is not going to work, but I don't remember the exact reason.

I just tried it and it works fine, the contract completed successfully:

image

errfrom (approved):
LGTM

Issue #688, opened 2022-07-06
Allow blake2b hashing outside of `Aff` context


ngua:
#520 made it so all blake2b hashing now requires an `Aff` context. It would be ideal if hashing was pure from the perspective of users. Hopefully we can figure out a way to load the WASM in JS without requiring `Aff` for the FFI bindings. Would top-level `await` work?

klntsky:
> Would top-level await work?

No.


```
       > Module parse failed: Cannot use keyword 'await' outside an async function (25594:2)
       > You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
       > |   }
       > |
       > >   await new Promise(
       > |       (resolve, reject) =>
       > |           Blake2bWasm.ready(error => {
       >  @ ./examples/index.js 8:0-22
```

klntsky:
Reposting myself from slack:

The library loads wasm asynchronously via a call to `blake2b.ready`. We can do something like:

  - call ready on module initialization
  -  call ready on contract initialization and wait for promise resolution
  -  expose ready as Aff function users need to wait on if they get errors when trying to use hashing before initialization
  -  make the hashing functions pure (throw exception in case wasm hasn't been loaded yet, explaining that the call should be delayed until ready)

PR #991, opened 2022-09-05
Allow `blake2b` hashing outside of `Aff` context

errfrom:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/688

### Pre-review checklist


ngua (commented):
I'm not sure if these changes are worth it. Using CSL for script and datum hashing has already been implemented on `babbage` and we're going to merge that into `develop` soon. For other uses of blake2b, I suggest just switching back to `blakejs`:

- I could perhaps see the motivation for switching to what theoretically should be a more performant version when we did our own script and datum hashing, but this has been solved by using CSL
- `blakejs` is probably fast enough
- it's better maintained
- I don't like the fact that a side effect must be called to use what should be a pure function (to be clear: I'm not criticizing your approach per se, you are pretty much constrained to do something like this to continue using the WASM package)

ngua (approved):

PR #1018, opened 2022-09-09
Add UtxoDistribution instance for Array of InitialUtxos and KeyWallet

marijanp:
For convenience I would like to be able to pass in the inital UTxO distribution as an Array and also get the KeyWallets as an Array.
I also added a basic testcase, that passes as you can see in the pipeline results.

~I created a basic testcase that fails currently. Tbh idk what all those functions are supposed to do. A little bit of help would be great!~

~`nix build .#checks.x86_64-linux.ctl-plutip-test -L` fails with the following error:~
```
Impossible happened: could not decode wallets. Please report as bug
```

### Pre-review checklist


marijanp:
I've implemented this instance for an older revision of ctl and that one does not cause any failures:
```
instance UtxoDistribution InitialUTxODistribution (Array KeyWallet) where
  encodeDistribution amounts = amounts
  decodeWallets privateKeyResponses = Just $ (\(PrivateKeyResponse key) -> privateKeysToKeyWallet (PrivatePaymentKey key) Nothing) <$> privateKeyResponse
```

src/Plutip/UtxoDistribution.purs
```diff
@@ -94,6 +94,18 @@ instance UtxoDistribution InitialUTxOsWithStakeKey KeyWallet where
         tail
   keyWallets _ wallet = [ wallet ]
 
+instance UtxoDistribution (Array InitialUTxOs) (Array KeyWallet) where
+  encodeDistribution amounts = amounts
+  decodeWallets d p = decodeWalletsDefault d p
```
klntsky, line 99: Avoid using `decodeWalletsDefault`, it has `guard $ Array.null remainingPKeys` which causes `Nothing` as a result

klntsky (changes requested):

src/Plutip/UtxoDistribution.purs
marijanp: thank you, I've changed the implementation and the test is passing now :)

klntsky:
Please also update the docs.

src/Plutip/UtxoDistribution.purs
```diff
@@ -94,6 +94,23 @@ instance UtxoDistribution InitialUTxOsWithStakeKey KeyWallet where
...
+  encodeDistribution amounts = amounts
+  decodeWallets _ privateKeyResponses =
+    Just $
+      ( \(PrivateKeyResponse key) -> privateKeysToKeyWallet
+          (PrivatePaymentKey key)
+          Nothing
+      ) <$> privateKeyResponses
+  decodeWallets' listOfInitialUTxOs privateKeyResponses =
+    let
+      wallets = Array.mapMaybe
```
klntsky, line 107: I doubt we should use mapMaybe here. If one decoding action fails, everything should fail. Otherwise the indices will all be messed up, right?

klntsky (changes requested):

src/Plutip/UtxoDistribution.purs
marijanp: Right, I replaced the logic to return `Nothing` if one of the decodings resutls in `Nothing`

src/Plutip/UtxoDistribution.purs
```diff
@@ -102,13 +103,10 @@ instance UtxoDistribution (Array InitialUTxOs) (Array KeyWallet) where
...
       ) <$> privateKeyResponses
-  decodeWallets' listOfInitialUTxOs privateKeyResponses =
-    let
-      wallets = Array.mapMaybe
-        (\utxos -> decodeWallets utxos privateKeyResponses)
-        listOfInitialUTxOs
-    in
-      Just (wallets /\ privateKeyResponses)
+  decodeWallets' listOfInitialUTxOs privateKeyResponses = do
+    wallets <- sequence $ (\utxos -> decodeWallets utxos privateKeyResponses)
```
klntsky, line 107: Use `traverse`.

klntsky (changes requested):
We also need to support `InitialUTxOsWithStakeKey` (see `withStakeKey`)

marijanp:
> We also need to support `InitialUTxOsWithStakeKey` (see `withStakeKey`)

Added an instance as well as a test

klntsky (approved):
Thanks!

klntsky:
@marijanp just realized there's no changelog entry. Could you please make a new PR?

Issue #875, opened 2022-08-10
Avoid using try/catch pattern in CSL bindings


klntsky:
When switching `CostModel` from `Data.Int` to `Types.Int` (#874) I discovered that in case when we try to create too many `BigNum`s with `from_str` method unsuccessfully, at some point we start hitting memory access errors. 

To reproduce:

```
lib = require("@emurgo/cardano-serialization-lib-nodejs");

for (var i = 0; i < 100000; i++) {
    try {
        lib.BigNum.from_str('-1');
    } catch (e) {
        console.log(e);
    }
}

lib.BigNum.from_str('1');
```

Error example:

```
> lib.BigNum.from_str('1');
Uncaught RuntimeError: memory access out of bounds
    at <dlmalloc::sys::System as dlmalloc::Allocator>::alloc::h1d32aef92d923b70 (<anonymous>:wasm-function[3639]:0x13219d)
    at dlmalloc::dlmalloc::Dlmalloc<A>::malloc::h817d0bdbe748994d (<anonymous>:wasm-function[57]:0x2f15f)
    at dlmalloc::Dlmalloc<A>::malloc::h25d3e8a45fd7c029 (<anonymous>:wasm-function[631]:0xb2243)
    at __rdl_alloc (<anonymous>:wasm-function[4717]:0x13be80)
    at __rust_alloc (<anonymous>:wasm-function[4534]:0x13b4a6)
    at __wbindgen_malloc (<anonymous>:wasm-function[3825]:0x13496f)
    at passStringToWasm0 (/nix/store/dqg2a0mp8ra1zgscb7az11xhjpdr8ppa-node-dependencies-cardano-transaction-lib-2.0.0-alpha/lib/node_modules/@emurgo/cardano-serialization-lib-nodejs/cardano_serialization_lib.js:79:15)
    at Function.from_str (/nix/store/dqg2a0mp8ra1zgscb7az11xhjpdr8ppa-node-dependencies-cardano-transaction-lib-2.0.0-alpha/lib/node_modules/@emurgo/cardano-serialization-lib-nodejs/cardano_serialization_lib.js:1437:20)

```

[...]

PR #1025, opened 2022-09-13
Pb99u069/875 eliminate try catch

pb99u069:
Partially addresses #875 

### Pre-review checklist


src/Serialization.js
```diff
@@ -12,13 +12,17 @@ const setter = prop => obj => value => () => obj["set_" + prop](value);
 exports.hashTransaction = body => () => lib.hash_transaction(body);
 
 exports.newBigNum = maybe => string => {
-  try {
-    return maybe.just(lib.BigNum.from_str(string));
-  } catch (_) {
-    return maybe.nothing;
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
+  if (/^-?[0-9]+$/.test(str)) {
```
klntsky, line 17: This check is not enough, it does not account for overflows. Please find the boundary sizes for BigNum and ensure that the numbers are in range by converting their string representations to native JS BigInt and comparing with the boundary,

src/Serialization.js
```diff
@@ -12,13 +12,17 @@ const setter = prop => obj => value => () => obj["set_" + prop](value);
...
+      return maybe.just(lib.BigInt.from_str(str));
```
klntsky, line 18: We should not remove try/catch entirely. We want to just add more checks _before_ using try-catch. The problem with wasm-bindgen arises only when an object is allocated in a _failing_ try block.

src/Serialization.js
```diff
@@ -12,13 +12,17 @@ const setter = prop => obj => value => () => obj["set_" + prop](value);
...
+  } else {
+      return maybe.nothing;
   }
 };
 
+console.log(/^-?[0-9]+$/.test("123"));
```
klntsky, line 24: Leftover

src/Serialization.js
```diff
@@ -83,10 +87,13 @@ exports.addVkeywitness = witnesses => witness => () => witnesses.add(witness);
...
 
 exports._publicKeyFromBech32 = maybe => bech32 => {
-  try {
-    return maybe.just(lib.PublicKey.from_bech32(bech32));
-  } catch (_) {
-    return maybe.nothing;
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
+  // prefixes defined in CIP-5: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0005 
+  if (/^(acct|acct_shared|addr|addr_shared|policy|pool|root|root_shared|stake|stake_shared)_x?vk1[0-9a-z]+$/.test(bech32)) {
```
klntsky, line 93: You need to find a single prefix. Construct a PublicKey value and call `to_bech32` on it to see the prefix. The simplest way is to add `console.log(key.to_bech32())` and run the test suite.

src/Serialization.js
```diff
@@ -83,10 +87,13 @@ exports.addVkeywitness = witnesses => witness => () => witnesses.add(witness);
...
+      return maybe.just(lib.BigInt.from_str(bech32));
```
klntsky, line 94: Not `BigInt`

src/Serialization.js
```diff
@@ -152,18 +159,17 @@ exports.setTxBodyMint = setter("mint");
...
-  try {
-    const str = bigInt.to_str();
-    if (str[0] == "-") {
-      return maybe.just(
-        lib.Int.new_negative(lib.BigNum.from_str(str.slice(1)))
-      );
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
+  const str = bigInt.to_str();
+  if (/^-[0-9]+$/.test(str)) {
```
klntsky, line 165: Same - no overflow checks

src/Serialization/BigInt.js
```diff
@@ -7,10 +7,12 @@ if (typeof BROWSER_RUNTIME != "undefined" && BROWSER_RUNTIME) {
...
 
-exports._BigInt_from_str = helper => str => {
-  try {
-    return helper.just(lib.BigInt.from_str(str));
-  } catch (_) {
-    return helper.nothing;
+exports._BigInt_from_str = maybe => str => {
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
+  if (/^-?[0-9]+$/.test(str)) {
```
klntsky, line 13: I think we can try constructing JS BigInt to check CSL's BigInt string for validity instead of relying on a regexp.

src/Types/BigNum.js
```diff
@@ -14,15 +14,34 @@ exports.zero = lib.BigNum.zero();
...
+    const r = a * b;
+    if (r < limit) {
+      return r;
+    } else {
+      throw new Error("Overflow detected");
+    }
+  };
+
   try {
+    check_mul(lhs.to_str(), rhs.to_str(), BigInt("18446744073709551616")); // 2 ^ 64
```
klntsky, line 44:
```suggestion
    check_mul(BigInt(lhs.to_str()), BigInt(rhs.to_str()), BigInt("18446744073709551616")); // 2 ^ 64
```

same above

src/Types/BigNum.js
```diff
@@ -14,15 +14,34 @@ exports.zero = lib.BigNum.zero();
 exports.one = lib.BigNum.one();
 
 exports.bnAdd = maybe => lhs => rhs => {
+  const check_add = (a, b, limit) => {
```
klntsky, line 17: These two functions can be factored out into a single one. Provide the operation (`+` or `*`) as a function-parameter.

klntsky (changes requested):

src/Serialization.js
pb99u069: removed the regex, as `BigInt(string)` in the boundary check already throws an error if the string is incorrect

src/Serialization.js
pb99u069: removed regex as well, as str is constructed from bigInt it is not necessary

src/Serialization.js
```diff
@@ -89,11 +96,14 @@ exports.newVkeyFromPublicKey = public_key => () => lib.Vkey.new(public_key);
 exports._publicKeyFromBech32 = maybe => bech32 => {
   // this is needed because try/catch overuse breaks runtime badly
   // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
-  // prefixes defined in CIP-5: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0005 
-  if (/^(acct|acct_shared|addr|addr_shared|policy|pool|root|root_shared|stake|stake_shared)_x?vk1[0-9a-z]+$/.test(bech32)) {
-      return maybe.just(lib.BigInt.from_str(bech32));
-  } else {
-      return maybe.nothing;
+  try {
+    if (/^ed25519_pk1[0-9a-z]+$/.test(bech32)) {
```
klntsky, line 113: Why `pk1` and not `pk`? I'm not sure about this, have you found the code it originates from?

src/Serialization.js
```diff
@@ -161,15 +171,19 @@ exports.newMint = () => lib.Mint.new();
...
   // this is needed because try/catch overuse breaks runtime badly
   // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
-  const str = bigInt.to_str();
-  if (/^-[0-9]+$/.test(str)) {
-  return maybe.just(lib.Int.new_negative(lib.BigNum.from_str(str.slice(1)))
-  );
-  } else if (/^[0-9]+$/.test(str)) {
-    return maybe.just(lib.Int.new(lib.BigNum.from_str(str)));
+  try {
+    check_limit(bigInt, -18446744073709551616, 18446744073709551616); // 2 ^ 64
```
klntsky, line 196: These should be JS `BigInt`s and not JS `Number`s. In fact, they overflow here.

klntsky, line 196: Also I am not sure about the exact limits, is it strictly less for positive and less-or-equal for negative? Please add the corner cases to the tests.

src/Serialization/BigInt.js
```diff
@@ -7,12 +7,13 @@ if (typeof BROWSER_RUNTIME != "undefined" && BROWSER_RUNTIME) {
...
-exports._BigInt_from_str = maybe => str => {
+exports._BigInt_from_str = helper => str => {
   // this is needed because try/catch overuse breaks runtime badly
   // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
-  if (/^-?[0-9]+$/.test(str)) {
-    return maybe.just(lib.BigInt.from_str(str));
-  } else {
-    return maybe.nothing;
+  try {
+    const check_str = BigInt(str);
```
klntsky: No need for a variable.

klntsky (changes requested):

test/Serialization.purs
```diff
@@ -116,6 +117,16 @@ suite = do
         txSerializedRoundtrip txFixture5
       test "Deserialization is inverse to serialization #6" $
         txSerializedRoundtrip txFixture6
+    group "BigNum tests" $ do
+      test "BigNum ok" $ do
+        let bn = "18446744073709551615"
+        (newBigNum maybeFfiHelper bn) `shouldSatisfy` isJust
```
klntsky, line 123:
```suggestion
        newBigNum maybeFfiHelper bn `shouldSatisfy` isJust
```

No need for parentheses here and below

klntsky (changes requested):

src/Serialization.js
jy14898: @klntsky 1 is the separator for bech32, so this is fine.

@pb99u069 The payload (hex part) must be 32 bytes long
https://github.com/Emurgo/cardano-serialization-lib/blob/04c4e7a0ee1e65253a47543be303740902820156/rust/src/chain_crypto/algorithms/ed25519.rs#L55
https://docs.rs/ed25519-dalek/0.8.0/ed25519_dalek/constant.PUBLIC_KEY_LENGTH.html

So we can also check if the string is `32/4+"ed25519_pk1".length` long

jy14898: @klntsky 1 is the separator for bech32, so this is fine.

@pb99u069 The payload (hex part) must be 32 bytes long
https://github.com/Emurgo/cardano-serialization-lib/blob/04c4e7a0ee1e65253a47543be303740902820156/rust/src/chain_crypto/algorithms/ed25519.rs#L55
https://docs.rs/ed25519-dalek/0.8.0/ed25519_dalek/constant.PUBLIC_KEY_LENGTH.html

So we can also check if the string is `32*8/4+"ed25519_pk1".length` long

src/Serialization.js
```diff
@@ -173,8 +190,15 @@ exports.setTxBodyMint = setter("mint");
...
 exports._bigIntToInt = maybe => bigInt => {
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
   try {
     const str = bigInt.to_str();
+    check_limit(
+      BigInt(str),
+      BigInt("-18446744073709551615"),
+      BigInt("18446744073709551616")
+    ); // 2 ^ 64
```
jy14898, line 201:
```suggestion
    check_limit(
      bigInt,
      BigInt("-18446744073709551615"),
      BigInt("18446744073709551616")
    ); // 2 ^ 64
    const str = bigInt.to_str();
```

src/Types/BigNum.js
```diff
@@ -13,16 +13,57 @@ exports.zero = lib.BigNum.zero();
...
 exports.bnMul = maybe => lhs => rhs => {
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
   try {
+    check_limit_with_func(
+      lhs.to_str(),
+      rhs.to_str(),
+      BigInt("18446744073709551616"),
+      mul
+    ); // 2 ^ 64
```
jy14898, line 66: Here, the arguments to `mul` are given as strings, but `mul` is using the built-in `*` js operator.

Perhaps this fixes:
```suggestion
    check_limit_with_func(
      BigInt(lhs.to_str()),
      BigInt(rhs.to_str()),
      BigInt("18446744073709551616"),
      mul
    ); // 2 ^ 64
```

jy14898: Same with the `add` version. I also don't think we need this `check_limit_with_func`, just apply the function directly?

test/Serialization.purs
```diff
@@ -116,6 +117,16 @@ suite = do
         txSerializedRoundtrip txFixture5
       test "Deserialization is inverse to serialization #6" $
         txSerializedRoundtrip txFixture6
+    group "BigNum tests" $ do
+      test "BigNum ok" $ do
+        let bn = "18446744073709551615"
+        newBigNum maybeFfiHelper bn `shouldSatisfy` isJust
+      test "BigNum owerflow" $ do
```
jy14898, line 124:
```suggestion
      test "BigNum overflow" $ do
```

jy14898 (changes requested):

src/Serialization.js
```diff
@@ -11,8 +11,19 @@ const setter = prop => obj => value => () => obj["set_" + prop](value);
 
 exports.hashTransaction = body => () => lib.hash_transaction(body);
 
+const check_limit = (num, lower_limit, upper_limit) => {
```
ngua, line 14: This applies to several other new functions in this PR: please change the names to use `camelCase` instead of `snake_case`, since that is the idiom in Javascript. CSL exports `snake_case` identifiers because the library itself is written in Rust which uses that casing convention (the JS bindings are generated from the Rust code)

src/Types/BigNum.js
```diff
@@ -13,16 +13,57 @@ exports.zero = lib.BigNum.zero();
...
 
+const add = (a, b) => {
+  return a + b;
+};
+
+const mul = (a, b) => {
+  return a * b;
+};
+
+const check_limit_with_func = (a, b, limit, func) => {
```
ngua, line 24: See my other comments about casing

src/Types/BigNum.js
```diff
@@ -13,16 +13,57 @@ exports.zero = lib.BigNum.zero();
 
 exports.one = lib.BigNum.one();
 
+const add = (a, b) => {
```
ngua, line 16: Since you're using an arrow function (which is good, since that's consistent with the rest of our JS code) and the function body is a single line, you can omit the `return` and the braces

```suggestion
const add = (a, b) => a + b;
```

src/Types/BigNum.js
```diff
@@ -13,16 +13,57 @@ exports.zero = lib.BigNum.zero();
...
+  return a + b;
+};
+
+const mul = (a, b) => {
```
ngua, line 20: Same here

src/Types/BigNum.js
```diff
@@ -13,16 +13,57 @@ exports.zero = lib.BigNum.zero();
...
+};
+
 exports.bnAdd = maybe => lhs => rhs => {
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
   try {
+    check_limit_with_func(
+      lhs.to_str(),
+      rhs.to_str(),
+      BigInt("18446744073709551616"),
```
ngua, line 48: What is the significance of this number? I can guess, but it should be bound to a variable with a descriptive name. The literal is repeated as well in different places, which makes all of the checks more susceptible to typos

src/Serialization.js
```diff
@@ -11,8 +11,19 @@ const setter = prop => obj => value => () => obj["set_" + prop](value);
...
+  } else {
+    throw new Error("Overflow detected");
+  }
+};
+
 exports.newBigNum = maybe => string => {
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
   try {
+    check_limit(BigInt(string), BigInt("0"), BigInt("18446744073709551616")); // 2 ^ 64
```
ngua, line 26: See my other comment about magic numbers

ngua (changes requested):

src/Internal/Serialization.js
```diff
@@ -96,13 +111,25 @@ exports.addVkeywitness = witnesses => witness => () => witnesses.add(witness);
...
+      return maybe.just(lib.PublicKey.from_str(bech32));
+    } else {
+      throw new Error("Wrong prefix");
+    }
   } catch (_) {
     return maybe.nothing;
   }
 };
 
+console.log((32 * 8) / 4 + "ed25519_pk1".length);
```
ngua, line 128: I'm assuming these are leftover `log`s?

src/Internal/Serialization.js
```diff
@@ -173,8 +200,17 @@ exports.setTxBodyMint = setter("mint");
 exports.newMint = () => lib.Mint.new();
 
 exports._bigIntToInt = maybe => bigInt => {
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
   try {
     const str = bigInt.to_str();
+    checkLimit(BigInt(str), bigNumLimitNeg, bigNumLimit);
+    // checkLimit(
```
ngua, line 208: Redundant comment

ngua (commented):
Just a few final comments from me

src/Serialization.js
klntsky: @jy14898 what did you mean by this suggestion? The first BigInt is CSL BigInt, and the other is native. Can't pass one instead of the other.

src/Serialization.js
jy14898: The maths isn't that simple (nor is it hex)

src/Internal/Serialization.purs
```diff
@@ -19,6 +19,7 @@ module Ctl.Internal.Serialization
   , publicKeyFromPrivateKey
   , privateKeyFromBytes
   , makeVkeywitness
+  , newBigNum
```
jy14898, line 22: We should remove `newBigNum` from serialization - nothing uses it. Replace your uses in `Test.Serialization` with `Types.BigNum.fromString`

src/Internal/Serialization.js
```diff
@@ -11,8 +11,23 @@ const setter = prop => obj => value => () => obj["set_" + prop](value);
...
+};
+
 exports.newBigNum = maybe => string => {
+  // this is needed because try/catch overuse breaks runtime badly
+  // https://github.com/Plutonomicon/cardano-transaction-lib/issues/875
   try {
+    checkLimit(BigInt(string), BigInt("0"), bigNumLimit);
     return maybe.just(lib.BigNum.from_str(string));
   } catch (_) {
     return maybe.nothing;
```
jy14898, line 33: See the other comment on this function. TL;DR it should be removed

jy14898 (commented):
Generally LGTM. It does worry me when we have PRs that don't test the specific issue.

I did have an idea of how we can test specifically for memory leaks: allocate something (BigNum is fine) and access the internal pointer, run the test, and allocate again and check the distance between the pointers. Might be fickle though, depends on how `malloc` is implemented in CSL.

That shouldn't hold this PR up however

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Removed
 
+- `newBigNum` from module Ctl.Internal.Serialization as it is not used ([#1025](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1025))
+
```
jy14898, line 51: Generally we want to avoid including internal changes in the changelog, otherwise it'll get cluttered and it'll be of less use to users

```suggestion
```

jy14898 (changes requested):

jy14898 (approved):
LGTM If the merge succeeds on CI

PR #1028, opened 2022-09-13
Exit on plutip test failure

### LINKED ISSUE #983: `ctl-plutip-test` check is green with failing tests

jy14898:
Closes #983.

### Pre-review checklist


test/Plutip.purs
```diff
@@ -146,9 +146,7 @@ import Wallet.Cip30Mock
 main :: Effect Unit
 main = launchAff_ do
   Utils.interpretWithConfig
-    -- we don't want to exit because we need to clean up after failure by
-    -- timeout
-    defaultConfig { timeout = Just $ wrap 50_000.0, exit = false }
+    defaultConfig { timeout = Just $ wrap 50_000.0, exit = true }
```
jy14898, line 149: I think it got set back to false because of the misleading comment. It's the case (today at least) that the test runner only exists on all tests being run, not the first failure.

test/Utils.purs
```diff
@@ -67,9 +67,7 @@ interpret = interpretWithConfig defaultConfig { timeout = Just (wrap 50000.0) }
 interpretWithTimeout
   :: Maybe Milliseconds -> TestPlanM (Aff Unit) Unit -> Aff Unit
 interpretWithTimeout timeout spif = do
-  plan <- planT spif
-  runSpec' defaultConfig { timeout = timeout } [ consoleReporter ] $
-    planToSpec plan
+  interpretWithConfig (defaultConfig { timeout = timeout }) spif
```
jy14898, line 70: Made this change while I was beginning to use `runSpecT` instead of `runSpec'`, but might as well keep it

test/Plutip.purs
klntsky: Could you please check that when one of the tests fails by timeout, the services are killed and the subsequent tests can run

jy14898: I checked with a 5 second timeout, which is enough for processes to start being launched, and the following tests continue to hit their timeout, implying they weren't immediately failing due to port conflicts.

I'll try and find a value in which some do pass and some timeout

klntsky: Check with 15 sec to be sure the action gets cancelled in Contract code

jy14898:
```
  ✓︎ runPlutipContract
  ✓︎ runPlutipContract: Pkh2Pkh
  ✗ runPlutipContract: Pkh2Pkh with stake key:

  test timed out after 20000ms
  ✓︎ runPlutipContract: parallel Pkh2Pkh
  ✗ runPlutipContract: parallel Pkh2Pkh with stake keys:

  test timed out after 20000ms
```

Looks good to me

jy14898: This can be seen actually in this PRs currently running check:

https://ci.staging.mlabs.city/build/12715/nixlog/1

A bit concerning that some are still hitting the 50s timeout

jy14898: No longer the case, I guess it's dependent on CI load

PR #1034, opened 2022-09-13
Eternl and Lode E2E on preview

klntsky:
Closes #1007 #1030

### Pre-review checklist

- [ ] All code has been formatted using our config (`make format`)
- [ ] Any new API features or modification of existing behavior is covered as defined in our [test plan](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/test-plan.md)
- [ ] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

PR #1039, opened 2022-09-14
Module Reorganisation

### LINKED ISSUE #820: Move all internal/non-`Contract` modules to clearly delineated namespace
Our current module naming scheme needs improvement. It 
- is more appropriate for applications (e.g. `Types.*`)
- occupies too much of the top-level namespace
- leads to unclear and confusing documentation
- doesn't reinforce the fact that users shouldn't import non-`Contract` modules

**All** internal/non-`Contract` modules should be moved to a clearly delineated internal namespace. I propose `Ctl.Internal.*`. This includes all of our examples, test, etc...

jy14898:
Closes #820.

https://github.com/Plutonomicon/cardano-transaction-lib/blob/9459d92ee6483cdc89990b2f08bd0b9e8e791706/CHANGELOG.md?plain=1#L40-L43

### Pre-review checklist


ngua (commented):
We should merge this in as soon as possible, but I have one comment: our convention so far has been to only capitalize the first letter in initialisms/acronyms. You've chosen `CTL`, but `Ctl` would be more consistent with our naming scheme. I won't insist on changing this, but what are your thoughts on that?

ngua (commented):
Also, I don't think that we should change `Contract` to `CTL.Contract`:
- It's reasonable to reserve the `Contract` namespace for CTL, especially since the likelihood of name clashes with user modules or modules from other libraries is very low
- it will make upgrading less pleasant for users

jy14898:
> our convention so far has been to only capitalize the first letter in initialisms/acronyms. You've chosen `CTL`, but `Ctl` would be more consistent with our naming scheme. 

I thought about this a little, modules in the package set are somewhat split, I hadn't surveyed our own. I could make an argument to keep it how it is (but maybe not convincing): CIP/JSON/CBOR are all acronyms, they're pronounced the way they're spelled and are a new word. CTL is an initialism, where we pronounce the individual letters, and afaik no-one says seetle.

It's easy to change though 🤷

> Also, I don't think that we should change `Contract` to `CTL.Contract`

I placed them all under CTL for consistency, but I suppose ideally users don't see the CTL namespace anyway, so it's no more consistent.

What about `CTL.Plutus`? Drop the CTL there too?

jy14898:
I suppose by my argument, JsonWsp should be JsonWSP and JsWebSocket JSWebSocket, FfiHelpers FFIHelpers... I'll change it to Ctl

jy14898:
@ngua 

> I would move `Plutus` modules to `Ctl.Internal`. Ideally we expose everything through `Contract`

I'll have to make the re-exports then, we don't re-export `Plutus` yet we use it in the examples. Actually I'll leave re-exporting for a later PR

ngua:
@jy14898 I'd like to merge #1034 first, then we can merge this. Sorry to make you fix more merge conflicts :(

jy14898:
> @jy14898 I'd like to merge #1034 first, then we can merge this. Sorry to make you fix more merge conflicts :(

No worries, better to be me than others

ngua (changes requested):
Very excited for this change!

Please update the module name in the `spago test` invocation in `test/ctl-e2e-test.sh`

templates/ctl-scaffold/flake.nix
```diff
@@ -10,7 +10,7 @@
       type = "github";
       owner = "Plutonomicon";
       repo = "cardano-transaction-lib";
-      rev = "0742820c84b964a894deda884677fd2b461458cf";
+      rev = "39fc516c7c837401427b46637dacd9c7ada6274d";
```
ngua, line 13: You should also run `nix flake lock` to update the lockfile

ngua (changes requested):

ngua (approved):

Issue #1036, opened 2022-09-14
Improve failure handling for `Contract` testing utilities


errfrom:
Currently, the process of handling `Contract.Test.Utils` assertions will stop on the first failure. 
We need to extend the current behaviour to accumulate failures before exiting.

PR #1044, opened 2022-09-15
Some improvements to the `Contract` testing utilities 

errfrom:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/1036

### Updates
1. Now all assertion failures are accumulated and displayed to the user as a list (previously only the first failure message was displayed)
2. New assertions: `assertOutputHasRefScript`
3. New checks: `checkOutputHasRefScript`, `checkTxHasMetadata`

### Pre-review checklist


jy14898 (commented):
Apologies for taking a while on this review, it's quite confusing to me the way things are done. I feel like it's maybe over generalised for what we use it for? Maybe more examples in `ContractTestUtils`?

jy14898 (approved):
To summarize the changes to make it clear I've understood
 - Assertion throwing has been lifted into a new monad, `ContractAssertionM`, rather than the native `Contract`, so that assertion failure can be caught and recorded, and the continuation of the contract/further checks can continue. `runContractAssertionM` implements this logic.
 - The assertions which use `runContractAssertionM'` are 'hard' assertions, they do not continue the chain of contract checks.

I suppose we have to be careful with `runContractAssertionM` - the assertion must set the `Last` writer after the contract has run, otherwise it may be run twice on failure which breaks the contract of `ContractWrapAssertion`.

LGTM nonetheless

errfrom:
@jy14898 Thanks for your review! 
The idea behind having both `ContractTestM` and `ContractAssertionM` is to allow individual assertions to fail and stop executing, while still allowing further assertions to be performed. As for your second point, `runContractAssertionM'` is only used with basic assertions which don't need to be wrapped in order to compose, and which are simply traversed, so they **don't actually break the chain of contract checks.**
```PureScript
instance
  ContractAssertions (Array (ContractBasicAssertion r a b)) r a where
  wrapAndAssert contract assertions =
    contract >>= \r -> traverse_ (_ $ r) assertions *> pure r
```

PR #1052, opened 2022-09-19
Add an error message for submitting to a closed WebSocket

### LINKED ISSUE #1032: Throw an error when attempting to send data using closed websocket

MitchyCola:
Closes #1032 .

### Pre-review checklist


src/QueryM.purs
```diff
@@ -1141,10 +1142,17 @@ mkRequestAff' listeners' webSocket logger jsonWspCall getLs inp = do
...
+      wsStatus <- _wsIsOpen webSocket
+
+      let
+        closedWS = logger Error "WebSocket is closed, could not submit request."
+        submitReq = _wsSend webSocket (logger Debug) sBody
       -- Uncomment this code fragment to test `SubmitTx` request resend logic:
       -- when (isJust $ getRequestInputToStore inp) $
       --   _wsReconnect webSocket
+
+      if wsStatus then submitReq else closedWS
```
klntsky, line 1154: I think we should throw as well.

klntsky (changes requested):

src/QueryM.purs
MitchyCola: Done.

klntsky (approved):

src/QueryM.purs
```diff
@@ -1141,10 +1142,20 @@ mkRequestAff' listeners' webSocket logger jsonWspCall getLs inp = do
               _ -> cont (lmap dispatchErrorToError result)
         )
       respLs.addRequest id (sBody /\ getRequestInputToStore inp)
-      _wsSend webSocket (logger Debug) sBody
+      wsStatus <- _wsIsOpen webSocket
```
Luis-omega, line 1145: I'm not familiar with how WebSockets can fail in purescript, but, this pattern seems similar to : 

- Check if file exist 
- Open file 

And the usual advice is not to do that since files can be deleted or moved after the check and before we try to open.

MitchyCola: In general, I do agree with the sentiment. However, there really isn't a way to avoid it since this is the same logic that is being used in the JavaScript libraries that we depend on: [ws](https://github.com/websockets/ws/blob/00c34d726dca1c558fe5ee5e346979159b2297fe/lib/websocket.js#L438-L453) and [reconnecting-websocket](https://github.com/pladaria/reconnecting-websocket/blob/ef6e6a7fc9d5078753cf36ba270e382af1f11d9e/reconnecting-websocket.ts#L252-L255)

jy14898 (changes requested):
Thinking about what @Luis-omega said, it made me think about how this applies to reconnecting websockets. When you send, messages get buffered until the ws is connected again. Thus, it's fine for us to send a message when the socket isn't open (eg connecting) (which is what this PR checks for). Not only that, if the reconnecting websocket is manually closed to stop reconnecting, and then reconnect is called, those messages would be eventually sent.

I'm not sure what we should do if that is true

src/QueryM.purs
Luis-omega: really? all my experience with sockets is in python and there we would try and catch. It's fine then.

klntsky:
@jy14898 your concerns are valid.

MitchyCola:
@jy14898  Now the request will be submitted to the closed WS where it will be sent to the buffer, it will still throw the error.

jy14898:
@klntsky What was the original motivation for the issue? I'm struggling to see the need to merge this PR

klntsky:
It definitely should not be merged in its current state. It should probably be closed in favor of https://github.com/Plutonomicon/cardano-transaction-lib/pull/965

PR #1062, opened 2022-09-22
new functions addressWithNetworkTagFromBech32 and addressFromBech32

pb99u069:
Closes # .

### Pre-review checklist


src/Contract/Address.purs
```diff
@@ -4,7 +4,9 @@ module Contract.Address
   , enterpriseAddressStakeValidatorHash
   , enterpriseAddressValidatorHash
   , getNetworkId
+  , addressWithNetworkTagFromBech32
   , addressWithNetworkTagToBech32
+  , addressFromBech32'
```
pb99u069, line 9: If I try to export `addressFromBech32` I get an `ExportConflict` error (Export for value Serialization.Address.addressFromBech32 conflicts with value Contract.Address.addressFromBech32) which I find strange as I don't export but import `Serialization.Address.addressFromBech32`. Anyway, should I just give this function another name or is there a workaround?

pb99u069:
no need for tests for these functions, if I read the test plan correctly?

src/Contract/Address.purs
jy14898: It is exported, via the `module SerializationAddress` export. I'd argue the one from `SerializationAddress` shouldn't be exported given the new functions and can replace it. @klntsky what do you think?

pb99u069: thanks for the tip, I removed `addressFromBech32` from `module SerializationAddress`. however, the new `addressFromBech32` changes the API as it returns a `Contract r (Maybe Address)` instead of a `Maybe Address`. a problem?

jy14898: We'll need to have the changelog reflect under Changes that `addressFromBech32` is lifted to `Contract` now.

@ngua What do you think the migration strategy should be? I think for users to use the old one, they had to dip into internal functions anyway to get from CSL address to plutus address, so really it's unsupported to use the old `addressFromBech32`?

ngua:
@pb99u069, I think it would be best to add some tests. Maybe I should update the test plan to make this clearer, but any functions or effects exported from `Contract.*` should have a test moving forward (these can either be a dedicated test or inclusion in a larger test, as long as the test depends on the new functionality succeeding). We weren't as good about this in the past as we should have been to be quite honest

The tests can be quite simple. For expected failures (e.g. the actual network ID in the contract environment not matching the network tag), just use `try` and if you get a `Left` back, the test passes

ngua (changes requested):
The implementation looks fine to me, please add the tests mentioned in my comment above

src/Contract/Address.purs
```diff
@@ -192,6 +207,21 @@ addressToBech32 address = do
...
+-- | corresponds to the contract environment `networkId`
+addressFromBech32
+  :: forall (r :: Row Type). Bech32String -> Contract r (Maybe Address)
+addressFromBech32 str = do
+  networkId <- getNetworkId
+  cslAddress <- liftContractM "unable to read address" $
+    SA.addressFromBech32 str
+  address <- liftContractM "unable to convert to plutus address" $
+    toPlutusAddress cslAddress
+  if networkId == addressNetworkId cslAddress then
```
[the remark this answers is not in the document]
pb99u069: I removed the maybe from the type signature. Is throwing with `assertTrue` the right choice? and in `Test.Ctl.NetworkId` I'm not sure how to test a mainnet address that is supposed to throw an error. is there a function for this?

test/NetworkId.purs
```diff
@@ -0,0 +1,35 @@
...
+
+testTestnetAddress :: Aff Unit
+testTestnetAddress = runContract testnetConfig do
+  let
+    bechstr =
+      "addr_test1qqm0z9quxyefwwq902p5f9t4s35smhegjthhhqpeclnpx2rzhuq2p6jahnky7qqua9nz9tcw6nlgy6cpjvmlaaye4apqzc6ppq"
+  addr <- Address.addressFromBech32 bechstr
+  addressBech32 (fromPlutusAddress TestnetId addr) `shouldEqual` bechstr
+
+-- testMainnetAddress :: Aff Unit
```
klntsky, line 31: A leftover

klntsky (changes requested):

test/NetworkId.purs
pb99u069: are tests ok without `testMainnetAddress`? wasn't sure how to do this (see question above), but I can also just delete it

test/Integration.purs
```diff
@@ -42,6 +43,7 @@ testPlan = do
   Collateral.suite
   PrivateKey.suite
   Logging.suite
+  NetworkId.suite
```
klntsky, line 46: Please let's move this test to plutip tests. The reason for this is that integration tests are not run automatically, while this test definitely can.

klntsky (changes requested):

klntsky (approved):

packages.dhall
```diff
@@ -105,8 +105,8 @@ in  upstream
 -------------------------------
 -}
 let upstream =
-      https://github.com/purescript/package-sets/releases/download/psc-0.14.5-20211116/packages.dhall
-        sha256:7ba810597a275e43c83411d2ab0d4b3c54d0b551436f4b1632e9ff3eb62e327a
+      https://github.com/purescript/package-sets/releases/download/psc-0.14.5/packages.dhall
+        sha256:7ed6350fe897a93926d16298e37d2324aabbe5eca99810204719dc3632fb555f
```
jy14898, line 109: What's the reason for downgrading the packageset?

jy14898 (changes requested):

packages.dhall
pb99u069: sorry, no idea why this happened. I made some psc changes on my local machine to make lsp run in my emacs but don't know why this affected this file. I have reset it.

jy14898:
Updated to develop to re-run CI. One of the contracts timed out, which is surprising given our current limit

CHANGELOG.md
```diff
@@ -38,11 +38,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
 - `Parallel` instance to `Contract` monad. Parallel capabilities are in the associated `ParContract` datatype ([#1037](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1037))
+- new functions `addressWithNetworkTagFromBech32` and `addressFromBech32`in `Contract.Address`, the second checking that address network Id corresponds to the contract environment network Id. ([#1062](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1062))
 
 ### Changed
 
 - Reorganised the library into new namespaces. Namely: library internals, tests, and examples are now under `Ctl.Internal.*`, `Test.Ctl.*`, and `Ctl.Examples.*` respectively. Documentation and comments have been updated to use these new names, but not entries of previous releases in the changelog. ([#1039](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1039))
 - Switched to `preview` testnet by default ([#1030](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1030))
+- `addressFromBech32` checks that address network Id corresponds to the contract environment Id and is therefore lifted to the `Contract` monad ([#1062](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1062))
```
klntsky, line 47:
```suggestion
```

src/Contract/Address.purs
```diff
@@ -192,6 +208,20 @@ addressToBech32 address = do
   pure $ addressWithNetworkTagToBech32
     (AddressWithNetworkTag { address, networkId })
 
+-- | Convert `Bech32String` to `Address`, asserting that the address `networkId`
+-- | corresponds to the contract environment `networkId`
+addressFromBech32
+  :: forall (r :: Row Type). Bech32String -> Contract r Address
+addressFromBech32 str = do
+  networkId <- getNetworkId
+  cslAddress <- liftContractM "unable to read address" $
```
klntsky, line 217:
```suggestion
  cslAddress <- liftContractM "addressFromBech32: unable to read address" $
```

src/Contract/Address.purs
```diff
@@ -192,6 +208,20 @@ addressToBech32 address = do
...
+    SA.addressFromBech32 str
+  address <- liftContractM "unable to convert to plutus address" $
```
klntsky, line 219:
```suggestion
  address <- liftContractM "addressFromBech32: unable to convert to plutus address" $
```

src/Contract/Address.purs
```diff
@@ -118,6 +125,7 @@ import Ctl.Internal.Types.UnbalancedTransaction
 import Data.Array (head)
 import Data.Maybe (Maybe)
 import Data.Traversable (for, traverse)
+import Test.Ctl.Utils (assertTrue)
```
klntsky, line 128: We can't use this module. It is from the test dir, and using it breaks upstream, because `test/` dir is not included as a source dir when CTL is used as dependency

src/Contract/Address.purs
```diff
@@ -192,6 +208,20 @@ addressToBech32 address = do
...
+-- | corresponds to the contract environment `networkId`
+addressFromBech32
+  :: forall (r :: Row Type). Bech32String -> Contract r Address
+addressFromBech32 str = do
+  networkId <- getNetworkId
+  cslAddress <- liftContractM "unable to read address" $
+    SA.addressFromBech32 str
+  address <- liftContractM "unable to convert to plutus address" $
+    toPlutusAddress cslAddress
+  assertTrue "address has wrong Network Id"
```
klntsky, line 221:
```suggestion
  assertTrue "addressFromBech32: address has wrong NetworkId"
```

klntsky (changes requested):

klntsky (approved):

PR #1067, opened 2022-09-26
Add `Contract` functions to load CBOR from hex

### LINKED ISSUE #850: Add a function to load scripts from hex, without envelope
`textEnvelopeBytes` is useful, but if the user has just cborhex, it's tedious to wrap it.

MitchyCola:
Closes #850 .

### Pre-review checklist


src/Contract/Cbor.purs
```diff
@@ -0,0 +1,30 @@
...
+
+import Contract.Monad (Contract)
+import Data.Newtype (wrap)
+import Effect.Exception (error)
+import Types.ByteArray (ByteArray)
+import Types.ByteArray (hexToByteArray) as BA
+import Types.Cbor (Cbor)
+import Types.CborBytes (CborBytes)
+
+hexToByteArray :: String -> Contract () ByteArray
```
klntsky, line 17: There is no reason for this function to be impure.

klntsky (changes requested):
.

src/Contract/Cbor.purs
```diff
@@ -0,0 +1,18 @@
+module Contract.Cbor
+  ( module ByteArray
+  , module CborBytes
+  , hexToCbor
+  ) where
+
+import Contract.Prelude
+
+import Ctl.Internal.Types.ByteArray (hexToByteArray) as ByteArray
```
klntsky, line 9: `hexToByteArray` is exported from Contract.Prim.ByteArray, no need to re-export

klntsky (changes requested):

src/Contract/Cbor.purs
MitchyCola: Fixed

src/Contract/Cbor.purs
```diff
@@ -0,0 +1,16 @@
...
+  ) where
+
+import Contract.Prelude
+
+import Ctl.Internal.Types.Cbor (Cbor)
+import Ctl.Internal.Types.CborBytes (hexToCborBytes) as CborBytes
+import Data.Newtype (wrap)
+
+-- | Decode a hexadecimal string into a `Cbor`.
+-- | Results in `Nothing` if the string contains a non-hexadeciaml number,
```
Luis-omega, line 13:
```suggestion
-- | Results in `Nothing` if the string contains a non-hexadecimal number,
```
little error.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - `Ctl.Examples.PlutusV2.OneShotMinting` example, demonstrating the use of `applyArgs` with a PlutusV2 script ([#1038](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1038))
 - New `Contract.Test.Utils` assertions and checks: `assertOutputHasRefScript`, `checkOutputHasRefScript`, `checkTxHasMetadata` ([#1044](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1044))
 - `Parallel` instance to `Contract` monad. Parallel capabilities are in the associated `ParContract` datatype ([#1037](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1037))
+- `Contract.Cbor.hexToCborBytes` and `Contract.Cbor.hexToCbor` to decode a hexadecimal string to a `CborBytes` or `Cbor`, respectively. ([#850](https://github.com/Plutonomicon/cardano-transaction-lib/issues/850))
```
ngua, line 41: I'm not sure if there's a reason I'm not aware of, but is it worth it to export both of these? I think we could just export a function to work with `Cbor` and keep `CborBytes` for internal use. I think users will wonder what the difference is between the two types, then they'll need to read through the documentation/source more closely to figure it out. They can just use `unwrap` to get the `CborBytes` if that's really necessary

src/Contract/Cbor.purs
```diff
@@ -0,0 +1,16 @@
+module Contract.Cbor
```
ngua, line 1: Please add module documentation

ngua (commented):
It might also be nice to export a wrapper around `byteArrayToHex` just for completeness' sake (so `cborToHex`). That seems like a pretty common use case to me and it would save users from having to import another module and then `wrap` the result themselves

src/Contract/Cbor.purs
MitchyCola: Added

CHANGELOG.md
MitchyCola: I agree. It has been removed.

klntsky (changes requested):
@ngua `Cbor` is our internal type, unlike `CborBytes`, isn't it? I don't get why do we need to expose it at all. `TextEnvelope` is the only place where it is used, and I think there's no reason to expose `TextEnvelope` at all (it is not exposed).

I think this PR should be made for CborBytes instead.

jy14898 (approved):
LGTM, please merge develop

src/Contract/CborBytes.purs
```diff
@@ -0,0 +1,16 @@
+-- | A module with CBOR-related functionality.
+module Contract.CborBytes (module CborBytes) where
+
+import Ctl.Internal.Types.CborBytes
+  ( CborBytes(CborBytes)
+  , byteLength
```
klntsky, line 6: Let's not export functions more than once from `Contract.*` modules. This one is in `Contract.Prim.ByteArray`.

klntsky (changes requested):

src/Contract/CborBytes.purs
MitchyCola: It is not exactly the same, but it does use `Contract.Prim.ByteArray` inside the function: https://github.com/Plutonomicon/cardano-transaction-lib/blob/014e2b0119a16bd5ca936ac28d26ffeea8e9db35/src/Internal/Types/CborBytes.purs#L61-L62

We can be easily remove from the export list, as it is a simple function. Or maybe rename it to `cborByteLength`?

klntsky: OK I didn't notice it's different

klntsky: Yes, a rename would be good.

klntsky (approved):

PR #1069, opened 2022-09-27
Refactor `MintingPolicy` to sum type consisting of plutus or native scripts

### LINKED ISSUE #1056: Native Script as Minting Policy / `MintingPolicyNotFound`
I am experiencing an unexpected error when trying to mint a token using a Native Script.

For example, I am trying to mint a token using a Native Script with a TimelockExpiry condition. See example: https://github.com/codybutz/ctl-examples/tree/minting-policy-issue You can run the example with `nix build .#checks.x86_64-linux.ctl-scaffold-plutip-test -L`

You will see that we get the following error:
```
Uncaught Error: (MintingPolicyNotFound (MintingPolicyHash (ScriptHash 51ccadf8b63ad3483381586afd8ff2e140fa54d28333b99fc3fcaaf3)))
```
I believe that this is because the lookup requires that a plutus script to be included with a minting policy. However, not all minting policies are Plutus Scripts. See https://github.com/Plutonomicon/cardano-transaction-lib/blob/7921daeda855f8822b678d44fe46ffd37e0f860c/src/Types/ScriptLookups.purs#L836 and https://github.com/Plutonomicon/cardano-transaction-lib/blob/7921daeda855f8822b678d44fe46ffd37e0f860c/src/Types/ScriptLookups.purs#L995

Is it intended that all minting policies in the CTL require the use of Plutus Scripts or can we change the interface to allow for native script tokens?

amirmrad:
Closes #1056  .

### Pre-review checklist


CHANGELOG.md
```diff
@@ -38,11 +38,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
 - `Parallel` instance to `Contract` monad. Parallel capabilities are in the associated `ParContract` datatype ([#1037](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1037))
+- Corresponding `DecodeAeson` instance for `NativeScript` internal data type ([#1069](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1069)).
 
 ### Changed
 
 - Reorganised the library into new namespaces. Namely: library internals, tests, and examples are now under `Ctl.Internal.*`, `Test.Ctl.*`, and `Ctl.Examples.*` respectively. Documentation and comments have been updated to use these new names, but not entries of previous releases in the changelog. ([#1039](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1039))
 - Switched to `preview` testnet by default ([#1030](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1030))
+- `MintingPolicy` to an enum consisting of `PlutusScript` or `NativeScript` ([#1069](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1069))
+- `Serialization.NativeScript`'s `convertNativeScript` is now pure ([#1069](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1069))
```
klntsky, line 48: This function does not face the user (it's internal), so no need to mention it in the changelog.

klntsky (changes requested):
Otherwise perfect!

klntsky (approved):

klntsky (changes requested):
Wait. Just noticed there's no test

klntsky (changes requested):

src/Internal/Types/ScriptLookups.purs
```diff
@@ -1028,13 +1034,20 @@ processConstraint mpsMap osMap = do
...
-      -- Remove mint redeemers from array before reindexing.
-      _redeemersTxIns %= filter \(T.Redeemer { tag } /\ _) -> tag /= Mint
-      -- Reindex mint redeemers.
-      mintRedeemers <- lift $ reindexMintRedeemers mpsHash redeemer
-      -- Append reindexed mint redeemers to array.
-      _redeemersTxIns <>= map (_ /\ Nothing) mintRedeemers
-      _cpsToTxBody <<< _mint <>= map wrap mintVal
+
+      let mp = lookupMintingPolicy mpsHash mpsMap
+      case mp of
```
klntsky, line 1039: I think we should use `mp <- except $ lookupMintingPolicy`, because the `Left` case is an exception. It will be handled a few lines above, but still I'd prefer the code style without wildcard patterns.

klntsky (changes requested):

src/Internal/Types/ScriptLookups.purs
amirmrad: It was `except $ ...` originally, but there is a case in the `ReferenceInputs` example where the hash of the minting policy is not in `mpsMap` and so `except` will throw an error. Although, I agree that it can be refactored somehow to use the pattern match above.

klntsky: But it is possible that the minting policy is not in `mpsMap` simply because the user forgot to add it to lookups, right? And in this case we will not be able to detect that the intention was to use a `NativeScript`, so the failure will be the same as before this change. Not a great UX.

If so, all of it becomes a bit hairy. I think we should introduce another set of constraints for native scripts then, since the logic is different in non-trivial way.

If we have to provide the native script anyways for minting, let's pass it directly, without providing a hash.

So,

```
mustMintCurrencyUsingNativeScript
  :: forall (i :: Type) (o :: Type)
   . NativeScript
  -> TokenName
  -> BigInt
  -> TxConstraints i o
```

Please also add a new `TxConstraint` constructor for minting using native scripts.

My motivation:

I think we should introduce another constraint sooner, because in the future we'd like to remove lookups, and modifying `mustMintCurrency` so that it is able to handle any script will require something like:

```
mustMintCurrency
  :: forall (i :: Type) (o :: Type)
   . Either PlutusScript NativeScript
  -> TokenName
  -> BigInt
  -> TxConstraints i o
```

which is really ugly (because of the need to tag with `Either`).

test/Plutip.purs
```diff
@@ -566,6 +561,37 @@ suite = do
...
           submitAndLog bsTx
 
+    test "runPlutipContract: NativeScriptMints" do
+      let
+        distribution :: InitialUTxOs
+        distribution =
+          [ BigInt.fromInt 5_000_000
+          , BigInt.fromInt 2_000_000_000
+          ]
+      runPlutipContract config distribution \alice -> do
```
klntsky, line 571: Could you please also demonstrate moving the tokens around, e.g. sending to self?

src/Internal/Types/ScriptLookups.purs
klntsky: One more argument for separating the mint constraint into native/plutus variants: we have already done the same for `mustPayTo*Script`

amirmrad: Yes this separation makes sense,  I'll do the refactor.

src/Internal/Types/TxConstraints.purs
```diff
@@ -417,6 +420,15 @@ mustMintCurrency
 mustMintCurrency mph =
   mustMintCurrencyWithRedeemer mph unitRedeemer
 
+mustMintCurrencyUsingNativeScript
+  :: forall (i :: Type) (o :: Type)
+   . NativeScriptHash
```
klntsky, line 425: Please use `NativeScript` itself instead of the hash, because it must be provided anyway. We are going to remove lookups in the future, so it's best to implement new interfaces without the need to use lookups.

src/Internal/Types/ScriptLookups.purs
```diff
@@ -995,13 +1012,27 @@ processConstraint mpsMap osMap = do
...
     MustMintValue mpsHash red tn i scriptRefUnspentOut -> runExceptT do
-      case scriptRefUnspentOut of
+      isNative <- case scriptRefUnspentOut of
         Nothing -> do
-          plutusScript <-
-            except $ unwrap <$> lookupMintingPolicy mpsHash mpsMap
-          ExceptT $ attachToCps attachPlutusScript plutusScript
-        Just scriptRefUnspentOut' ->
-          ExceptT $ processScriptRefUnspentOut mpsHash scriptRefUnspentOut'
+          mp <- except $ lookupMintingPolicy mpsHash mpsMap
```
klntsky, line 1017: Please remove the ability to mint with native script using `MustMintValue` constraint. The users should call `MustMintValueUsingNativeScript` only (having two ways of doing things is bad)

src/Internal/Types/ScriptLookups.purs
```diff
@@ -1018,24 +1049,57 @@ processConstraint mpsMap osMap = do
...
+        mintRedeemers <- lift $ reindexMintRedeemers mpsHash redeemer
+        -- Append reindexed mint redeemers to array.
+        _redeemersTxIns <>= map (_ /\ Nothing) mintRedeemers
+
+      _cpsToTxBody <<< _mint <>= map wrap mintVal
+
+    -- TODO: amir decide whether to split MustMintValue & MustMintValueUsingNativeScript
+    MustMintValueUsingNativeScript nsHash tn i -> runExceptT do
+      let mpHash = wrap <<< unwrap $ nsHash
+      mp <- except $ lookupMintingPolicy mpHash mpsMap >>=
```
klntsky, line 1077: The code will be much simpler after the hash is replaced with the script itself.

klntsky (changes requested):

src/Internal/Types/TxConstraints.purs
amirmrad: My bad I hadn't pushed my latest changes

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - New `Contract.Transaction` functions accepting balancer constraints as a parameter: `balanceTxWithConstraints`, `balanceTxsWithConstraints`, `withBalancedTxWithConstraints`, `withBalancedTxsWithConstraints` ([#1053](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1053))
 - New functions `addressWithNetworkTagFromBech32` and `addressFromBech32` in `Contract.Address`, the second checking that address network Id corresponds to the contract environment network Id. ([#1062](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1062))
 - `Contract.CborBytes` for CBOR-related functionality. ([#850](https://github.com/Plutonomicon/cardano-transaction-lib/issues/850))
+- Corresponding `DecodeAeson` instance for `NativeScript` internal data type ([#1069](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1069)).
```
klntsky, line 45: It is not internal.
```suggestion
- `DecodeAeson` instance for `NativeScript` data type ([#1069](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1069)).
```

PR #1073, opened 2022-09-29
Organize `FromBytes` and `ToBytes`

### LINKED ISSUE #560: Improve code organization (`from_bytes` / `to_bytes`)
Currently we have too many styles of serialization/deserialization. Various modules wrap `to_bytes` and `from_bytes` call very differently.

In fact we have a uniform interface for these methods: `Serialization.ToBytes` and `Deserialization.FromBytes` which we need to use instead.

The ask:

- find all FFI uses of `to_bytes` and `from_bytes` and use the aforementioned interfaces instead of repeating the same code.

Additionally, the interface itself could be improved:

- Move `ToBytes` from `Serialization` to a new module `Serialization.ToBytes`
- make `fromBytes` and `toBytes` functions use `CborBytes`
- For `toBytes`, use `Castable` from `untagged-union` to remove the need to call `asOneOf`
- For `FromBytes`, make the type class instance only contain CSL class name as string (and move logic for dispatching based on that string to JS side). That way we can have 1 FFI function for all, instead of one for each type.

MitchyCola:
Closes #560 .

### Pre-review checklist


src/Internal/Serialization/ToBytes.purs
```diff
@@ -0,0 +1,93 @@
...
+  , TransactionWitnessSet
+  , Value
+  )
+import Ctl.Internal.Types.CborBytes (CborBytes)
+
+-- NOTE returns cbor encoding for all but hash types, for which it returns raw bytes
+foreign import _toBytes :: forall (a :: Type). a -> CborBytes
+
+class ToBytes a where
+  toBytes' :: a -> CborBytes
```
jy14898, line 33: This is not accurate for hash types. https://github.com/Emurgo/cardano-serialization-lib/issues/263

Also, the original issue did not request a class here:

> For toBytes, use Castable from untagged-union to remove the need to call asOneOf

What's the reason for the class? If we're keeping it, we need to make sure it's closed such that no other instances could be made, maybe with an instance chain and an error at the end of it.

jy14898 (changes requested):

src/Internal/Cardano/Types/Transaction.purs
```diff
@@ -288,7 +288,7 @@ instance EncodeAeson TxBody where
     , referenceInputs = encodeSet r.referenceInputs
     }
 
-newtype ScriptDataHash = ScriptDataHash ByteArray
+newtype ScriptDataHash = ScriptDataHash CborBytes
```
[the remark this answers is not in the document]
MitchyCola: I got the idea in my head that the datatypes that I changed would be always be CBOR encoded. I will change them back to keep them general.

src/Internal/Serialization/ToBytes.purs
klntsky: No need for the class. Castable would work better. Ping me if you find the Castable machinery confusing, I have a grasp of it.

rynoV:
Replaced by #1285

PR #1088, opened 2022-10-06
Improve reference scripts UX

errfrom:
### Progress
1. [x] Automatically include provided utxos with reference scripts in the utxo lookup map, so that the user doesn't need to use the `unspentOutputs` lookup for that
2. [x] Throw an error on utxo lookup failure when collecting reference Plutus scripts used to determine the language set for script integrity hash calculation (previously, these failures were ignored, sometimes leading to the cryptic `extraDataMismatch` error from Ogmios)
3. [x] Update `Examples.PlutusV2.ReferenceScripts`
4. [x] Add doc-comments for `InputWithScriptRef` and reference scripts constraints 

### Pre-review checklist


jy14898 (approved):
LGTM

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -239,7 +239,7 @@ runBalancer utxos changeAddress =
 
     case newMinFee == minFee of
       true -> do
-        finalizedTransaction <- lift $ finalizeTransaction balancedTx utxos -- TODO: all available?
+        finalizedTransaction <- finalizeTransaction balancedTx utxos -- TODO: all available?
```
klntsky, line 242: Let's convert this TODO into an issue and put the link here

klntsky (changes requested):
Otherwise perfect

src/Internal/BalanceTx/BalanceTx.purs
errfrom: @klntsky I've pushed a tiny fix for this TODO. Generally, we want to have all utxos provided when evaluating and finalizing a transaction, not just the available ones. 

Consider the following scenario: The user submits two transactions in parallel. The first transaction spends some output containing a reference script. The second one includes the same output as a reference input. In case this output is locked for the second transaction, `finalizeTransaction` will fail with `UtxoLookupFailedFor`, masking the actual problem.

klntsky (approved):

Issue #1081, opened 2022-10-04
Better docs for mustSatisfyAny


klntsky:
Make it clear that it's just a kind of declarative sugar for try-catch, and that it does not combine cases in any way.

One of our users thought that `mustSatisfyAny` combines alternatives when used with `mustBeSignedBy`.

PR #1103, opened 2022-10-11
docs for mustSatisfyAnyOf

pb99u069:
Closes # .

### Pre-review checklist

### DIFF
--- a/src/Internal/Types/TxConstraints.purs
+++ b/src/Internal/Types/TxConstraints.purs
@@ -490,6 +490,11 @@ mustHashDatum
   :: forall (i :: Type) (o :: Type). DataHash -> Datum -> TxConstraints i o
 mustHashDatum dhsh = singleton <<< MustHashDatum dhsh
 
+-- | Takes a list of constraint lists and checks
+-- | if ALL constraints in ANY of these sublists are satisfied.
+-- | If all sublists fail, the function fails.
+-- | So does not take into consideration constraints
+-- | created with other functions like e.g. mustBeSignedBy.
 mustSatisfyAnyOf
   :: forall (f :: Type -> Type) (i :: Type) (o :: Type)
    . Foldable f


src/Internal/Types/TxConstraints.purs
```diff
@@ -490,6 +490,11 @@ mustHashDatum
   :: forall (i :: Type) (o :: Type). DataHash -> Datum -> TxConstraints i o
 mustHashDatum dhsh = singleton <<< MustHashDatum dhsh
 
+-- | Takes a list of constraint lists and checks
+-- | if ALL constraints in ANY of these sublists are satisfied.
+-- | If all sublists fail, the function fails.
+-- | So does not take into consideration constraints
+-- | created with other functions like e.g. mustBeSignedBy.
```
jy14898, line 497: I prefer

```suggestion
-- | Attempts to solve, in order, a sequence of constraints until the first
-- | successful solve.
-- | If all constraints fail, this constraint will fail.
-- | NOTE: 
-- |  - Attempts are shallow, ALL constraints inside an item in this
-- |    sequence must succeed.
-- |  - Results from failed items in the sequence are discarded.
```

pb99u069: thanks, I agree. In the final note I tried to address the [issue](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1081)) someone had with the function according to Vladimir. but I guess it doesn't make much sense without knowing the issue?

klntsky: I think the note about `mustBeSignedBy` is too specific (because one of our users made this assumption). We should probably use this wording;

```
`mustSatisfyaAnyOf` is just a way to express a chain of try-catch expressions in a declarative manner. It does not do any analysis of the constraints' semantics.
```

pb99u069: So we take the description from @klntsky? much more abstract but I guess understandable for an experienced developer?

src/Internal/Types/TxConstraints.purs
```diff
@@ -490,13 +490,8 @@ mustHashDatum
...
 
--- | Attempts to solve, in order, a sequence of constraints until the first
--- | successful solve.
--- | If all items in the sequence fail, this constraint will fail.
--- | NOTE: 
--- |  - Attempts are shallow, ALL constraints inside an item in this
--- |    sequence must succeed.
--- |  - Results from failed items in the sequence are discarded.
+-- | `mustSatisfyaAnyOf` is just a way to define a chain of try-catch expressions
+-- | in a declarative manner. It does not do any analysis of the constraints' semantics.
```
klntsky, line 494:
```suggestion
-- | Attempts to solve, in order, a sequence of constraints until the first
-- | successful try.
-- | `mustSatisfyaAnyOf` is just a way to define a chain of try-catch expressions
-- | in a declarative manner. It does not do any analysis of the constraints' semantics.
```

klntsky (approved):

PR #1125, opened 2022-10-18
Handle multiple time constraints

Luis-omega:
Closes # .

### Pre-review checklist


Luis-omega:
Ther's still some test to be added an some corrections based on test results.

Luis-omega:
Also I still need to check if the `Plutus Schema`, `ToData`, `FromData` and `Aeson` instances are right.

test/BalanceTx/Time.purs
```diff
@@ -60,37 +86,43 @@ emptyLookup :: ScriptLookups Void
 emptyLookup = mempty
 
 now :: POSIXTime
-now = mkPosixTime "1666918454999"
+now = mkPosixTime "1666918454000"
+
+unsafeSubstractOne :: BigNum -> BigNum
```
klntsky, line 91:
```suggestion
unsafeSubtractOne :: BigNum -> BigNum
```

klntsky (changes requested):

src/Internal/Types/Interval.purs
```diff
@@ -232,23 +238,24 @@ derive instance Functor UpperBound
...
+derive instance Generic (PlutusInterval a) _
+derive newtype instance Eq a => Eq (PlutusInterval a)
+derive instance Functor PlutusInterval
+derive instance Newtype (PlutusInterval a) _
 
 instance
-  HasPlutusSchema (Interval a)
-    ( "Interval"
+  HasPlutusSchema (PlutusInterval a)
+    ( "PlutusInterval"
```
Luis-omega, line 258: This is the more concerning change to me, I don't really know if this is fine to do.

Luis-omega:
the failing test seems to be related to #864

test/BalanceTx/Time.purs
```diff
@@ -88,15 +87,14 @@ emptyLookup = mempty
 now :: POSIXTime
 now = mkPosixTime "1666918454000"
 
-unsafeSubstractOne :: BigNum -> BigNum
-unsafeSubstractOne value = unsafePartial $ fromJust
+unsafeSubtractOne :: BigNum -> BigNum
```
klntsky, line 90: Do not use unsafePartial here. It should be used in the caller, if ever.

test/BalanceTx/Time.purs
```diff
@@ -132,24 +130,18 @@ validityToPosixTime { validityStartInterval, ttl: timeToLive } =
         case timeToLive of
           Just end ->
             let
-              end' = substractOne end
+              end' = subtractOne end
```
klntsky, line 133: Is there a guarantee that end is > 1? Why? each unsafePartial call should come with a comment, if it's not obvious.

test/BalanceTx/Time.purs
```diff
@@ -132,24 +130,18 @@ validityToPosixTime { validityStartInterval, ttl: timeToLive } =
...
           Nothing ->
             from <$> toPosixTime start
     Nothing ->
       case timeToLive of
         Nothing -> pure always
-        Just end -> to <$> toPosixTime (substractOne end)
+        Just end -> to <$> toPosixTime (subtractOne end)
   where
-  substractOne = wrap <<< unsafeSubstractOne <<< unwrap
+  subtractOne = wrap <<< unsafeSubtractOne <<< unwrap
```
klntsky, line 144: No type annotation

klntsky (changes requested):

src/Internal/Types/Interval.purs
klntsky: No. What was the motivation for renaming?

klntsky: Why do we ever need it?

Luis-omega: If we use the name `Interval` for the new interval style and users have been using `to`, `from`, `never`, `always` and `mkInterval` instead of the old `Interval` constructor, then they only need to replace `mkInterval` with `mkFiniteInterval` to use the new type, that's why I renamed the old type to a new name `PlutusInterval`. 

Original I keep `PlutusInterval` since I don't properly understand `HasPlutusSchema`, `ToData`, and `FromData`, so It was easy to just translate a new `Interval` to `PlutusInterval` and use its instances. After some thought, `PlutusInterval` is interesting to keep as it follows the `Plutus` code and it can help to understand (and debug) the new `Interval` type. 

Of course, we still need to rewrite the instances of `HasPlutusSchema`, `ToData`, and `FromData` for the new `Interval` type to avoid using `PlutusInterval`

src/Internal/Types/Interval.purs
```diff
@@ -259,35 +266,226 @@ instance
...
+-- | A new abstraction for `Interval` on top of the old one
+-- |(PlutusInterval), it restricts the creation of intervals
+-- |to the only ones that can be used in practice:
+-- | [a,b+1) , [-infinity,b+1), [a,infinity], [-infinity,infinity],
+-- | empty.
+-- see https://github.com/Plutonomicon/cardano-transaction-lib/issues/1041
+data Interval :: Type -> Type
+data Interval a
+  = FiniteInterval a a
+  | LowerRay a
```
klntsky, line 310: Let's rename to `StartingAt` and `EndingAt`, or better names of your choice.

klntsky (changes requested):
Let's drop the old type. I fail to see why our users might need it.

We will have to keep the ToData/FromData instances the same, which means we can drop HasPlutusSchema and write them manually. They should be defined so that the old and the new versions translate to the same PlutusData. How to test it: for each interval type, write an example and convert it to PlutusData *before* the change - using old Interval type. Get the PlutusData fixtures and use them in tests, but compare them to values generated from the values of the new type, that correspond to old type value semantically.

klntsky:
@Luis-omega I think this link might be helpful to understand how to translate intervals:

https://github.com/input-output-hk/cardano-ledger/blob/2acff66e84d63a81de904e1c0de70208ff1819ea/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxInfo.hs#L206

src/Internal/Types/Interval.purs
Luis-omega: I forgot to do this before begin to add the `ToData` instance, I will do this change at the end.

Luis-omega:
> Alternatively, you can move the old type to `test/` to be able to generate values with ease, without switching branches.

I'm doing both of them. At this time I have some error in the `FromData` instance, but the `ToData` instance seems correct. 

Also I realize I still need to implement AESON instances for `Interval`.

src/Internal/Types/ScriptLookups.purs
```diff
@@ -803,6 +811,42 @@ addOwnOutput (OutputConstraint { datum: d, value }) = do
...
+resumeTimeConstraints constraints =
+  let
+    { no: nonTimeConstraints, yes: timeConstraints } = partition
+      isTimeConstraint
+      constraints
+    intervals = mapMaybe constraintToInterval timeConstraints
+  in
+    do
+      newInterval <- foldM mergeIntervals always intervals
+      pure $ cons (MustValidateIn newInterval) nonTimeConstraints
```
jy14898, line 825: It's a shame we go over the constraints twice, [`partitionMap`](https://pursuit.purescript.org/packages/purescript-filterable/5.0.0/docs/Data.Filterable#v:partitionMap) gives us what we need but it's not worth pulling in just for this

```suggestion
resumeTimeConstraints constraints = do
  let
    { no: nonTimeConstraints, yes: timeConstraints } = partition
      isTimeConstraint
      constraints
    intervals = mapMaybe constraintToInterval timeConstraints
  newInterval <- foldM mergeIntervals always intervals
  pure $ cons (MustValidateIn newInterval) nonTimeConstraints
```

test/BalanceTx/Time.purs
```diff
@@ -0,0 +1,208 @@
...
+emptyLookup :: ScriptLookups Void
+emptyLookup = mempty
+
+now :: POSIXTime
+now = mkPosixTime "1666918454000"
+
+unsafeSubtractOne :: forall (a :: Type). Partial => Newtype a BigNum => a -> a
+unsafeSubtractOne value = wrap <<< fromJust
+  $ BigNum.fromInt
+  <$> (flip (-) 1 <$> BigNum.toInt (unwrap value))
```
jy14898, line 144:
```suggestion
  <$> ((_ - 1) <$> BigNum.toInt (unwrap value))
```

jy14898 (changes requested):
Good work on this. I'm happy to see lots of tests too. My requested changes are all superficial syntax

src/Internal/Types/Interval.purs
```diff
@@ -333,53 +435,59 @@ lowerBound a = LowerBound (Finite a) true
...
+-- | represented as [a,a+1).
+singleton :: forall (a :: Type). Ord a => Semiring a => a -> Interval a
+singleton s = mkFiniteInterval s s
 
 -- | `from a` is an `Interval` that includes all values that are
--- | greater than or equal to `a`.
-from :: forall (a :: Type). a -> Interval a
-from s = mkInterval (lowerBound s) (UpperBound PosInf true)
+-- | greater than or equal to `a`, represented as [a,infinity].
+from :: forall (a :: Type). Ord a => a -> Interval a
```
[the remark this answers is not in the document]
Luis-omega: To form an Interval in a set you need at least an order. In this case, we don't need it for computations but is morally wrong to allow to form of an Interval without an Order.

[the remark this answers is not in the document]
Luis-omega: In fact, for all the functions that perform some arithmetic (`Semiring` and `Ring`), we should ask for consistency between the order and operations ( x<x+1 is true?).  But encoding this consistency it's not worth the effort for the limited set of cases we need.

src/Internal/Types/Interval.purs
```diff
@@ -232,62 +238,163 @@ derive instance Functor UpperBound
...
+                  Nothing
+                else
+                  pure $ to (endValue - one)
+              NegInf ->
+                pure $ never
+              PosInf ->
+                pure $ always
+          PosInf ->
+            pure $ never
+  fromData _ = Nothing
```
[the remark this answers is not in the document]
Luis-omega: Right, I was through in `BigNumber` when I wrote it.

CHANGELOG.md
```diff
@@ -79,6 +80,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - `awaitTxConfirmedWithTimeout` not respecting its timeout ([#1021](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1021))
 - Absence of `serializeData` in Plutip ([#1078](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1078))
 - Fix excessive reconnection attempts after `Contract` runtime finalization ([#965](https://github.com/Plutonomicon/cardano-transaction-lib/pull/965))
+- Now we can process multiple time constraints([#1124](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1124))
```
klntsky, line 83:
```suggestion
- Now we can process multiple time constraints ([#1124](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1124))
```

src/Internal/Types/Interval.purs
```diff
@@ -232,62 +238,145 @@ derive instance Functor UpperBound
...
--- | An interval of `a`s.
--- |
--- | The interval may be either closed or open at either end, meaning
--- | that the endpoints may or may not be included in the interval.
--- |
--- | The interval can also be unbounded on either side.
-newtype Interval :: Type -> Type
-newtype Interval a = Interval { from :: LowerBound a, to :: UpperBound a }
+-- | A new abstraction for `Interval` on top of the old one
+-- |(wrap), it restricts the creation of intervals
```
klntsky, line 242: > (wrap)

What?

klntsky: Please rephrase

klntsky: No need to mention that it's new

src/Internal/Types/Interval.purs
```diff
@@ -232,62 +238,145 @@ derive instance Functor UpperBound
...
-        :+ PNil
-    )
+instance (Show a, Ord a, Semiring a) => Show (Interval a) where
+  show (FiniteInterval start end) = "(FiniteInterval " <> show start <> " "
+    <> show end
+    <> ")"
+  show (StartAt end) = "(StartAt " <> show end <> ")"
+  show (EndAt start) = "(EndAt " <> show start <> ")"
+  show AlwaysInterval = "(AlwaysInterval)"
+  show EmptyInterval = "(EmptyInterval)"
```
klntsky, line 269:
```suggestion
  show AlwaysInterval = "AlwaysInterval"
  show EmptyInterval = "EmptyInterval"
```

src/Internal/Types/Interval.purs
```diff
@@ -232,62 +238,145 @@ derive instance Functor UpperBound
...
 
-instance Show a => Show (Interval a) where
-  show = genericShow
+instance Ord a => MeetSemilattice (Interval a) where
+  meet = intersection
+
+instance (ToData a, Ord a, Semiring a) => ToData (Interval a) where
```
klntsky, line 274: Please add a comment that this instance is written in plutus-compatible way

klntsky (changes requested):

src/Internal/Types/Interval.purs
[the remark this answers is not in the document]
jy14898: I suppose. I don't think that argument really works for `always` and `never` though - they're the free (in the same sense Maybe is the free monoid of semigroups). You don't need ord for them because Interval implements them for us

Also: 
```purescript
data NoOrd = NoOrd

iLied :: Interval NoOrd
iLied = from unit <#> const NoOrd
```

I don't mind if you keep it, it does help typing fail faster.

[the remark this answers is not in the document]
Luis-omega: Yes, When I introduced the `Generic` instance and tried to hide the constructors, the compiler began to warn about how it could be used to construct intervals. 

In the case of `Generic` we cannot remove it without having problems with other types. But in the case of `Functor`, I introduce it for convenience, we can just use a `mapInterval` function that won't be exported.

klntsky (approved):

PR #1135, opened 2022-10-19
WIP: Add stake constraints

klntsky:
Closes # .

### Pre-review checklist

### DIFF
--- a/src/Internal/BalanceTx/BalanceTx.purs
+++ b/src/Internal/BalanceTx/BalanceTx.purs
@@ -370,7 +370,8 @@ addTransactionInputs changeAddress utxos unbalancedTx = do
     changeValue = (unwrap txChangeOutput).amount
 
     requiredInputValue :: Value
-    requiredInputValue = nonMintedValue <> minFeeValue txBody <> changeValue
+    requiredInputValue =
+      nonMintedValue <> minFeeValue txBody <> changeValue
 
   newTxInputs <-
     except $ collectTransactionInputs txInputs utxos requiredInputValue
--- a/src/Internal/BalanceTx/BalanceTx.purs
+++ b/src/Internal/BalanceTx/BalanceTx.purs
@@ -64,11 +64,13 @@ import Ctl.Internal.BalanceTx.Types
 import Ctl.Internal.BalanceTx.Types (FinalizedTransaction(FinalizedTransaction)) as FinalizedTransaction
 import Ctl.Internal.BalanceTx.UtxoMinAda (utxoMinAdaValue)
 import Ctl.Internal.Cardano.Types.Transaction
-  ( Transaction(Transaction)
+  ( Certificate(StakeRegistration)
+  , Transaction(Transaction)
   , TransactionOutput(TransactionOutput)
   , TxBody(TxBody)
   , UtxoMap
   , _body
+  , _certs
   , _fee
   , _inputs
   , _mint
@@ -78,6 +80,7 @@ import Ctl.Internal.Cardano.Types.Transaction
 import Ctl.Internal.Cardano.Types.Value
   ( Coin(Coin)
   , Value
+  , coinToValue
   , geq
   , getNonAdaAsset
   , minus
@@ -106,8 +109,9 @@ import Ctl.Internal.Wallet (cip30Wallet)
 import Data.Array (head)
 import Data.Array as Array
 import Data.BigInt (BigInt)
+import Data.BigInt as BigInt
 import Data.Either (Either(Left, Right), hush, note)
-import Data.Foldable (foldMap, foldl, foldr)
+import Data.Foldable (fold, foldMap, foldl, foldr, sum)
 import Data.Lens.Getter ((^.))
 import Data.Lens.Setter ((%~), (.~), (?~))
 import Data.Log.Tag (tag)
@@ -141,8 +145,12 @@ balanceTxWithAddress
   -> UnattachedUnbalancedTx
   -> QueryM (Either BalanceTxError FinalizedTransaction)
 balanceTxWithAddress ownAddrs unbalancedTx = runExceptT do
+  pparams <- asks $ _.runtime >>> _.pparams
   changeAddr <- liftMaybe CouldNotGetWalletAddress $ head ownAddrs
-
+  let
+    depositValuePerCert = (unwrap pparams).stakeAddressDeposit
+    certsFee = getStakeCertsFee (unbalancedTx ^. _transaction')
+      depositValuePerCert
   utxos <- ExceptT $ traverse utxosAt ownAddrs <#>
     traverse (note CouldNotGetUtxos) -- Maybe -> Either and unwrap UtxoM
 
@@ -168,7 +176,7 @@ balanceTxWithAddress ownAddrs unbalancedTx = runExceptT do
   logTx "unbalancedCollTx" availableUtxos unbalancedCollTx
 
   -- Balance and finalize the transaction:
-  ExceptT $ runBalancer availableUtxos changeAddr
+  ExceptT $ runBalancer availableUtxos changeAddr certsFee
     (unbalancedTx # _transaction' .~ unbalancedCollTx)
   where
   unbalancedTxWithNetworkId :: QueryM Transaction
@@ -204,9 +212,10 @@ type Iteration = Int
 runBalancer
   :: UtxoMap
   -> ChangeAddress
+  -> Coin
   -> UnattachedUnbalancedTx
   -> QueryM (Either BalanceTxError FinalizedTransaction)
-runBalancer utxos changeAddress =
+runBalancer utxos changeAddress certsFee =
   runExceptT <<<
     (mainLoop one zero <=< addLovelacesToTransactionOutputs)
   where
@@ -229,6 +238,7 @@ runBalancer utxos changeAddress =
     let
       prebalancedTx =
         addTransactionChangeOutput changeAddress utxos unbalancedTxWithInputs
+          certsFee
 
     traceMainLoop "added transaction change output" "prebalancedTx"
       prebalancedTx
@@ -305,8 +315,9 @@ buildTransactionChangeOutput
   :: ChangeAddress
   -> UtxoMap
   -> UnattachedUnbalancedTx
+  -> Coin
   -> TransactionChangeOutput
-buildTransactionChangeOutput changeAddress utxos tx =
+buildTransactionChangeOutput changeAddress utxos tx certsFee =
   let
     txBody :: TxBody
     txBody = tx ^. _body'
@@ -317,7 +328,10 @@ buildTransactionChangeOutput changeAddress utxos tx =
     changeValue :: Value
     changeValue = posValue $
       (totalInputValue <> mintValue txBody)
-        `minus` (totalOutputValue txBody <> minFeeValue txBody)
+        `minus`
+          ( totalOutputValue txBody <> minFeeValue txBody <> coinToValue
+              certsFee
+          )
   in
     TransactionOutput
       { address: changeAddress
@@ -330,10 +344,27 @@ addTransactionChangeOutput
   :: ChangeAddress
   -> UtxoMap
   -> UnattachedUnbalancedTx
+  -> Coin
   -> PrebalancedTransaction
-addTransactionChangeOutput changeAddress utxos unbalancedTx =
+addTransactionChangeOutput changeAddress utxos unbalancedTx certsFee =
   PrebalancedTransaction $ unbalancedTx # _body' <<< _outputs %~
-    Array.cons (buildTransactionChangeOutput changeAddress utxos unbalancedTx)
+    Array.cons
+      (buildTransactionChangeOutput changeAddress utxos unbalancedTx certsFee)
+
+getStakeCertsFee :: Transaction -> Coin -> Coin
+getStakeCertsFee tx cost =
+  let
+    fee :: BigInt.BigInt
+    fee =
+      (tx ^. _body <<< _certs) # fold
+        >>> map
+          ( \(cert :: _) -> case cert of
+              StakeRegistration _ -> unwrap cost
+              _ -> zero
+          )
+        >>> sum
+  in
+    Coin fee
 
 -- | Selects a combination of unspent transaction outputs from the wallet's
 -- | utxo set so that the total input value is sufficient to cover all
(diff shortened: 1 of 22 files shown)


jy14898 (changes requested):
So far looking good, I'll continue reviewing tomorrow

flake.nix
```diff
@@ -9,7 +9,7 @@
 
     # for the purescript project
     ogmios = {
-      url = "github:mlabs-haskell/ogmios/9c04524d45de2c417ddda9e7ab0d587a54954c57";
+      url = "github:klntsky/ogmios/1c8cc39320903440e0e0ff525627b7cb9dc1096a";
```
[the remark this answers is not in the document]
klntsky: I didn't have the rights, I'm in progress of making a proper merge in our fork:
The PR is here
https://github.com/mlabs-haskell/ogmios/pull/20

src/Internal/Types/TxConstraints.purs
```diff
@@ -540,6 +633,66 @@ mustHashDatum
   :: forall (i :: Type) (o :: Type). DataHash -> Datum -> TxConstraints i o
 mustHashDatum dhsh = singleton <<< MustHashDatum dhsh
 
+mustRegisterStakePubKey :: forall i o. StakePubKeyHash -> TxConstraints i o
```
jy14898, line 636: Missing kinds

src/Internal/Types/TxConstraints.purs
```diff
@@ -389,7 +445,19 @@ mustPayToScript
   -> Value
   -> TxConstraints i o
 mustPayToScript vh dt dtp vl =
-  singleton (MustPayToScript vh dt dtp Nothing vl)
+  singleton (MustPayToScript vh Nothing dt dtp Nothing vl)
+    <> guard (dtp == DatumWitness) (singleton $ MustIncludeDatum dt)
+
+mustPayToScriptAddress
```
jy14898, line 451: The naming is unfortunate imo. I seem to think we'll be tackling simplifying the number/naming of constraints in the future so doesn't matter atm

test/Deserialization.purs
```diff
@@ -210,12 +210,12 @@ suite = do
...
         liftEffect $ testNativeScript nativeScriptFixture7
       -- This is here just to acknowledge the problem
-      test "too much nesting leads to recursion error" do
+      skip $ test "too much nesting leads to recursion error" do
         expectError $ do
           let
             longNativeScript =
               Array.foldr (\_ acc -> T.ScriptAny [ acc ]) nativeScriptFixture1 $
-                Array.range 0 5000
+                Array.range 0 50 -- change this to 50000
```
jy14898, line 218: Is this just testing or is there a reason for this change?

test/Plutip.purs
```diff
@@ -169,11 +167,15 @@ import Test.Spec.Runner (defaultConfig)
...
-      NetworkId.suite
+    defaultConfig { timeout = Just $ Milliseconds 70_000.0, exit = true }
+    $ do
+        suite
+        UtxoDistribution.suite
+        NetworkId.suite
+  Utils.interpretWithConfig
+    defaultConfig { timeout = Just $ Milliseconds 450_000.0, exit = true }
+    $ do
+        Staking.suite
```
jy14898, line 178: This wont work, the second tests don't get run due to `exit = true` on the first set.

exit is necessary as that's how hydra knows the job succeeded/failed.

Two options:
  - `exit = false` and have interpret return the results of the tests, and after both test sets we decide if we want to exit with an error code
  - add the staking suite as a separate check to hydra

jy14898 (changes requested):

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -290,6 +303,31 @@ addLovelacesToTransactionOutput txOutput = do
...
+getStakingBalance :: Transaction -> Coin -> Coin
+getStakingBalance tx depositLovelacesPerCert =
+  let
+    stakeDeposits :: BigInt
+    stakeDeposits =
+      (tx ^. _body <<< _certs) # fold
+        >>> map
+          ( case _ of
+              StakeRegistration _ -> unwrap depositLovelacesPerCert
+              StakeDeregistration _ -> zero - unwrap depositLovelacesPerCert
```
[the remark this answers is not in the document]
klntsky: It's not Value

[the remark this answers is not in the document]
jy14898: Well, `Coin` has an instance, so I suppose either use `negate` or `negation` depending on whether you do it before or after the unwrap

test/Deserialization.purs
klntsky: Ooops. Absolutely forgot about that. I removed the `Effect` from conversion, which makes it impossible to catch the recursion errors now. This is a very rare case, so we can ignore this.

src/Internal/Types/TxConstraints.purs
klntsky: I made it consistent with mustPayToPubKeyAddress, that is also horrible.

src/Internal/Types/ScriptLookups.purs
```diff
@@ -966,7 +1017,7 @@ processConstraint mpsMap osMap = do
       es <- lift getEraSummaries
       ss <- lift getSystemStart
       runExceptT do
-        { timeToLive, validityStartInterval } <- ExceptT $ liftEffect $
+        ({ timeToLive, validityStartInterval }) <- ExceptT $ liftEffect $
```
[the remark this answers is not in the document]
klntsky: My editor integration breaks in these cases (i get constant "parse failed" errors when trying to auto-indent).

src/Internal/Types/ScriptLookups.purs
```diff
@@ -1043,7 +1094,7 @@ processConstraint mpsMap osMap = do
               redeemer = T.Redeemer
                 { tag: Spend
                 , index: zero -- hardcoded and tweaked after balancing.
-                , data: unwrap red
+                , "data": unwrap red
```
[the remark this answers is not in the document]
klntsky: yes.

flake.nix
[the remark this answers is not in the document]
klntsky: It's still unusable. It's not that simple to just update the revision, because a few things changes just before I merged the PR in.

test/Deserialization.purs
klntsky: The reason why i removed this test is: the range call is executed even if there's a `skip` call, because it's just normal, non-effect value which gets pre-computed and bound to a variable.

src/Internal/Types/ScriptLookups.purs
```diff
@@ -1161,14 +1211,21 @@ processConstraint mpsMap osMap = do
             }
         _cpsToTxBody <<< _outputs %= Array.(:) txOut
         _valueSpentBalancesOutputs <>= provideValue amount
-    MustPayToScript vlh dat datp scriptRef plutusValue -> do
+    MustPayToScript vlh mbCredential dat datp scriptRef plutusValue -> do
```
MitchyCola, line 1214: Should `MustSpendScriptOutput` be updated as well? since we can now pay to a base address but `MustSpendScriptOutput` still expects an enterprise address.

Could at least add a Plutip test to see if we can spend from a base address.

klntsky: Yep, sounds right.

flake.nix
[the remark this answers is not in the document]
klntsky: There's hope:
https://github.com/Plutonomicon/cardano-transaction-lib/tree/gergely/CHaP

src/Internal/Types/ScriptLookups.purs
klntsky: I updated `MustSpendScriptOutput` handler to also take base addresses into account.

test/Plutip/Staking.purs
```diff
@@ -0,0 +1,671 @@
...
+
+    test "Plutus Stake script: delegate to existing pool & withdraw rewards" do
+      let
+        distribution = withStakeKey privateStakeKey
+          [ BigInt.fromInt 1_000_000_000 * BigInt.fromInt 1_000
+          , BigInt.fromInt 2_000_000_000 * BigInt.fromInt 1_000
+          ]
+      runPlutipContract config distribution \alice ->
+        withKeyWallet alice do
+          pure unit
```
jy14898, line 356:
```suggestion
```

test/Plutip/Staking.purs
```diff
@@ -0,0 +1,671 @@
...
+
+        -- List pools: the pool must not appear in the list
+        do
+          pools <- getPoolIds
+          logInfo' "Pool IDs:"
+          logInfo' $ show pools
+          for_ pools \poolId -> do
+            logInfo' "Pool parameters"
+            logInfo' <<< show =<< getPoolParameters poolId
+          pools `shouldSatisfy` (not <<< Array.elem poolOperator)
```
jy14898, line 346:
```suggestion
          pools `shouldSatisfy` Array.notElem poolOperator
```

test/Plutip/Staking.purs
```diff
@@ -0,0 +1,671 @@
...
+          -- (usually) significantly longer.
+          do
+            { rewards: rewardsAfter } <- liftedM "Unable to get rewards" $
+              getValidatorHashDelegationsAndRewards stakeValidatorHash
+            rewardsAfter `shouldSatisfy` \after -> after < rewardsBefore
+
+    test "Native Stake script: delegate to existing pool & withdraw rewards" do
+      let
+        distribution =
+          withStakeKey privateStakeKey
```
jy14898, line 462:
```suggestion
```

If she has the same stake key as bob, it's no longer just bob who can withdraw against the validator?

jy14898 (changes requested):

jy14898:
I've gone ahead and added the relevant ogmios aeson tests to this PR too. Hope that's okay (they all pass ⭐)

klntsky:
Thank you @jy14898 that's a huge help

jy14898:
Has the change in easy-purescript-nix changed spagos behaviour? `nix run .#hydraJobs.x86_64-linux.ctl-example-bundle-web` is failing due to spago reaching out to the network

src/Internal/Address.purs
```diff
@@ -41,26 +37,25 @@ addressToOgmiosAddress :: Address -> Ogmios.OgmiosAddress
...
+addressScriptHash addr =
+  baseAddressScriptHash addr <|> enterpriseAddressScriptHash addr
 
--- | Get the `StakeValidatorHash` with an internal `Address`
-enterpriseAddressStakeValidatorHash :: Address -> Maybe StakeValidatorHash
-enterpriseAddressStakeValidatorHash =
-  map StakeValidatorHash <<< enterpriseAddressScriptHash
+-- | Get the `ValidatorHash` of an address (base or enterprise).
+addressValidatorHash :: Address -> Maybe ValidatorHash
+addressValidatorHash = map ValidatorHash <<< addressScriptHash
```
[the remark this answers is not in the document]
klntsky: There functions don't make much sense, because they use payment component scripts.

Come to think of it, the current behavior could also be very confusing. I definitely want to give it a second thought.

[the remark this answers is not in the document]
klntsky: There was a bug

jy14898 (commented):
Other than the comment I'm happy with this.

However, it looks like upgrading the package set broke Medea? A test that used to pass no longer does. I don't think upgrading the package set will fix the issue with spago, I think they broke no-install/no-build in the recent changes it seems.

examples/AlwaysSucceeds.purs
```diff
@@ -62,13 +60,24 @@ example cfg = launchAff_ do
...
+            Constraints.DatumWitness
+            $ Value.lovelaceValueOf
+            $ BigInt.fromInt 2_000_000
+        Just stakeKeyHash ->
+          Constraints.mustPayToScriptAddress vhash
+            (PubKeyCredential $ unwrap stakeKeyHash)
+            unitDatum
+            Constraints.DatumWitness
+            $ Value.lovelaceValueOf
+            $ BigInt.fromInt 2_000_000
```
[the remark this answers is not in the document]
klntsky: I think mine is more readable, because the cases are separated more clearly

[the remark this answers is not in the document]
klntsky: The problem with easy-purescript-nix upgrade is not so trivial, we seemingly can't just revert it, because we have `nixpkgs.follows = "ogmios/nixpkgs";`, and that version does not have `nodejs_12x`...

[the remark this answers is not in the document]
jy14898: This seems to work, we'll need to add a TODO issue here to drop this overlay once it's fixed in later versions (we should tell spago)
```diff
diff --git a/flake.nix b/flake.nix
index 26da7e38f..b9b37348f 100644
--- a/flake.nix
+++ b/flake.nix
@@ -220,6 +220,22 @@
           easy-ps = import inputs.easy-purescript-nix { pkgs = final; };
           purescriptProject = import ./nix { pkgs = final; };
         };
+        spago = final: prev: {
+          easy-ps = prev.easy-ps // {
+            spago = prev.easy-ps.spago.overrideAttrs (_: rec {
+              version = "0.20.7";
+              src = if final.stdenv.isDarwin
+                then final.fetchurl {
+                  url = "https://github.com/purescript/spago/releases/download/${version}/macOS.tar.gz";
+                  sha256 = "0s5zgz4kqglsavyh7h70zmn16vayg30alp42w3nx0zwaqkp79xla";
+                }
+                else final.fetchurl {
+                  url = "https://github.com/purescript/spago/releases/download/${version}/Linux.tar.gz";
+                  sha256 = "0bh15dr1fg306kifqipnakv3rxab7hjfpcfzabw7vmg0gsfx8xka";
+                };
+            });
+          };
+        };
         # This is separate from the `runtime` overlay below because it is
         # optional (it's only required if using CTL's `applyArgs` effect).
         # Including it by default in the `overlays.runtime` also requires that
```

packages.dhall
```diff
@@ -105,8 +105,8 @@ in  upstream
 -------------------------------
 -}
 let upstream =
-      https://github.com/purescript/package-sets/releases/download/psc-0.14.5-20211116/packages.dhall
-        sha256:7ba810597a275e43c83411d2ab0d4b3c54d0b551436f4b1632e9ff3eb62e327a
+      https://github.com/purescript/package-sets/releases/download/psc-0.14.5-20220224/packages.dhall
+        sha256:67cc3d4f0e8fb72bb1413ba94ddd72a3ceb0783eb725e3b22ad7568b3b581163
```
jy14898, line 109: I think we should revert this change, but in case we keep it I wanted to say we should make sure the templates upstream is set to the same

jy14898: I went through our spago-packages.nix and the diff between the two package sets, these are the only differences we can observe:
```diff
   , repo = "https://github.com/purescript/purescript-free.git"
-  , version = "v6.1.0"
+  , version = "v6.2.0"
   , repo = "https://github.com/purescript-node/purescript-node-fs.git"
-  , version = "v6.1.0"
+  , version = "v6.2.0"
   , repo = "https://github.com/purescript-contrib/purescript-parsing.git"
-  , version = "v7.0.1"
+  , version = "v8.2.0"
   , repo = "https://github.com/purescript-contrib/purescript-unicode.git"
-  , version = "v5.0.0"
+  , version = "v5.0.1"
```

klntsky: It was purescript-parsing. Downgrading

jy14898: Thought so - annoyingly there was no breaking changes between those versions, only bug fixes. So it's odd that something broke

klntsky: I think the culprit is 

>     Parser.String.eof Set consumed on success so that this parser combines correctly with notFollowedBy eof. Added a test for this. (#140 by @jamesdbrock)

https://github.com/purescript-contrib/purescript-parsing/blob/main/CHANGELOG.md

jy14898 (approved):
LGTM once CI tests complete

PR #1144, opened 2022-10-26
Cleanup temporary postgres directory in plutip

### LINKED ISSUE #1140: Plutip: clean up temporary directory
Currently there's no cleanup process

pb99u069:
Closes #1140.

### Pre-review checklist


src/Internal/Plutip/Server.purs
```diff
@@ -395,6 +401,12 @@ startPostgresServer pgConfig _ = do
     ]
     defaultSpawnOptions
   liftEffect $ killOnExit pgChildProcess
+  liftEffect $ onExit pgChildProcess \_ ->
```
pb99u069, line 404: not sure what goes wrong here. If I try to remove `databaseDir` directly (commented out line 409) I get an `ENOTEMPTY` error and if I try to delete files first like here, I get an `ENOENT` error, meaning there are no files. Is it because I need to run the operation in `Aff`? tried to do this but haven't succeeded yet..

src/Internal/Plutip/Server.js
```diff
@@ -0,0 +1,7 @@
+const fs = require("fs").promises;
```
klntsky: Please use the sync version (without the use of promises)

src/Internal/Plutip/Server.purs
```diff
@@ -447,11 +444,11 @@ startPostgresServer pgConfig _ = do
     defaultExecSyncOptions
   pure pgChildProcess
 
-delDir :: FilePath -> Aff Unit
-delDir dir = do
-  files <- A.readdir dir
-  liftEffect $ foreachE files unlink
-  A.rmdir dir
+-- delDir :: FilePath -> Aff Unit
+-- delDir dir = do
```
klntsky: A leftover

klntsky (changes requested):

src/Internal/Plutip/Server.purs
pb99u069: same as above

src/Internal/Plutip/Spawn.purs
```diff
@@ -88,3 +84,20 @@ killOnExit child = do
...
+-- | is still running.
+killOnExitAndRemDir :: ChildProcess -> FilePath -> FilePath -> Effect Unit
+killOnExitAndRemDir child dir emptyDir = do
+  aliveRef <- Ref.new true
+  ChildProcess.onExit child \_ -> do
+    Ref.write false aliveRef
+    _rmdirSync dir
+  Process.onExit \_ -> do
+    alive <- Ref.read aliveRef
+    when (dir /= emptyDir) do
```
klntsky, line 100: What is emptyDir and why can't this check be moved out of this function?

klntsky (changes requested):

src/Internal/Plutip/Spawn.js
```diff
@@ -1,3 +1,11 @@
 const fs = require("fs");
+const path = require("path");
 
 exports._rmdirSync = path => () => fs.rmdirSync(path, { recursive: true });
+
+exports._removeDirs = dirPath => () => {
+  const dirs = fs.readdirSync(dirPath);
```
klntsky, line 7: Why don't you remove the top-level dir with `recursive: true`?

klntsky (changes requested):

src/Internal/Plutip/Spawn.js
pb99u069: the problem is that I don't know the name of the test-cluster folder (respec. the six digits after test-cluster) as it's created on the plutip side. 
also, calling _rmdirSync inside Process.onExit still doesn't cleanup when we exit with ctrl+c (contrary to what I wrote yesterday).
so my solution is to just delete everything inside the tmpDir on exiting a child process and if Plutip tests are aborted with ctrl+c it gets cleaned up on the next run. as far as I can see there is nothing inside the tmpDir that has to be conserved?

pb99u069: thanks, that's what I was looking for. there is still a problem with `Process.onExit` though. I try to remove the two directories there but if the process is stopped with ctrl+c nothing happens. I tried to use `Process.onSignal SIGINT` without success

src/Internal/Plutip/Spawn.purs
```diff
@@ -88,3 +84,20 @@ killOnExit child = do
...
+killOnExitAndRemDir child workingDir testClusterDir = do
+  aliveRef <- Ref.new true
+  ChildProcess.onExit child \_ -> do
+    Ref.write false aliveRef
+    _rmdirSync workingDir
+  Process.onExit \_ -> do
+    alive <- Ref.read aliveRef
+    _rmdirSync testClusterDir
+    _rmdirSync workingDir
+    when alive do
```
klntsky, line 102: This is not needed. It will be killed anyway.

src/Internal/Plutip/Spawn.purs
```diff
@@ -88,3 +84,20 @@ killOnExit child = do
     alive <- Ref.read aliveRef
     when alive do
       kill SIGINT child
+
+foreign import _rmdirSync :: FilePath -> Effect Unit
+
+-- foreign import _removeDirs :: FilePath -> Effect Unit
```
klntsky, line 90: Please remove this leftover

src/Internal/Plutip/Spawn.js
```diff
@@ -0,0 +1,11 @@
+const fs = require("fs");
+//const path = require("path");
+
+exports._rmdirSync = path => () => fs.rmdirSync(path, { recursive: true });
+
+// exports._removeDirs = dirPath => () => {
```
klntsky, line 6: Please remove the leftovers

klntsky (changes requested):

src/Internal/Plutip/Spawn.purs
pb99u069: but that's all that `killOnExit` (used 4 times in `Plutip.Server`) does. does that mean that we should remove all those effects?

src/Internal/Plutip/Spawn.purs
```diff
@@ -88,3 +84,13 @@ killOnExit child = do
...
       kill SIGINT child
+
+foreign import _rmdirSync :: FilePath -> Effect Unit
+
+killOnExitAndRemDir :: ChildProcess -> FilePath -> FilePath -> Effect Unit
+killOnExitAndRemDir child workingDir testClusterDir = do
+  ChildProcess.onExit child \_ -> do
+    _rmdirSync workingDir
+  Process.onExit \_ -> do
+    _rmdirSync testClusterDir
```
pb99u069, line 95: this doesn't do anything at the moment I think. at least it doesn't get triggered by a ctrl+c exit

src/Internal/Plutip/Spawn.purs
```diff
@@ -88,3 +84,13 @@ killOnExit child = do
...
+    _rmdirSync workingDir
```
pb99u069, line 96: these effects don't do anything at the moment I think. at least they don't get triggered by a ctrl+c exit

src/Internal/Plutip/Spawn.purs
```diff
@@ -88,3 +84,13 @@ killOnExit child = do
     alive <- Ref.read aliveRef
     when alive do
       kill SIGINT child
+
+foreign import _rmdirSync :: FilePath -> Effect Unit
+
+killOnExitAndRemDir :: ChildProcess -> FilePath -> FilePath -> Effect Unit
+killOnExitAndRemDir child workingDir testClusterDir = do
```
klntsky, line 91: Now this function should be renamed

klntsky (changes requested):

src/Internal/Plutip/Spawn.purs
```diff
@@ -87,10 +87,13 @@ killOnExit child = do
 
 foreign import _rmdirSync :: FilePath -> Effect Unit
 
-killOnExitAndRemDir :: ChildProcess -> FilePath -> FilePath -> Effect Unit
-killOnExitAndRemDir child workingDir testClusterDir = do
+cleanupTmpDir :: ChildProcess -> FilePath -> FilePath -> Effect Unit
+cleanupTmpDir child workingDir testClusterDir = do
   ChildProcess.onExit child \_ -> do
     _rmdirSync workingDir
+  -- TODO
```
klntsky, line 94: Please put a link to the new issue alongside the TODO

klntsky (changes requested):

klntsky (approved):

PR #1150, opened 2022-10-26
Cache ProtocolParameters during balancing

jy14898:
Requesting and parsing protocol params is quite slow, so this PR improves speed (locally Pkh plutip test takes 5 seconds less, 17s -> 12s)

### Pre-review checklist


klntsky (changes requested):
No need to pass them around. They are available in QueryM - `asks $ _.runtime >>> _.pparams` or something like that.

klntsky:
Aeson parsing performance has been bothering me as well, we should probably pay more attention to it - https://github.com/Plutonomicon/cardano-transaction-lib/issues/1151

jy14898:
> No need to pass them around. They are available in QueryM - `asks $ _.runtime >>> _.pparams` or something like that.

It looks like this was originally the case and it got lost at some point :+1:

klntsky:
Let's replace `getProtocolParameters` in `QueryM` with just an `asks` call then.

src/Internal/QueryM/ProtocolParameters.purs
```diff
@@ -1,16 +1,23 @@
...
+import Control.Monad.Reader.Class (asks)
 import Ctl.Internal.QueryM (QueryM, mkOgmiosRequest)
 import Ctl.Internal.QueryM (getProtocolParametersAff) as GetProtocolParametersAff
 import Ctl.Internal.QueryM.Ogmios
   ( ProtocolParameters
   , queryProtocolParametersCall
   )
-import Prelude (unit)
+
+askProtocolParameters :: QueryM ProtocolParameters
```
klntsky, line 18: Let's replace `getProtocolParameters` with this implementation. We don't support protocol parameter changes during application runtime anyway (because they can change e.g. during balancing - extremely hard to take into account, but also very rare).

klntsky (changes requested):

test/AffInterface.purs
```diff
@@ -16,7 +16,6 @@ import Ctl.Internal.QueryM
 import Ctl.Internal.QueryM.CurrentEpoch (getCurrentEpoch)
 import Ctl.Internal.QueryM.EraSummaries (getEraSummaries)
 import Ctl.Internal.QueryM.Ogmios (OgmiosAddress)
-import Ctl.Internal.QueryM.ProtocolParameters (getProtocolParameters)
```
jy14898, line 19: No point in testing `getProtocolParameterAff, it's necessarily called to setup a contract, and getProtocolParameter is just asks, which doesn't need testing in the way we do here

test/AffInterface.purs
```diff
@@ -55,7 +54,6 @@ suite = do
     test "Get ChainTip" testGetChainTip
     test "Get waitUntilSlot" testWaitUntilSlot
     test "Get EraSummaries" testGetEraSummaries
-    test "Get ProtocolParameters" testGetProtocolParameters
```
jy14898, line 58: No point in testing `getProtocolParameterAff, it's necessarily called to setup a contract, and `getProtocolParameter` is just asks, which doesn't need testing in the way we do here

klntsky (approved):

PR #1169, opened 2022-11-01
E2E Allow for fetching necessary assets by adding url options

### LINKED ISSUE #1160: Do not store E2E wallets & settings in the repo
I noticed that the repo is getting heavier because we store wallet extension binaries and the settings archive as-is - it is especially worrying when using CTL as dependency. Initially I thought we are not going to update them too often, but now it's becoming noticeable.

Maybe we can prioritize moving them off the repo sooner rather than later?

I see two ways to resolve this problem:

-  a separate repo included as a git submodule - spago does not clone submodules, which is good for us. Downside: still keeping all versions and downloading them, even though only one is needed
-  a separate empty repo, with binaries kept in releases + a script to download one of the releases. Upside: we'll only download what we need. Downside: we have to implement a script to fetch the binaries.

amirmrad:
Closes #1160  .

### Pre-review checklist


test/e2e.env
```diff
@@ -32,28 +41,37 @@ if [[ "$(which chromium)" =~ "/snap/bin/" ]]; then
...
-# export NAMI_EXTID=lpfcbjknijpeeillifnkikgncikgfhdo
+# export NAMI_CRX="test-data/chrome-extensions/nami_3.4.1_0.crx"
+export NAMI_CRX_URL="https://github.com/mlabs-haskell/ctl-e2e-assets/releases/download/Latest/nami_3.4.1_0.crx"
+export NAMI_PASSWORD=ctlctlctl
+export NAMI_EXTID=lpfcbjknijpeeillifnkikgncikgfhdo
 
 # export FLINT_CRX="test-data/chrome-extensions/flint_1.16.2_0.crx"
-# export FLINT_PASSWORD=VZVfu5rp1r
-# export FLINT_EXTID=hnhobjmcibchnmglfbldbfabcgaknlkj
+export FLINT_CRX_URL="https://github.com/mlabs-haskell/ctl-e2e-assets/releases/download/Latest/flint_1.16.2_0.crx"
```
klntsky, line 70: We can't identify the files by URLs alone. We need to allow to pass files.

If the file is not present, then we should use the URL to download the remote file to the file location.

Otherwise, just use the file. We should make the URLs optional, but not the files.

klntsky (changes requested):

src/Internal/Test/E2E/Options.purs
```diff
@@ -174,126 +183,78 @@ browserOptionsParser = ado
...
+  crx <- option (Just <$> str) $ fold
+    [ long $ formattedWallet <> "-crx"
+    , metavar "FILE"
+    , help $ wallet <> " wallet extension (.crx) file"
+    , value Nothing
+    ]
+  crxUrl <- option (Just <$> str) $ fold
+    [ long $ formattedWallet <> "-crx-url"
+    , metavar "URL"
+    , help $ wallet <> " wallet extension (.crx) url"
```
klntsky, line 251:
```suggestion
    , help $ wallet <> " wallet extension (.crx) url to download from if the file is not present"
```

src/Internal/Test/E2E/Options.purs
```diff
@@ -331,14 +292,18 @@ testUrlsOptionParser =
...
-  ]
+chromeUserDataOptionParser
+  :: Parser (Maybe ChromeUserDataDir)
+chromeUserDataOptionParser = ado
+  dataDir <- option (Just <$> str) $ fold
+    [ long "chrome-user-data"
+    , help "Chrome/-ium user data dir"
+    , value Nothing
+    , showDefaultWith show
+    -- $ const "E2E_CHROME_USER_DATA"
```
klntsky, line 303: Please restore this line.

src/Internal/Test/E2E/Options.purs
```diff
@@ -350,7 +315,20 @@ settingsOptionsParser = ado
     , showDefaultWith show
     , metavar "DIR"
     ]
-  in { chromeUserDataDir, settingsArchive }
+  settingsArchiveUrl <- option (Just <$> str) $ fold
+    [ long "settings-archive-url"
+    , help
+        "Settings archive (.tar.gz) url that will be used to store the settings"
+    , value Nothing
+    , showDefaultWith show
```
klntsky, line 323: equivalent to just `showDefault`

src/Internal/Test/E2E/Runner.purs
```diff
@@ -647,6 +725,25 @@ spawnAndCollectOutput
...
+    { method = Left GET
+    , url = url
+    , responseFormat = Affjax.ResponseFormat.arrayBuffer
+    , headers = []
+    }
+
+  case eRes of
+    Left err -> do
+      liftEffect $ log $ "HTTP Request error: " <> printError err
+      pure unit
```
klntsky, line 741:
```suggestion
```

src/Internal/Test/E2E/Runner.purs
```diff
@@ -647,6 +725,25 @@ spawnAndCollectOutput
...
+    Right res -> do
+      buf <- liftEffect $ fromArrayBuffer res.body
+      writeFile filePath buf
+      pure unit
```
klntsky, line 745:
```suggestion
```

src/Internal/Test/E2E/Runner.purs
```diff
@@ -478,21 +539,38 @@ readExtensionParams extensionName mbCliOptions = do
...
+  toExtensionParams { crxFile, password, extensionId, crxUrl } =
     case crxFile, password, extensionId of
       Nothing, Nothing, Nothing -> pure Nothing
-      Just crx, Just pwd, Just extId -> pure $ Just
-        { crx, password: pwd, extensionId: extId }
-      _, _, _ -> throw $ "Please ensure that either none or all of"
+      Just crx, Just pwd, Just extId -> do
+        doesExist <- exists crx
+        unless doesExist $ do
+          liftEffect $ log $ crx <> " does not exist."
```
klntsky, line 552: Do not log this line here. The test suite should not be too verbose. It's a normal part of the workflow, no need to tell the user anything.

src/Internal/Test/E2E/Runner.purs
```diff
@@ -478,21 +539,38 @@ readExtensionParams extensionName mbCliOptions = do
...
+            maybe
+              ( liftedM
+                  ( error $ "Please specify  --" <> toLower extensionName
+                      <> "-crx-url or ensure"
+                      <> toUpper extensionName
+                      <> "_CRX_URL is set"
+                  ) $ liftEffect $ lookupEnv $ extensionName <> "_CRX_URL"
+              )
+              pure $ crxUrl
+          liftEffect $ log $ "Downloading from " <> crxFileUrl
```
klntsky, line 563: "Downloading URL to FILE..."

.gitignore
```diff
@@ -22,6 +22,6 @@ result-*
 output.js
 .DS_Store
 .idea/
-test-data/chrome-user-data/
+test-data
```
klntsky, line 25: Not like that. We should add a `.gitkeep` in test-data to make sure the user does not have to create the directory manually.

klntsky: Same goes for the template.

klntsky (changes requested):

src/Internal/Test/E2E/Options.purs
```diff
@@ -248,7 +249,8 @@ parseWallet wallet = ado
   crxUrl <- option (Just <$> str) $ fold
     [ long $ formattedWallet <> "-crx-url"
     , metavar "URL"
-    , help $ wallet <> " wallet extension (.crx) url"
+    , help $ wallet <>
+        " wallet extension (.crx) url to download from if the file is not present"
```
klntsky, line 253:
```suggestion
        " wallet extension (.crx) URL to download from if the file is not present"
```

src/Internal/Test/E2E/Wallets.purs
```diff
@@ -169,10 +169,13 @@ namiConfirmAccess extId re =
```
klntsky, line 169: inWalletPageOptional must be used in every `*confirmAccess` to account for the case where the access is already granted.

klntsky: See how it is used for eternl

klntsky (changes requested):

src/Internal/Test/E2E/Wallets.purs
```diff
@@ -169,10 +169,13 @@ namiConfirmAccess extId re =
 
 namiSign :: ExtensionId -> WalletPassword -> RunningE2ETest -> Aff Unit
 namiSign extId wpassword re = do
-  inWalletPage (Pattern $ unExtensionId extId) re signTimeout \nami -> do
-    clickButton "Sign" nami
-    reactSetValue (Selector ":password") wpassword nami
-    clickButton "Confirm" nami
+  inWalletPage (Pattern $ unExtensionId extId) re signTimeout \page -> do
+    clickButton "Sign" page
```
klntsky, line 173: We have to Toppokki.pageWaitForSelector for the button too. It does not work on my side.

klntsky (approved):
All good. Although it does not work for me for some reason, it breaks only at a later stage, after the asset downloader finishes.

PR #1171, opened 2022-11-02
Add whitespace checks #1162

### LINKED ISSUE #1162: Add whitespace checks to syntax checks
We should add a formatting check that disallows:

- whitespaces at the end-of-line: `grep -rn ' $'`
- extra newlines at the end-of-file (only one should be present).

These should go to the makefile and included as dependencies into the `check-format` command

uhbif19:
Closes #1162.

New line checks do not check json/skeys - not sure about is that is right.

### Pre-review checklist


whitespace_checks.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
klntsky, line 1: Please don't use `/bin/bash` in scripts. Use `/usr/bin/env bash`, it's more portable, e.g. works on nixos and rare non-linux unixes.

whitespace_checks.sh
```diff
@@ -0,0 +1,28 @@
...
+    echo "No trailing spaces"
+else
+    echo "Has trailing spaces in files: "
+    echo $trailing_whitespaces
+    exit 1
+fi
+
+# Check newline on end of a file
+
+no_newline=$(find . -not -path './.git/*' -not -path './.spago/*' -not -path './output/*' -not -path '*.json' -not -path '*dist-newstyle*' -type f -exec sh -c 'file -b "{}" | grep -q text' \; -exec sh -c '[ "$(tail -c 1 "{}" | od -An -a | tr -d "[:space:]")" != "nl" ]' \; -print)
```
klntsky, line 18: `node_modules`, `.node`, `generated-docs`, `dist`, `test-data`, `.spago2nix` should be excluded as well

klntsky (changes requested):

uhbif19:
@klntsky Nothing in CHANGELOG, cuz I guess it is only for user-visible changes.

whitespace_checks.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+# Check trailing whitespaces
+
+trailing_whitespaces=$(grep -I '[[:blank:]]$' -Rl . | grep -v -E '.*(spago|.git|dist-newstyle).*')
```
klntsky, line 5: Need more exclusion patterns here as well.

whitespace_checks.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+# Check trailing whitespaces
+
+trailing_whitespaces=$(grep -I '[[:blank:]]$' -Rl . | grep -v -E '.*(spago|.git|dist-newstyle|node_modules|.node|generated-docs|dist|test-data|.spago2nix).*')
```
klntsky, line 5: We should omit looking into the directories in the first command, instead of letting it grep everything and then filter.

Also, please add `tmp` to the list of excluded dirs, it's used in E2E

klntsky (approved):

PR #1070, opened 2022-09-27
add nixos module

brainrake:
Closes # .

### Pre-review checklist

- [x] All code has been formatted using our config (`make format`)
- [x] Any new API features or modification of existing behavior is covered as defined in our [test plan](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/test-plan.md)
- [x] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

PR #1194, opened 2022-11-08
NixOS module for server

aciceri:
I've close #1070 and re-opened it here using @brainrake branch since he was working on a different repo and then CI wasn't working.

### Pre-review checklist

### DIFF
--- a/scripts/whitespace-check.sh
+++ b/scripts/whitespace-check.sh
@@ -3,22 +3,19 @@
 # Check trailing whitespaces
 
 trailing_whitespaces=$(grep -I '[[:blank:]]$' -R \
-    $(find . \
-        -not -path '.' \
-        -not -path './tmp*' \
-        -not -path './.git*' \
-        -not -path './.spago*' \
-        -not -path './output*' \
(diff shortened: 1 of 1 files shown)


klntsky (approved):

nix/ctl-server-nixos-module.nix
```diff
@@ -0,0 +1,87 @@
+{ config, lib, pkgs, ... }:
```
klntsky, line 1: Can we have some CI check to ensure this code will be up to date?

aciceri: What do you mean?

klntsky: I mean, right now it's not actually executed in CI, right?

aciceri: It is, this PR also adds a new `nixosConfiguration` called `test` which is a minimal VM that makes use of the new NixOS modules. Moreover this `nixosConfiguration` is also added to `hydraJobs`.

PR #1195, opened 2022-11-08
format-check with include pattern

### LINKED ISSUE #1192: Improve format checking scripts: use inclusion-patterns instead of exclusion for source paths
What [has once been seen as a good idea](https://github.com/Plutonomicon/cardano-transaction-lib/blob/0d2fe4d6a77d568f486773071259dce5558217a8/scripts/whitespace-check.sh#L7), has shown its downsides in practice: sometimes the developers have their own temporary/leftover directories with logs, etc, so this script takes a really long time to traverse the directory tree.

We would like to replace the exclusion patterns with inclusion ones, e.g. `-path ./src -path ./examples` and so on. this should apply to both functions.

pb99u069:
Closes #1192.

### Pre-review checklist

### DIFF
--- a/scripts/whitespace-check.sh
+++ b/scripts/whitespace-check.sh
@@ -3,22 +3,19 @@
 # Check trailing whitespaces
 
 trailing_whitespaces=$(grep -I '[[:blank:]]$' -R \
-    $(find . \
-        -not -path '.' \
-        -not -path './tmp*' \
-        -not -path './.git*' \
-        -not -path './.spago*' \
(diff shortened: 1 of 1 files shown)


pb99u069:
The `trailing whitespaces` check is much faster now, but the `newline on end of file` check seems to be slower than before. Can we maybe exclude the check on some dir/files that I included?

klntsky (approved):

klntsky (approved):

PR #1066, opened 2022-09-25
Add remaining CIP-30 methods

Luis-omega:
Closes #974 .

### Pre-review checklist

- [x] All code has been formatted using our config (`make format`)
- [x] Any new API features or modification of existing behavior is covered as defined in our [test plan](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/test-plan.md)
- [x] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

PR #1198, opened 2022-11-09
Change Contract interfaces to support multi-address

### LINKED ISSUE #1045: Change Contract interfaces to support multi-address
Initially our interfaces were designed for single-address wallets, but with addition of Eternl we need to pass all the addresses we get from the wallets.

Luis-omega:
Closes #1045 

### Pre-review checklist

### DIFF
--- a/examples/ContractTestUtils.purs
+++ b/examples/ContractTestUtils.purs
@@ -53,7 +53,11 @@ import Contract.TxConstraints as Constraints
 import Contract.Utxos (utxosAt)
 import Contract.Value (CurrencySymbol, TokenName, Value)
 import Contract.Value (lovelaceValueOf, singleton) as Value
-import Ctl.Examples.Helpers (mustPayToPubKeyStakeAddress) as Helpers
+import Ctl.Examples.Helpers
+  ( liftedHead
+  , maybeArrayToHead
+  , mustPayToPubKeyStakeAddress
+  ) as Helpers
 import Data.BigInt (BigInt)
 import Data.Lens (view)
 import Data.Map (empty) as Map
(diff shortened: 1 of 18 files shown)


Luis-omega:
Most of our tests are written to use only one address, I just adapted all of them to get the head of the returned array instead. I don't think we need more tests for this, right?

examples/ContractTestUtils.purs
```diff
@@ -84,7 +88,7 @@ mkAssertions
        )
 mkAssertions params@(ContractParams p) = do
   senderAddress <-
-    liftedM "Failed to get sender address" getWalletAddress
+    Helpers.liftedHead "Failed to get sender address" getWalletAddress
```
klntsky, line 91: I'm against using these helpers. The code in the examples must be copy-paste-able, but we don't have these helpers in the public interface.

Some of them are re-exported from prelude, but not all - perhaps we should add these missing ones there.

klntsky (changes requested):

examples/ContractTestUtils.purs
Luis-omega: I defined them in `examples/Helpers.purs`, but since I'm also using them in the tests, moving them to `Interanal/Helpers.purs` sounds better.

klntsky (changes requested):
- New functions should be added with different name (adding `-s`/`-es`). No `Maybe`s are necessary in return types
- The old functions that work with singular obects should be implemented using the new ones (a `head` call). A deprecation warning should be added to them, using `Warn` constraint. The warning should be something like: `This function returns only one X even in case multiple Xes are available. Use getXes instead`

Luis-omega:
After the last comment and looking at the changes in #1066 , it's better to redo it in #1211

Issue #1080, opened 2022-10-04
Complete CTL Test Plan Coverage


MitchyCola:
## CTL Test Plan Progress
This issue will be used to track the progress of the CTL Test Plan coverage.

Please link PRs when adding coverages. Make sure to update [doc/test-plan.md](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/test-plan.md).

### Constraints/lookups
Each of the following constraints should be covered, along with the lookup that it implies (indicated where applicable):
- [x] `mustSpendPubKeyOutput` https://github.com/Plutonomicon/cardano-transaction-lib/pull/1024
  - **Note**: This constraint is included in our stake key integration with Plutip. We should still write a full contract that uses it
- [ ] `mustBeSignedBy` https://github.com/Plutonomicon/cardano-transaction-lib/pull/1101
  - **Note**: This constraint is also included in our Plutip stake key integration, as with `mustSpendPubKeyOutput`
  - **Relevant issue**: https://github.com/Plutonomicon/cardano-transaction-lib/issues/569
- [ ] `mustHashDatum`
- [x] `mustIncludeDatum` https://github.com/Plutonomicon/cardano-transaction-lib/pull/1087
- [x] `mustProduceAtLeastTotal`. Also implies:
  - [x] `mustProduceAtLeast`
- [x] `mustSatisfyAnyOf`
- [x] `mustSpendAtLeastTotal`. Also implies:
  - [x] `mustSpendAtLeast`
- [ ] `mustValidateIn` https://github.com/Plutonomicon/cardano-transaction-lib/pull/1101

In addition, several redeemer combinations in a **single transaction** must be covered:
- [x] Two or more `Spend` redeemers
- [x] (At least) One each of a `Spend` and `Mint` redeemer

### Unit and integration testing
Currently, we require parsing tests for the following data structures:
- `ogmios-datum-cache`
  - [x] `GetDatumsByHashesR` https://github.com/Plutonomicon/cardano-transaction-lib/pull/408 https://github.com/Plutonomicon/cardano-transaction-lib/pull/959
  - [x] `GetTxByHashR`

klntsky:
Done

PR #1199, opened 2022-11-09
Bunch of Plutip tests for CTL Test Plan Coverage

uhbif19:
#1080

### Pre-review checklist


examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,161 @@
...
+    constraints =
+      (mconcat $ fst <$> lcs) :: Constraints.TxConstraints Void Void
+  txHash <- buildBalanceSignAndSubmitTx
+    (Lookups.mintingPolicy mintingPolicy <> lookups)
+    constraints
+  awaitTxConfirmed txHash
+
+  --- Return
+
+  pure unit
```
klntsky, line 82:
```suggestion
  void $ awaitTxConfirmed txHash
```

examples/MultipleRedeemers.purs
```diff
@@ -0,0 +1,161 @@
...
+  txHash <- buildBalanceSignAndSubmitTx
+    ( Lookups.mintingPolicy mintingPolicy <> Lookups.mintingPolicy mp <>
+        unlockingLookups
+    )
+    (unlockingConstraints <> mintingConstraints)
+  awaitTxConfirmed txHash
+
+  --- Return
+
+  pure unit
```
klntsky, line 114:
```suggestion
  void $ awaitTxConfirmed txHash
```

examples/PlutusV2/ReferenceInputs.purs
```diff
@@ -202,3 +209,28 @@ alwaysMintsPolicyScriptV2 =
...
+          $ BigInt.fromInt sum
+      ]
+
+    lookups :: Lookups.ScriptLookups Void
+    lookups = Lookups.mintingPolicy mp
+
+  txHash <- Helpers.buildBalanceSignAndSubmitTx lookups constraints
+  awaitTxConfirmed txHash
+
+  pure unit
```
klntsky, line 236:
```suggestion
  void $ awaitTxConfirmed txHash
```

test/MustSpendTotal.purs
```diff
@@ -0,0 +1,69 @@
...
+testIsAuthomorhism valueToConstraint constraintToValue = do
+  token1 <- mkTokenName "Token 1"
+  token2 <- mkTokenName "Token 2"
+  cs <- mkCurrencySymbol alwaysMintsPolicyMaybe
+  let
+    value1 = Value.singleton cs token1 one
+    value2 = Value.singleton cs token2 one
+    constraint = valueToConstraint value1 <> valueToConstraint value2
+  (constraintToValue constraint) `shouldEqual` (value1 <> value2)
+  pure unit
```
klntsky, line 58:
```suggestion
```

test/Plutip.purs
```diff
@@ -567,6 +569,184 @@ suite = do
...
+            constraints' :: Constraints.TxConstraints Void Void
+            constraints' = Constraints.mustProduceAtLeast
+              $ Value.singleton cs tn
+              $ BigInt.fromInt 100
+            lookups' = lookups <> Lookups.ownPaymentPubKeyHash pkh
+
+          txHash' <- buildBalanceSignAndSubmitTx lookups' constraints'
+          awaitTxConfirmed txHash'
+
+          pure unit
```
klntsky, line 614:
```suggestion
          void $ awaitTxConfirmed txHash'
```

test/Plutip.purs
```diff
@@ -567,6 +569,184 @@ suite = do
...
+            constraints' = Constraints.mustProduceAtLeast
+              $ Value.singleton cs tn
+              $ BigInt.fromInt 101
+            lookups' = lookups <> Lookups.ownPaymentPubKeyHash pkh
+
+          ubTx <- liftedE $ Lookups.mkUnbalancedTx lookups' constraints'
+          result <- balanceTx ubTx
+          result `shouldSatisfy` isLeft
+
+          pure unit
```
klntsky, line 659:
```suggestion
```

test/Plutip.purs
```diff
@@ -567,6 +569,184 @@ suite = do
...
+            constraints' :: Constraints.TxConstraints Void Void
+            constraints' = Constraints.mustSpendAtLeast
+              $ Value.singleton cs tn
+              $ BigInt.fromInt 100
+            lookups' = lookups <> Lookups.ownPaymentPubKeyHash pkh
+
+          txHash' <- buildBalanceSignAndSubmitTx lookups' constraints'
+          awaitTxConfirmed txHash'
+
+          pure unit
```
klntsky, line 703:
```suggestion
          void $ awaitTxConfirmed txHash'
```

test/Plutip.purs
```diff
@@ -567,6 +569,184 @@ suite = do
...
+            constraints' = Constraints.mustSpendAtLeast
+              $ Value.singleton cs tn
+              $ BigInt.fromInt 101
+            lookups' = lookups <> Lookups.ownPaymentPubKeyHash pkh
+
+          ubTx <- liftedE $ Lookups.mkUnbalancedTx lookups' constraints'
+          result <- balanceTx ubTx
+          result `shouldSatisfy` isLeft
+
+          pure unit
```
klntsky, line 748:
```suggestion
```

klntsky (changes requested):
Please also tick all the appropriate boxes in the test-plan itself

test/MustSpendTotal.purs
```diff
@@ -0,0 +1,68 @@
...
+   . (Value.Value -> TxConstraints i o)
+  -> (TxConstraints i o -> Value.Value)
+  -> Aff Unit
+testIsAuthomorhism valueToConstraint constraintToValue = do
+  token1 <- mkTokenName "Token 1"
+  token2 <- mkTokenName "Token 2"
+  cs <- mkCurrencySymbol alwaysMintsPolicyMaybe
+  let
+    value1 = Value.singleton cs token1 one
+    value2 = Value.singleton cs token2 one
```
Luis-omega, line 55: I think this must be a property test instead.

examples/Helpers.purs
```diff
@@ -81,3 +86,16 @@ mustPayToPubKeyStakeAddress pkh Nothing =
...
   Constraints.mustPayToPubKeyAddress pkh skh
 
+submitAndLog
+  :: forall (r :: Row Type). BalancedSignedTransaction -> Contract r Unit
+submitAndLog bsTx = do
+  txId <- submit bsTx
+  logInfo' $ "Tx ID: " <> show txId
+  awaitTxConfirmed txId
+  mbTransaction <- getTxByHash txId
+  logInfo' $ "Tx: " <> show mbTransaction
```
Luis-omega, line 96:
```suggestion
  logInfo' $ "Retrieved tx: " <> show mbTransaction
```

or something like that?

test/Plutip.purs
```diff
@@ -567,6 +569,176 @@ suite = do
           bsTx <- signTransaction =<< liftedE (balanceTx ubTx)
           submitAndLog bsTx
 
+    test "runPlutipContract: mustProduceAtLeast success" do
```
Luis-omega, line 572: I think most of the code of the test can be put in a new function and use it to short the size of all of the tests.

This would allow the test to just look like 

```
makeTest FirstConstraintValue SecondConstraintValue =
    do
    (firstTx /\ secondTx) <- newFunction FirstConstraintValue SecondConstraintValue
    testWith someFunction  firstTx secondTx
  where
    someFunction = ...
```

improving readability.

uhbif19:
@klntsky 

Couple of questions.

1. Why do we need to distinguish between `Ctl.Internal.Types.Transaction (TransactionHash)` and `Ctl.Internal.QueryM.Ogmios (TxHash)`
2. Why getTxByHash decodes content of GetTxByHashR into Transaction, while for Datum content is already decoded in GetDatumsByHashesR?

klntsky:
> Why do we need to distinguish between Ctl.Internal.Types.Transaction (TransactionHash) and Ctl.Internal.QueryM.Ogmios (TxHash)

The first one is cardano domain, the second is ogmios query domain

> Why getTxByHash decodes content of GetTxByHashR into Transaction, while for Datum content is already decoded in GetDatumsByHashesR?

In ogmios domain, tx is base64 string and not json.

test/Plutip.purs
klntsky: @uhbif19 please don't do this. These tests are meant to be copy-paste-able by the users. No need to make them less readable just for a few code lines saved. We don't optimise for that metric, plus, it's easier to read when it's straightforward (less identifiers to keep in mind).

klntsky (commented):
All good except of the typo above

test/Plutip.purs
uhbif19: @Luis-omega @klntsky Yes I was thinking about parametrizing tests initially, but remembered that you said something lke this comment before.

uhbif19:
@klntsky 

I think `mkCurrencySymbol` has an issue in it, cuz when I use it for Arbitrary generation it generates a bunch of such messages:

`Deserialization failed in ScriptHash because: Invalid cbor: expected tuple 'hash length' of length 28 but got length Len(2).`

I think the problem is that imported Rust code for `scriptHashFromBytes` prints the error while not raising an exception here: 

`scriptHashFromBytes (wrap byteArr) $> CurrencySymbol byteArr`  

By removing  `scriptHashFromBytes` I am geting rid of the error, but I am not sure it is the right way.

test/MustSpendTotal.purs
uhbif19: Of course, you are right. I thought we have no property test on this project, but I checked and found I was wrong.

test/MustSpendTotal.purs
```diff
@@ -0,0 +1,68 @@
...
+
+suite :: TestPlanM (Aff Unit) Unit
+suite = do
+  group "mustXAtLeastTotal works correctly" do
+    test "mustSpendAtLeastTotal works correctly" $ do
+      testIsAuthomorhism mustSpendAtLeast mustSpendAtLeastTotal
+    test "mustProduceAtLeastTotal works correctly" $ do
+      testIsAuthomorhism mustProduceAtLeast mustProduceAtLeastTotal
+
+testIsAuthomorhism
```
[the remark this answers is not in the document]
uhbif19: The thing I wanted to say is probably that `mustProduceAtLeast` is homomorphism and has left (?) inverse homomorphism `mustProduceAtLeastTotal`. I think this test proves that.

But it is probably to strange property, so I will rewrite to something more straightforward.

test/MustSpendTotal.purs
```diff
@@ -0,0 +1,76 @@
...
+
+-- Properties
+
+prop_isHomomorphism
+  :: forall i o
+   . (Value.Value -> TxConstraints i o)
+  -> Value.Value
+  -> Value.Value
+  -> Boolean
+prop_isHomomorphism f x y = ((f x) <> (f y)) `customEq` (f (x <> y))
```
Luis-omega, line 50:
```suggestion
prop_isHomomorphism f x y = (f x <> f y) `customEq` f (x <> y)
```

src/Internal/Plutus/Types/Value.purs
```diff
@@ -82,6 +87,18 @@ instance EncodeAeson Value where
...
     { "getValue": encodeAeson mph }
 
+arbitrarySingletonValue :: Gen Value
+arbitrarySingletonValue = do
+  currencySymbol <- unMaybeGen $ mkCurrencySymbol <$> arbitrary
+  tokenName <- unMaybeGen $ mkTokenName <$> arbitrary
+  num <- BigInt.fromInt <$> arbitrary
+  pure $ singleton currencySymbol tokenName num
+
+instance Arbitrary Value where
```
Luis-omega, line 97: Did you look at plutus or cardano or someone else to write this instance? (I'm curious).

Also, I don't think we need wholly arbitrary names for the `CurrencySymbol`, you can use a fixed set of names with the `Test.QuickCheck.Gen.frequency` function instead. This would allow you to remove the need for `unMaybeGen` and with it the new module `Internal.Quickcheck` .

test/MustSpendTotal.purs
```diff
@@ -0,0 +1,76 @@
...
+import Test.QuickCheck (Result, (===))
+import Test.Spec.QuickCheck (quickCheck)
+
+-- Run with `spago test --main Test.Ctl.MustSpendTotal`
+main :: Effect Unit
+main = launchAff_ do
+  interpret suite
+
+suite :: TestPlanM (Aff Unit) Unit
+suite = do
```
Luis-omega, line 31: I think it would be better to have two tests for the `homomorphism`, one with a fixed name for the `currency symbol` and `token name` and the other with a mix of them. 

In fact I don't know if we need to test it with more than one `currency symbol` and `token name`.

Luis-omega (changes requested):

klntsky:
@uhbif19 
> I think mkCurrencySymbol has an issue in it, cuz when I use it for Arbitrary generation it generates a bunch of such messages:

No, it's a bug in `Internal/Serialization/Hash.js`. Of course we should not allow any `console.log` there. Could you please remove these calls in this PR?

klntsky:
> By removing scriptHashFromBytes I am geting rid of the error, but I am not sure it is the right way.

No, it's not. You are, in fact, removing the validation.

test/MustSpendTotal.purs
uhbif19: I do not understand what you mean by `one with a fixed name for the currency symbol and token name` and why it is better.

> In fact I don't know if we need to test it with more than one currency symbol and token name.

Why not? It does not produce complex or slow code, why creating special case instead of just using one Arbitrary instance?

src/Internal/Plutus/Types/Value.purs
uhbif19: What is wrong with having `Internal.Quickcheck` module? It is not a multilayer abstraction it is a simple utility module, which is used only in tests. Maybe it bloating bundle size, I do not know, maybe than move it to tests dir or something.

uhbif19: I do not see `Arbitrary` instance in API docs. Maybe I am looking not in the right place?

https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/Plutus-V1-Ledger-Value.html

test/MustSpendTotal.purs
Luis-omega: I thought it was slow, but otherwise is okay. 

The reason we would like to have two separate cases is to improve the possibility to have two different values with the same minting policy and token name in the test. I think this is the most interesting case and the one that needs more coverage from property testing.

src/Internal/Plutus/Types/Value.purs
Luis-omega: I don't know, once I tried to provide `Arbitrary` instances for them and other types, there was a discussion a bout how to do it properly, but at the end I didn't proceed to do it. To me this instance looks good.

Luis-omega (approved):
I think is ready to merge once you update it and remove the unneeded imports

klntsky (approved):

PR #1201, opened 2022-11-09
cleanup on SIGINT

### LINKED ISSUE #1176: clean up temporary directory when exiting with ctrl+c
This [PR](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1144) cleans up the temporary postgres databaseDir. But this doesn't work when we exit Plutip tests with ctrl+c.

pb99u069:
Closes #1176.

### Pre-review checklist

### DIFF
--- a/doc/side-by-side-ctl-plutus-comparison.md
+++ b/doc/side-by-side-ctl-plutus-comparison.md
@@ -315,7 +315,7 @@ grab
   -> TransactionHash
   -> Contract () Unit
 grab vhash validator txId = do
-  let scriptAddress = scriptHashAddress vhash
+  let scriptAddress = scriptHashAddress vhash Nothing
   utxos <- fromMaybe Map.empty <$> utxosAt scriptAddress
   case fst <$> find hasTransactionId (Map.toUnfoldable utxos :: Array _) of
     Just txInput ->
--- a/doc/test-plan.md
+++ b/doc/test-plan.md
@@ -38,11 +38,11 @@ CTL's primary user interface is its constraints and lookups API, modeled after t
 - [x] `mustIncludeDatum`
 - [x] `mustPayToPubKeyWithDatum`
 - [x] `mustPayToPubKeyAddressWithDatum`
-- [ ] `mustProduceAtLeastTotal`. Also implies
-  - [ ] `mustProduceAtLeast`
+- [x] `mustProduceAtLeastTotal`. Also implies
+  - [x] `mustProduceAtLeast`
 - [x] `mustSatisfyAnyOf`
-- [ ] `mustSpendAtLeastTotal`. Also implies
-  - [ ] `mustSpendAtLeast`
+- [x] `mustSpendAtLeastTotal`. Also implies
+  - [x] `mustSpendAtLeast`
 - [ ] `mustValidateIn`
 
 The following constraints were added for `PlutusV2` features as part of our `v2.0.0` release. They do not have direct correspondances in `plutus-apps`:
@@ -66,8 +66,8 @@ That release also included the following constraints for working with native scr
 In addition, several redeemer combinations in a **single transaction** must be covered by tests or examples as well, namely
 
 - [x] Two or more `Mint` redeemers
-- [ ] Two or more `Spend` redeemers
-- [ ] (At least) One each of a `Spend` and `Mint` redeemer
+- [x] Two or more `Spend` redeemers
+- [x] (At least) One each of a `Spend` and `Mint` redeemer
 
 #### Stake operations
 
@@ -215,8 +215,8 @@ Currently, we require parsing tests for the following data structures, organized
   - [x] `SubmitTxR`
 - `ogmios-datum-cache`
   - [x] `GetDatumByHashR`
-  - [ ] `GetDatumsByHashesR`
-  - [ ] `GetTxByHashR`
+  - [x] `GetDatumsByHashesR`
+  - [x] `GetTxByHashR`
 - `cardano-serialization-lib`
   - `Transaction`
     - [x] Serialization
--- a/examples/AlwaysMints.purs
+++ b/examples/AlwaysMints.purs
@@ -1,7 +1,13 @@
 -- | This module demonstrates how the `Contract` interface can be used to build,
 -- | balance, and submit a smart-contract transaction. It creates a transaction
 -- | that mints a value using the `AlwaysMints` policy
-module Ctl.Examples.AlwaysMints (main, example, contract, alwaysMintsPolicy) where
+module Ctl.Examples.AlwaysMints
+  ( alwaysMintsPolicy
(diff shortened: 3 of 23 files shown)


src/Internal/Plutip/Spawn.purs
```diff
@@ -86,15 +86,22 @@ killOnExit child = do
...
   ChildProcess.onExit child \_ -> do
     _rmdirSync workingDir
   -- TODO
   -- cleanup directories in case of ctrl+c exit,
   -- code below doesn't work as intended.
   -- https://github.com/Plutonomicon/cardano-transaction-lib/issues/1176
-  Process.onExit \_ -> do
-    _rmdirSync testClusterDir
-    _rmdirSync workingDir
+  -- Process.onSignal SIGINT do
```
pb99u069, line 99: This would be the cleanest solution. The problem is though that after registering the `onSignal SIGINT` listener, the `testClusterDir` and the `workingDir` change with each child process. So this handler tries to remove directories that have already been removed by the `ChildProcess.onExit` listener on line 93.

To solve this, we would need to know which are the current `testClusterDir` and `workingDir` when ctrl+c is pressed. But as far as I can see that's impossible to know at the moment of registering the listener.

klntsky: If they have already been deleted, just do nothing.

pb99u069: the problem is that it always tries to delete the tmp directories from the first child process (as that is the moment when the listener is created). but it doesn't delete the tmp directories created by the current (i.e when ctrl+c is pressed) child process

klntsky: Isn't `cleanupTmpDir` called for each of the child processes? If so, why the correct tmp dirs are not passed?

jy14898:
Before too much work is done on this, I'd like to say I did a quick exploratory test a little while back with `SIGINT` on my plutip PR and spago was messing with the signal. Calling `SIGINT` (ctl-c) from the terminal didn't trigger the handler, only via a direct signal to `node` which I performed via `htop`

src/Internal/Plutip/Spawn.purs
pb99u069: I'm stuck here. The problem from above is solved, the problem was that by using `Process.exit` in the SIGINT listener, the listeners registered in the second and following child processes never gets executed. 

so we must call `Process.exit` only in the last listener, which I'm trying to do in `cleanupOnSigint`. TTriggering SIGINT produces the following error though which I just can't figure out. Do I maybe have to work in Aff instead of Effect? 

  Error: (ClientHttpError There was a problem making the request: request failed)
    at Object.exports.error (/home/pekka/ctl/cardano-transaction-lib/output/Effect.Exception/foreign.js:8:10)
    at Object.$$throw [as throw] (/home/pekka/ctl/cardano-transaction-lib/output/Effect.Exception/index.js:18:45)
    at /home/pekka/ctl/cardano-transaction-lib/output/Ctl.Internal.Plutip.Server/index.js:151:57
    at /home/pekka/ctl/cardano-transaction-lib/output/Data.Either/index.js:203:24
    at /home/pekka/ctl/cardano-transaction-lib/output/Ctl.Internal.Plutip.Server/index.js:153:71
    at run (/home/pekka/ctl/cardano-transaction-lib/output/Effect.Aff/foreign.js:278:22)
    at /home/pekka/ctl/cardano-transaction-lib/output/Effect.Aff/foreign.js:348:19
    at drain (/home/pekka/ctl/cardano-transaction-lib/output/Effect.Aff/foreign.js:120:9)
    at Object.enqueue (/home/pekka/ctl/cardano-transaction-lib/output/Effect.Aff/foreign.js:141:11)
    at /home/pekka/ctl/cardano-transaction-lib/output/Effect.Aff/foreign.js:339:27

jy14898: Some thoughts: multiple listeners conflicting can be fixed by detatching the listener when the resources are cleaned up in the bracket. This would also be a good time to run the cleanup.
It sounds like maybe

I'm surprised you get that error after calling exit, exit should shut down the process as quickly as possible without allowing other events to be run (other than those that listen for exit).

jy14898: Also I'm curious if you have noticed the behaviour with ctrl c and spago like I have in my comment below

pb99u069: thanks for the input. I think detaching the `onSignal SIGINT` listener after each child process is exactly what I need but I couldn't figure out how to do it. would I need the ffi, or something similar to this: https://stackoverflow.com/questions/57259201/how-to-schedule-a-removeeventlistener-inside-an-addeventlistener ?

pb99u069: about the error: I think it occurs before `Process.exit`, but I couldn't yet figure out where exactly. 

ctrl-c for me doesn't behave as you describe below. The `onSignal SIGINT` handler gets triggered and that wasn't the case for you if I get it right.

jy14898: I guess really we shouldn't detach listeners (by name) in purescript because of referential transparency, we'd need to create a version of onSignal which returns the function it attached, so we can later detach.

```purescript
foreign import data OnSignalRef :: Type

foreign import onSignalImpl :: String -> Effect Unit -> Effect OnSignalRef

foreign import removeOnSignal :: OnSignalRef -> Effect Unit

onSignal :: Signal -> Effect Unit -> Effect OnSignalRef
onSignal sig = onSignalImpl (Signal.toString sig)
```

```javascript
exports.onSignalImpl = signal => callback => () => {
  process.on(signal, callback);
  return { signal, callback };
};

exports.removeOnSignal = { signal, callback } => () => {
  process.removeListener(signal, callback);
};
```

This also helps not make https://github.com/Plutonomicon/cardano-transaction-lib/issues/864 worse by making sure we cleanup

src/Internal/Plutip/Server.purs
```diff
@@ -376,6 +383,7 @@ startPostgresServer pgConfig params = do
     databaseDir = workingDir <> "/postgres/data"
     postgresSocket = workingDir <> "/postgres"
     testClusterDir = (dirname <<< dirname) params.nodeConfigPath
+  sig <- liftEffect $ cleanupOnSigint workingDir testClusterDir
```
pb99u069, line 386: The onSIGINT listener cleans up the directory as it should, except if the signal is sent between the creation of the directory (done on line 379) and the registering of the listener. but I guess this problem can't be solved, as the listener can only be registered once the directory name is known?

src/Internal/Plutip/Spawn.purs
klntsky: Fully agree @jy14898

klntsky: We can also make the FFI function that attaches a handler return a callback that de-attaches it.

PR #1214, opened 2022-11-11
[CI] fix

aciceri:
This fix an evaluation error in Hydra due to a typo in `hydraJobs`.

### Pre-review checklist


nix/test-nixos-configuration.nix
```diff
@@ -27,8 +27,9 @@
 
   services.cardano-node = {
     enable = true;
-    environment = "preprod";
     systemdSocketActivation = true;
+    nodeConfigFile = "${cardano-configurations}/network/mainnet/cardano-node/config.json";
```
klntsky, line 31: Why mainnet only? We probably need network as a parameter, does it make sense?

aciceri: It would makes sense! The `test-nixos-configuration.nix` has two usages right now:

- it's used to compose different modules (look `nixosConfigurations.test`) and
  test if there are conflicts between options. CI currently builds a VM with
  that `nixosConfiguration` (this is what this PR fixes) but doesn't run it.
  Then, unless the error is a type error during Nix evaluation, we are not
  looking for it here, having different `nixosConfiguration`s for different
  networks wouldn't make sense in my opinion.

- to have a `nix run .#vm` command that builds and run that VM. Here we could
  add two more commands that run the the same `nixosConfiguration` but with
  different networks.d

klntsky (approved):

PR #1219, opened 2022-11-13
Make wallet errors non-fatal in e2e.

### LINKED ISSUE #1209: Make exceptions in E2E browser automation non-fatal
```
              subscribeToBrowserEvents (Just $ wrap 1000.0) page
                case _ of
                  ConfirmAccess -> launchAff_ someWallet.confirmAccess
                  Sign -> launchAff_ someWallet.sign
                  Success -> pure unit
                  Failure err -> throw err
```

Currently if there are exceptions thrown in sign or confirmAccess functions, the node process exists, because they are uncatched. We should re-throw these exceptions in the Aff context in which `subscribeToBrowserEvents` is called, to ensure we are only error'ing one test, and not the whole suite.

zmrocze:
Closes #1209.

### Pre-review checklist

- [X] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

### note for reviewer

I verified that the error in wallet sign/confirm is non-fatal after the change and fails the test, in difference with previous version, by running the tests with `throwError` in place of sign/confirm.
But i am not fully sure the returned cancelor is correct. I worked under the assumption that a function like:
```haskell
id? :: Aff a -> Aff a
id? aff = makeAff $ \k -> 
    map fiberCanceler $ launchAff $ do
        res <- try aff
        liftEffect $ k res
``` 
is identity on Aff.


src/Internal/Test/E2E/Runner.purs
```diff
@@ -265,12 +273,20 @@ testPlan opts@{ tests } rt@{ wallets } =
...
+                  res <- try aff
+                  when (isLeft res) $ liftEffect $ k res
+
+              map fiberCanceler $ launchAff $ (try >=> k >>> liftEffect) $
+                subscribeToBrowserEvents (Just $ Seconds 10.0) page
+                  case _ of
+                    ConfirmAccess -> rethrow someWallet.confirmAccess
+                    Sign -> rethrow someWallet.sign
+                    Success -> pure unit
+                    Failure err -> throw err
```
klntsky, line 288: Here as well

src/Internal/Test/E2E/Runner.purs
```diff
@@ -265,12 +273,20 @@ testPlan opts@{ tests } rt@{ wallets } =
...
+                rethrow aff = launchAff_ do
+                  res <- try aff
+                  when (isLeft res) $ liftEffect $ k res
+
+              map fiberCanceler $ launchAff $ (try >=> k >>> liftEffect) $
+                subscribeToBrowserEvents (Just $ Seconds 10.0) page
+                  case _ of
+                    ConfirmAccess -> rethrow someWallet.confirmAccess
+                    Sign -> rethrow someWallet.sign
+                    Success -> pure unit
```
klntsky, line 287: Shouldn't we pass the `unit` to `k` callback to make it exit successfully? Have you tried to run manually the e2e test suite for all these cases (error from sign, error from confirm, success, failure)?

klntsky (changes requested):

src/Internal/Test/E2E/Runner.purs
zmrocze: I think not. Because we should pass result of the whole computation, to continue afterwards (so result of `subscribeToBrowserEvents`, and we do pass it to k). The "case" snippet gets called multiple times inside `subscribeToBrowserEvents`. Changing it to:
```
case _ of 
  Success -> k unit
``` 
would shortcircuit the computation and exit once this handler is called. Simmilar with `Failure -> throw Error` (but see below comment).

### testing
I wasn't able to run e2e tests succesfully, so i havent verified that succesful tests remain succesful.

So i tested like this: 
1. i substituted `throwError` for `confirmAccess`
2. `confirmAccess` then errored before the usual errors with my supplied error
3. it errors the test, not the whole test suite (in contrary to previous commit which i tested as well)

We could wait to first fix e2e tests and then Ill test this change again, but i think this works as intended

zmrocze:
I had confusion with 
```
Failure -> throw err
```
because this is called under `try` inside `subscribeToBrowserEvents` ([link](https://github.com/Plutonomicon/cardano-transaction-lib/blob/bfcca2a16654a0a4a315bb93017aa9dd9c24bffa/src/Internal/Test/E2E/Feedback/Node.purs#L76))
but now i see what happens.

It can be changed to
```
Failure -> pure unit
```
with no difference, because the error is caught and ignored no matter what.
Both cases `Failure` and `Success` can have `pure unit` because the logic for them is handled inside `subscribeToBrowserEvents` directly [link](https://github.com/Plutonomicon/cardano-transaction-lib/blob/bfcca2a16654a0a4a315bb93017aa9dd9c24bffa/src/Internal/Test/E2E/Feedback/Node.purs#L78-L79).

PR #1223, opened 2022-11-15
Update package-lock in template and add a template checker script

klntsky:
### Pre-review checklist


scripts/template-check.sh
```diff
@@ -0,0 +1,52 @@
...
+        CTL_REVISION="$1"
+        echo "Using CTL revision $CTL_REVISION"
+    fi;
+
+    echo "This script assumes that CTL runtime is running."
+
+    rm -rf "$TEST_DIR"
+    mkdir -p "$TEST_DIR"
+    cd "$TEST_DIR"
+    nix flake init -t github:Plutonomicon/cardano-transaction-lib/"$CTL_REVISION"
```
jy14898, line 43: We don't have to fetch from github (although maybe supporting all 3 cases is good)

```suggestion
    nix flake init -t ../.
```

Also might need to be something other than `../.`, I'm not sure where main actually ends up running

jy14898 (approved):

Issue #1229, opened 2022-11-15
Cleanup race condition with rmdirSync


Geometer1729:
Tests sometimes throw
```
Error: ENOENT: no such file or directory, stat '/build/test-cluster29'
    at Object.statSync (node:fs:1538:3)
    at __node_internal_ (node:internal/fs/utils:793:8)
    at Object.rmdirSync (node:fs:1156:15)
    at /nix/store/9vxqdmxsri53q0bginxacl85rr27a43r-Test.Main-compiled/Ctl.Internal.Plutip.Spawn/foreign.js:9:39
    at __do (/nix/store/9vxqdmxsri53q0bginxacl85rr27a43r-Test.Main-compiled/Ctl.Internal.Plutip.Spawn/index.js:148:63)
    at process.<anonymous> (/nix/store/9vxqdmxsri53q0bginxacl85rr27a43r-Test.Main-compiled/Node.Process/foreign.js:14:21)
    at process.emit (node:events:532:35) {
  errno: -2,
  syscall: 'stat',
  code: 'ENOENT',
  path: '/build/test-cluster29'
}
```
It seems to be a race condition with deleting the files here:
https://github.com/Plutonomicon/cardano-transaction-lib/blob/3789e92242c925523c02cad6b21a880490feba49/src/Internal/Plutip/Spawn.purs#L122-L132

The error should be safe to just ignore.

pb99u069:
This problem should be fixed with [#1201](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1201), where we don't clean up `testClusterDir` anymore which gets already removed from the plutip side

Geometer1729:
This seems to work downstream :+1:

PR #1231, opened 2022-11-16
cleanup on SIGINT

### LINKED ISSUE #1176: clean up temporary directory when exiting with ctrl+c

klntsky:
Closes #1176 #1229.

Re-creation of https://github.com/Plutonomicon/cardano-transaction-lib/pull/1201

### Pre-review checklist


jy14898:
Copy of my latest comment:

We need to be careful about how we handle SIGINT, it's not currently disallowed (we don't explicitly state it) for people to run multiple plutip clusters in parallel with different port configs.

2 solutions:
 1. Ban it. Not great, people are already doing it this way, although testPlutipContracts is now an alternative they can use
 2. Don't exit on SIGINT. Instead, we change our test runner to listen for SIGINT and cancel the tests, such that their brackets are respected (cancel the aff fibers). And then as no more work is scheduled the program will exit naturally (we may still need to set process.exitCode = 1).

I'm going to spend a moment investigating 2, I will try and get back on it soon

Update: This works, I'll describe how we should do it

jy14898:
```purescript
-- interruptOnSignal should go in Contract.Test.Util or something
-- Node.Process missing this binding, it lets you set the code that node will use when it gracefully exits
foreign import exitCode :: Int -> Effect Unit
-- exports.exitCode = code => () => { process.exitCode = code; };

interruptOnSignal signal fiber = Process.onSignal signal do
  launchAff_ do
    killFiber (error $ Signal.toString signal) fiber

-- Maybe this should go in interpretWithConfig by default, not sure
main :: Effect Unit
main = interruptOnSignal Signal.SIGINT =<< launchAff do
  flip cancelWith (effectCanceler (exitCode 1)) do
    Utils.interpretWithConfig
    ...
```

src/Contract/Test/Utils.purs
```diff
@@ -629,3 +632,14 @@ checkTxHasMetadata
...
+
+--------------------------------------------------------------------------------
+-- function to cancel aff fibers on signal
+--------------------------------------------------------------------------------
+
+foreign import exitCode :: Int -> Effect Unit
+
+interruptOnSignal :: forall a. Signal -> Fiber a -> Effect Unit
+interruptOnSignal signal fiber = Process.onSignal signal do
+  launchAff_ do
```
klntsky, line 644: Can we fork here though? I assumed signal handlers must be sync, or the default action (e.g. exiting) will be executed before the async action has a chance to run. Is that not true?

pb99u069: not sure, feel still a little bit lost with Aff/Effect.. But as I understand it the default action (not exiting, but removing the temp dirs) is supposed to run before the async action which is killing the fibre

jy14898: Indeed, SIGINT does not exit once you add a custom handler. So, instead of calling exit, we cancel the fiber which has the same effect in the end: the process will exit when there are no more possible events, which eventually happens once the fiber has run its cleanup.

src/Internal/Plutip/Spawn.purs
```diff
@@ -117,16 +120,25 @@ waitForStop (ManagedProcess cmd _ closedAVar) = do
...
+cleanupOnSigint workingDir testClusterDir = do
+  sig <- onSignal SIGINT do
     _rmdirSync workingDir
-  -- TODO
-  -- cleanup directories in case of ctrl+c exit,
-  -- code below doesn't work as intended.
-  -- https://github.com/Plutonomicon/cardano-transaction-lib/issues/1176
-  Process.onExit \_ -> do
     _rmdirSync testClusterDir
+  pure sig
```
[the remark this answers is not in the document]
pb99u069: @jy14898, sorry but I couldn't figure out how the high level 'kill fiber' SIGINT handler can invoke the cleanup. The temporary postgres directories are created with `startPostgresServer`, so they cannot be known to `interruptOnSignal` in `main`, no?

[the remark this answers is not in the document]
jy14898: > sorry but I couldn't figure out how the high level 'kill fiber' SIGINT handler can invoke the cleanup. The temporary postgres directories are created with startPostgresServer, so they cannot be known to interruptOnSignal in main, no?

`kill fiber` should invoke the cleanup handlers of brackets - this is only true if puresript-spec is actually using brackets to do the cleanup, however.

It looks like it isn't: https://github.com/purescript-spec/purescript-spec/blob/f0661f0f750205ccc7dcb78ee3d6ffc845c2e2b2/src/Test/Spec/Runner.purs#L132. Annoying

Maybe the whole kill fiber approach wont work then

jy14898 (changes requested):
We need a note somewhere in https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/plutip-testing.md and maybe in the code documentation saying that running plutip servers will attach on SIGINT handlers and therefore node will not exit by default, and guide them on how to use our interrupt on signal/cancel with helpers.

Also something in the changelog to reflect this new behaviour

pb99u069:
> We need a note somewhere in https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/plutip-testing.md and maybe in the code documentation saying that running plutip servers will attach on SIGINT handlers and therefore node will not exit by default, and guide them on how to use our interrupt on signal/cancel with helpers.
> 
> Also something in the changelog to reflect this new behaviour

I added an explanation to plutip-testing.md. I'm not sure though what you mean with 'guide them on how to use our interrupt on signal/cancel with helpers'.

doc/plutip-testing.md
```diff
@@ -183,3 +183,7 @@ let
 Although stake keys serve no real purpose in plutip context, they allow to use base addresses, and thus allow to have the same code for plutip testing, in-browser tests and production.
 
 Note that CTL re-distributes tADA from payment key-only ("enterprise") addresses to base addresses, which requires a few transactions before the test can be run. Plutip can currently handle only enterprise addreses (see [this issue](https://github.com/mlabs-haskell/plutip/issues/103)).
+
+## custom SIGINT handlers
+
+SIGINT does not exit the node process when running a plutip server. Instead, we use a custom handler listening on SIGINT that cancels the fiber. This has the same effect in the end: The process will exit when there are no more possible events, one of them being our custom `cleanupOnSigint` that gets attached and detached to every plutip cluster. We avoid exiting on SIGINT so that multiple plutip servers with different port configs can be run in parallel.
```
jy14898, line 189: Please move this to after the `Testing with Mote` section, and add it to the table of contents at the top of the file too

```suggestion
### Note on SIGINT

Due to `testPlutipContracts`/`runPlutipContract` adding listeners to the SIGINT signal, node's default behaviour of exiting on that signal no longer occurs. This was done to add cleanup handlers and let them run in parallel instead of exiting eagerly, which is possible when running multiple clusters in parallel. To restore the exit behaviour, we provide helpers to cancel an `Aff` fiber and set the exit code, to let node shut down gracefully when no more events are to be processed. 

```purescript
...
import Contract.Test.Utils (exitCode, interruptOnSignal)
import Data.Posix.Signal (Signal(SIGINT))
import Effect.Aff (cancelWith, effectCanceler, launchAff)

main :: Effect Unit
main = interruptOnSignal SIGINT =<< launchAff do
  flip cancelWith (effectCanceler (exitCode 1)) do
    ... test suite in Aff ...
```

```

jy14898 (changes requested):
Addressing your comment

doc/plutip-testing.md
pb99u069: done, and thanks for the corrections, wasn't sure how to exactly phrase this

PR #1232, opened 2022-11-16
Handle multiple time constraints

klntsky:
Closes # .

Re-creation of https://github.com/Plutonomicon/cardano-transaction-lib/pull/1125

### Pre-review checklist

- [x] All code has been formatted using our config (`make format`)
- [x] Any new API features or modification of existing behavior is covered as defined in our [test plan](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/test-plan.md)
- [x] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

Issue #1055, opened 2022-09-20
Plutip tests sometimes fail


jy14898:
Sometimes plutip tests fail with:

```
ClientHttpError There was a problem making the request: request failed
```

I've mostly seen it with the system start test, it's affecting CI

rmgaray:
Here's a `plutip-server` log which shows the problem:

```
127.0.0.1 - - [20/Sep/2022:08:42:05 -0300] "POST /stop HTTP/1.1" 200 - "" "Mozilla/5.0 (Linux x64) node.js/14.19.1 v8/8.4.371.23-node.85"

cardano-node: Network.Socket.bind: resource busy (Address already in use)cardano-node: thread killed
cardano-node: thread killed
cardano-node: thread killed
ProcessHasExited "cluster didn't start correctly: [Left (ProcessHasExited \"cardano-node\" (ExitFailure 1))]" (ExitFailure 1)
```

klntsky:
https://github.com/Plutonomicon/cardano-transaction-lib/pull/1092 can possibly fix it

jy14898:
Closing in favour of https://github.com/Plutonomicon/cardano-transaction-lib/issues/1174 tracking this issue

PR #1236, opened 2022-11-17
Move distribution tests to after the contract tests, to avoid the cluster failure during contract tests.

jy14898:
https://github.com/Plutonomicon/cardano-transaction-lib/issues/1232 has hit https://github.com/Plutonomicon/cardano-transaction-lib/pull/1055 twice in the contract tests, which is annoying because all of the tests fail. From my experience, the failures occur after several plutip clusters have been set up, so this change should hopefully avoid this (still not great that we have these failures)

Closes # .

### Pre-review checklist

### DIFF
--- a/test/Plutip/UtxoDistribution.purs
+++ b/test/Plutip/UtxoDistribution.purs
@@ -75,7 +75,7 @@ import Type.Prelude (Proxy(Proxy))
 suite :: TestPlanM (Aff Unit) Unit
 suite = group "UtxoDistribution" do
   test
-    "runPlutipContract: stake key transfers with distribution: [[1000000000,1000000000]]"
+    "stake key transfers with distribution: [[1000000000,1000000000]]"
     do
       let
         distribution :: Array InitialUTxOs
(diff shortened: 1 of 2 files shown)


test/Plutip/UtxoDistribution.purs
```diff
@@ -75,15 +75,15 @@ import Type.Prelude (Proxy(Proxy))
 suite :: TestPlanM (Aff Unit) Unit
 suite = group "UtxoDistribution" do
   test
-    "runPlutipContract: stake key transfers with distribution: [[1000000000,1000000000]]"
+    "stake key transfers with distribution: [[1000000000,1000000000]]"
```
Luis-omega, line 78: Are we changing this in every test?

Luis-omega (approved):
I think it's worth to try it and see

test/Plutip/UtxoDistribution.purs
jy14898: We already have - the PR that changed this missed the additions from my earlier PR

PR #1249, opened 2022-11-18
Fix CIP30 Mock throwing wrong error when no collateral can be picked

### LINKED ISSUE #1083: Investigate the "No utxos at address" problem with plutip and small Ada values
https://github.com/Plutonomicon/cardano-transaction-lib/blob/22eec454f250462d8c139dfb779614c6fdb8c0bf/test/Plutip.purs#L1049

jy14898:
Closes #1083.

### Pre-review checklist

### DIFF
--- a/src/Internal/Wallet/Cip30Mock.purs
+++ b/src/Internal/Wallet/Cip30Mock.purs
@@ -37,7 +37,7 @@ import Ctl.Internal.Wallet.Key
   )
 import Data.Array as Array
 import Data.Either (hush)
-import Data.Foldable (foldMap)
+import Data.Foldable (foldMap, fold)
 import Data.Function.Uncurried (Fn2, mkFn2)
 import Data.Lens ((.~))
 import Data.Lens.Common (simple)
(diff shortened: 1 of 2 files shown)


klntsky (approved):

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -136,7 +136,7 @@ mkCip30Mock pKey mSKey = do
         (unwrap keyWallet).selectCollateral coinsPerUtxoUnit
           maxCollateralInputs
           utxos
-          >>= liftMaybe (error "No UTxOs at address")
```
uhbif19, line 139: Sorry for late comment, but what does UTxO folding mean? Are not UTxOs things persistent in blockchain and cannot be created on spot?

jy14898: We're folding `Maybe (Array TransactionUnspentOutput)` into `Array TransactionUnspentOutput`, which is equivalent to `maybe [] identity`. I did this because we fail later when trying to use the collateral array with a better error. Tbh it's not clear which failure cases end up in `Maybe` from selectCollateral

uhbif19: Thanks!

PR #1250, opened 2022-11-18
Actually use safe CurrencySymbol generation to fix flacky recursion error in Arbitrary Value

uhbif19:
https://app.slack.com/client/T01UN827FQV/C02QXGRGAPK/thread/C02QXGRGAPK-1668658696.949579
https://ci.staging.mlabs.city/log/2filgm8x6azaqidlkm1vfdf7x9i7xrg9-ctl-unit-test.drv

### Pre-review checklist

### DIFF
--- a/src/Internal/Plutus/Types/Value.purs
+++ b/src/Internal/Plutus/Types/Value.purs
@@ -89,7 +89,7 @@ instance EncodeAeson Value where
 
 arbitrarySingletonValue :: Gen Value
 arbitrarySingletonValue = do
-  currencySymbol <- unMaybeGen $ mkCurrencySymbol <$> arbitrary
+  currencySymbol <- arbitrary
   tokenName <- unMaybeGen $ mkTokenName <$> arbitrary
   num <- BigInt.fromInt <$> genPositive
   pure $ singleton currencySymbol tokenName num


src/Internal/Plutus/Types/Value.purs
```diff
@@ -89,7 +89,7 @@ instance EncodeAeson Value where
 
 arbitrarySingletonValue :: Gen Value
 arbitrarySingletonValue = do
-  currencySymbol <- unMaybeGen $ mkCurrencySymbol <$> arbitrary
+  currencySymbol <- arbitrary
```
klntsky, line 92: Is it generating a `ByteArray`? Not every ByteArray is a valid currency symbol (and can pass validation). It looks like it must be 28 bytes long:

https://github.com/input-output-hk/cardano-ledger/blob/8b6f8e1a75034ca66fd66a39d437252eec927d71/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L403

I suggest you to use `oneOf` and a list of pre-defined examples. We don't care about the byte contents, because we simply pass the values around.

klntsky (changes requested):
See above

src/Internal/Plutus/Types/Value.purs
uhbif19: It does use the list of predefined examples inside. 

https://github.com/Plutonomicon/cardano-transaction-lib/blob/75d3da39ec0b1defd1d165f7fe8a1963285ae2df/src/Internal/Plutus/Types/CurrencySymbol.purs#L64

klntsky (approved):

PR #1235, opened 2022-11-16
Replace ODC with Kupo for its core functionality

errfrom:
Closes https://github.com/Plutonomicon/cardano-transaction-lib/issues/1234

### Pre-review checklist

- [x] All code has been formatted using our config (`make format`)
- [ ] Any new API features or modification of existing behavior is covered as defined in our [test plan](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/test-plan.md)
- [ ] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included

Issue #56, opened 2022-01-27
Purescript style guide


ngua:
We have a Haskell style guide which we can partially follow, but we should probably have one for Purescript code as well (and maybe even for FFI code)

klntsky:
[This one](https://github.com/ianbollinger/purescript-style-guide/blob/master/purescript-style.md) looks good for me, though it is a bit dated (effect types are not used anymore). Also I don't see a valid rationale for *Doubly indent nested record or array literals* .

ngua:
Hmmm, I don't understand that one either. I think some of the guidelines won't work with `purty` as well. It's probably a good starting point though, thanks for sharing it

klntsky:
Since we have an autoformatter now, can we just use it as an oracle instead of writing a guide?

adamczykm:
I think there's more to style guide than formatting. We should probably have one that is based on Haskell one where possible.

klntsky:
Some PS-specific questions to address:

- should we write instance names
- naming of FFI functions

uhbif19:
@klntsky Should not it be closed, as matter is settled?

rynoV:
How about:

- Prefer `f1 $ f2 $ f3 v` over `f1 <<< f2 <<< f3 $ v` and `(f1 <<< f2 <<< f3) v`

I think the first is preferable for purescript since the compose function is more characters.

klntsky:
@rynoV I agree, also, compose uses more stack.

rynoV:
Suggestion:

- Prefer `not $ f a b` over `not f a b`. The second will generate more complex JS code and use more stack.

I haven't checked this, but based on [the definition that would be used for the latter](https://github.com/purescript/purescript-prelude/blob/68f8012bc2309d9bf5832cdf7316ad052d586905/src/Data/HeytingAlgebra.purs#L71), I think recursively resolving the instances would be more less efficient.

PR #1260, opened 2022-11-22
Blockfrost Implementation

### LINKED ISSUE #734: Remove user-defined configs and `Row Type` `Contract` parameter.
Reasons:

- The library does not necessitates using the parameter, and Slack survey exposed little interest in it.
- It's not hard to provide extra parameters to the contract with ReaderT or simply via an extra argument.
- the intention was to keep extra configs in Contract, but it's unlikely that Contract is going to be used to manage user application configs/state.

So, it's tangential to the main goal of the library.

The change should happen in v3.

### LINKED ISSUE #1119: Make `Contract` query backend configurable
Make `Contract` runtime and `ConfigParams` parametrize-able with a backend. Move every query layer function to a handle (as in *handle pattern*):

```purescript
type QueryHandle = 
  { getChainTip :: Aff ChainTip
  , getDatumByHash :: DatumHash -> Aff Datum
  ...
  }
```

Particular handle contents are TBD. Please pay close attention to this description: https://github.com/Plutonomicon/cardano-transaction-lib/issues/524 - it should highlight what is required to be in the handle. In particular, note that we use `EraSummaries` ogmios endpoint in `QueryM` for time conversions (`slotToPosixTime`, `posixTimeToSlot`), so we should expose only these functions in the handle, and not the whole `eraSummaries`.

---

`QueryM` monad should be either removed or used only for Ogmios/ODC/ctl-server backends, and Contract should contain a reference to a handle instead. `ConfigParams` should be changed: we should include a `type BackendSpec { ctlServer :: Maybe _, queryBackend:: QueryBackend }` where `data QueryBackend = CtlRuntime { ... } | BlockfrostRuntime {..}`

We want a function of type `QueryBackend -> Aff QueryHandle` (attempt to use it with `BlockfrostRuntime` should throw an error until Blockfrost backend is implemented).

### LINKED ISSUE #1234: Replace ODC with Kupo for its core functionality

errfrom:
Closes #1119 
Closes #734 
Closes #1234

Includes #1364: Some tests for `getTxMetadata`, `getScriptByHash`, and `getScriptsByHashes`
Includes #1371: Framework for implementing Blockfrost queries
Includes #1385: Foundation for generating Blockfrost fixtures
Includes #1368: `getTxMetadata` and `isTxConfirmed` queries
Includes #1378: `getDatumByHash` and `getScriptByHash` queries
Includes #1366: `utxosAt` and  `getUtxoByOref` queries
Includes #1372: `getProtocolParameters` and `getCurrentEpoch` queries
Includes #1376: `submitTx` and `evaluateTx` queries
Includes #1377: `getChainTip`, `getEraSummaries`, and `getSystemStart` queries
Includes #1389: Revised Contract testing utilities
Includes #1401: Various fixes, bring back `submitE`, trace log blockfrost requests
includes #1412: Introduce `confirmTxDelay` to smooth over issues with Blockfrost API

Notable changes:
 - ConfigParams r -> ContractParams (outside of QueryM)
 - Contract r -> Contract
 - Removed unused bindings for ODC and Ogmios. Namely utxo lookup in Ogmos, and Datum fetching in ODC. We no longer use them and are not exposed in the API.
   - ODC has been removed completely
   - Postgres too
     - **This will only be for the CTL backend, it turns out blockfrost requires it**
 - Fleshed out the docs a little to use the new types/contract params, and mention there's multiple backends.
 - ~getTxByHash is currently CTLBackend only. Will be removed when Kupo replaces ODC fully.~
   - ~Actually this PR uses kupo for isTxConfirmed, so we can do so now. But I'm not sure I trust the implementation at the moment (tracked in https://github.com/Plutonomicon/cardano-transaction-lib/issues/1293)~
 - `getTxByHash` has been removed, and `getTxMetadata` has been added. This, as well as utxo, datum and script lookup, should cover all cases needed for getting a full transaction.
 - ~KeyWallet now receives a networkId on construction (for Contract based construction, this is asked)~
   - ~It makes no sense to re-use a wallet for multiple networks, and this simplifies implementation~
   - ~mkKeyWalletFromFiles/PrivateKeys is now in Contract and returns a KeyWallet instead of Wallet~
   - Reverted the above, except for mkKeyWalletFromFiles still returns KeyWallet
 - reindexSpentScriptRedeemers is no longer in Contract (it's pure)
 - Lots of logic that was previously QueryM is now in Contract, polymorphic to the backend.
 - ~Removed `submitE` - submission is only in maybe internally.~ Readded
 - All 'cardano' actions are now implemented via the abstract handlers, other than those relating to staking
 - Test.Ctl.AffInterface -> Test.Ctl.QueryM.AffInterface
 - Moved `ProtocolParameters`, `EraSummaries`, `Epoch`, and `SystemStart`to `Ctl.Internal.Types`
 - Moved `ClientError` to `Ctl.Internal.Service.Error`

Remaining issues:
 - ~ContractEnv/QueryHandle mention Ogmios types.~
   - ~Not a huge issue, but conceptually a bit odd.~
   - ~ProtocolParameters (and EraSummaries) should ideally be placed in Cardano.Types. One issue is the Epoch type, which already exists in Cardano.Types.Transaction. The underlying representations are different however.~
   - ~One problem with ProtocolParameters and EraSummaries, is the responses are different between Ogmios and Blockfrost, but I think only by camel case vs snake case. I don't think this is the case for tx evaluation though, which is explicitly implemented via ogmios by all backends.~
 - ~Similar to other types, ClientError is defined by QueryM yet is used in Contract~
 - In general I think backend-agnostic error responses/handling may need more thought
 - No metadata submission + fetching example
 - ~There's an issue with log supression in the Plutip tests~

### Pre-review checklist


doc/side-by-side-ctl-plutus-comparison.md
```diff
@@ -73,18 +44,11 @@ The Plutus `Contract` environment is specialized to just two values and is fixed
 Also, Plutus `Contract` uses a phantom type `s` to contract schema
```
rynoV, line 44:
```suggestion
Also, Plutus `Contract` uses a phantom type `s` for the contract schema
```

rynoV (changes requested):
I'm getting a [failing integration test](https://github.com/Plutonomicon/cardano-transaction-lib//blob/b21a403cadd33d2d9ac35dc6d76cc19df79136f3/test/Logging.purs#L34):

```
Suppressed logs do not appear when no error:

  Error: true ≠ false
    at Object.exports.error (/home/calum/code/mlabs/cardano-transaction-lib/output/Effect.Excep
tion/foreign.js:8:10)
    at /home/calum/code/mlabs/cardano-transaction-lib/output/Test.Spec.Assertions/index.js:15:3
7
    at /home/calum/code/mlabs/cardano-transaction-lib/output/Test.Spec.Assertions/index.js:38:1
46
    at /home/calum/code/mlabs/cardano-transaction-lib/output/Test.Ctl.Logging/index.js:60:141
    at run (/home/calum/code/mlabs/cardano-transaction-lib/output/Effect.Aff/foreign.js:278:22)
    at /home/calum/code/mlabs/cardano-transaction-lib/output/Effect.Aff/foreign.js:348:19
    at drain (/home/calum/code/mlabs/cardano-transaction-lib/output/Effect.Aff/foreign.js:120:9
)
    at Object.enqueue (/home/calum/code/mlabs/cardano-transaction-lib/output/Effect.Aff/foreign
.js:141:11)
    at /home/calum/code/mlabs/cardano-transaction-lib/output/Effect.Aff/foreign.js:339:27
    at __do (/home/calum/code/mlabs/cardano-transaction-lib/output/Ctl.Internal.QueryM/index.js
:260:170)
```

For some reason [the corresponding plutip test](https://github.com/Plutonomicon/cardano-transaction-lib//blob/b21a403cadd33d2d9ac35dc6d76cc19df79136f3/test/Plutip/Logging.purs#L36) doesn't fail.

The rest of the tests are passing.

src/Contract/Chain.purs
```diff
@@ -27,17 +25,17 @@ import Ctl.Internal.Types.Chain
...
 
-getTip :: forall (r :: Row Type). Contract r Chain.Tip
-getTip = wrapContract QueryM.getChainTip
+getTip :: Contract Chain.Tip
+getTip = Contract.getChainTip
 
-waitUntilSlot :: forall (r :: Row Type). Slot -> Contract r Chain.Tip
-waitUntilSlot = wrapContract <<< QueryM.waitUntilSlot
+waitUntilSlot :: Slot -> Contract Chain.Tip
+waitUntilSlot = Contract.waitUntilSlot
```
Luis-omega, line 32: Can we just export the original function as we don't need to `wrap` them anymore?

src/Contract/Wallet.purs
```diff
@@ -20,18 +20,18 @@ module Contract.Wallet
...
+import Control.Monad.Reader.Class (asks)
+import Ctl.Internal.Contract.Wallet
   ( getChangeAddress
-  , getNetworkId
   , getRewardAddresses
   , getUnusedAddresses
   , getWallet
   , signData
-  ) as QueryM
+  ) as Contract
```
Luis-omega, line 33: as before, `Contract.Wallet` or `Wallet` is the convention for the import unless this is done only for export.

src/Contract/Wallet/KeyFile.purs
```diff
@@ -6,8 +6,10 @@ module Contract.Wallet.KeyFile
 
 import Prelude
 
-import Ctl.Internal.Wallet (Wallet) as Wallet
-import Ctl.Internal.Wallet (mkKeyWallet)
+import Control.Monad.Reader.Class (asks)
+import Ctl.Internal.Contract.Monad (Contract)
+import Ctl.Internal.Wallet.Key (KeyWallet) as Wallet
```
Luis-omega, line 11: as `Wallet.Key` or `Key`

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -103,13 +104,10 @@ import Ctl.Internal.Cardano.Types.Value
...
-import Ctl.Internal.QueryM (QueryM, getProtocolParameters)
-import Ctl.Internal.QueryM (getChangeAddress, getWalletAddresses) as QueryM
-import Ctl.Internal.QueryM.Utxos
-  ( filterLockedUtxos
-  , getWalletCollateral
-  , utxosAt
-  )
+import Ctl.Internal.Contract.Monad (Contract, filterLockedUtxos)
+import Ctl.Internal.Contract.QueryHandle (getQueryHandle)
+import Ctl.Internal.Contract.Wallet (getChangeAddress, getWalletAddresses) as Contract
```
Luis-omega, line 109: same as before

src/Internal/BalanceTx/ExUnitsAndMinFee.purs
```diff
@@ -42,10 +41,9 @@ import Ctl.Internal.Cardano.Types.Transaction
   , _redeemers
   , _witnessSet
   )
+import Ctl.Internal.Contract.MinFee (calculateMinFee) as Contract
```
Luis-omega, line 44: same as before about imports

src/Internal/Contract/ApplyArgs.purs
```diff
@@ -0,0 +1,87 @@
...
+        ClientOtherError
+          "The `ctl-server` service is required to call `applyArgs`. Please \
+          \provide a `Just` value in `ContractParams.ctlServerConfig` and make \
+          \sure that the `ctl-server` service is running and available at the \
+          \provided host and port. The `ctl-server` packages can be obtained \
+          \from `overlays.ctl-server` defined in CTL's flake. Please see \
+          \`doc/runtime.md` in the CTL repository for more information"
+    Just config -> case traverse plutusDataToAeson args of
+      Nothing -> pure $ Left $ ClientEncodingError
+        "Failed to convert script args"
```
Luis-omega, line 60: It would be useful for users to know the argument that couldn't be converted and this is the right point to do it.

Luis-omega (changes requested):
Only review the first half of it, most comments are bout import style and all of them are about the aftereffects of the change rather than the core of the change. I will continue tomorrow. 

It's impressive how this change simplifies a lot of calls!

src/Contract/Address.purs
```diff
@@ -38,13 +38,20 @@ module Contract.Address
...
   ( addressPaymentValidatorHash
   , addressStakeValidatorHash
   ) as Address
+import Ctl.Internal.Contract.Wallet
+  ( getNetworkId
+  , getWalletAddresses
+  , getWalletCollateral
+  , ownPaymentPubKeyHashes
+  , ownStakePubKeysHashes
+  ) as Contract
```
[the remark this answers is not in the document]
jy14898: It's not a re-export, we re-define all the functions again in the module. They probably should be re-exports, however

src/Contract/PlutusData.purs
```diff
@@ -107,30 +103,33 @@ import Ctl.Internal.Types.Redeemer
...
 
--- | Get `PlutusData`s given an `Array` of `DataHash`.
--- | This function discards all possible error getting a `DataHash`.
-getDatumsByHashes
-  :: forall (r :: Row Type)
-   . Array DataHash
-  -> Contract r (Map DataHash Datum.Datum)
-getDatumsByHashes = wrapContract <<< QueryM.getDatumsByHashes
+-- | Retrieve full resolved datums associated with given datum hashes.
+-- | The resulting `Map` will only contain datums that have been successfully
```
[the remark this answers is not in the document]
jy14898: The change originates from this PR: https://github.com/Plutonomicon/cardano-transaction-lib/pull/1235

It's not impossible for Kupo to replicate ODCs behaviour here (we just call getDatumByHash in parallel), and actually it seems blockfrost also can only query by one hash.

@errfrom What was the motivation for the change?

src/Internal/Contract/ApplyArgs.purs
jy14898: This code is mostly just moved as-is from `Ctl.Internal.QueryM.applyArgs`, but contractified. I agree we need better errors but I think this suggestion is better as a separate issue

src/Contract/PlutusData.purs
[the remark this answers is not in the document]
errfrom: Agreed, we should keep ODCs behaviour here.
Initially I thought it would make sense to fail on the first error and that we don't need to explicitly represent the absence of `Datum` in the `Map`, as this can be easily checked by the user.

src/Contract/ProtocolParameters.purs
```diff
@@ -2,11 +2,11 @@ module Contract.ProtocolParameters
...
+import Contract.Monad (Contract)
+import Ctl.Internal.Contract (getProtocolParameters) as Contract
 import Ctl.Internal.QueryM.Ogmios (ProtocolParameters)
 
 -- | Returns the `ProtocolParameters` from the `Contract` environment.
 -- | Note that this is not necessarily the current value from the ledger.
-getProtocolParameters :: forall (r :: Row Type). Contract r ProtocolParameters
-getProtocolParameters = wrapContract QueryM.getProtocolParameters
+getProtocolParameters :: Contract ProtocolParameters
+getProtocolParameters = Contract.getProtocolParameters
```
errfrom, line 12: Why do we keep `getProtocolParameters` in a separate module?

src/Contract/PlutusData.purs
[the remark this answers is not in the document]
jy14898: I think we should just drop the plural handle and implement the traversal in Contract

jy14898:
@Luis-omega I'm going to have to revert some of the import/export changes, because there's scenarios where we import a module qualified with the same name as the module we're currently defining, which ends up exporting nothing from the qualified import.

test/OgmiosDatumCache.purs
```diff
@@ -27,31 +23,14 @@ import Test.Spec.Assertions (shouldEqual)
 
 suite :: TestPlanM (Aff Unit) Unit
 suite = group "Ogmios Datum Cache tests" $ do
+  -- TODO Move
```
amirmrad, line 26: Is TODO resolved?

test/OgmiosDatumCache.purs
```diff
@@ -27,31 +23,14 @@ import Test.Spec.Assertions (shouldEqual)
...
-getDatumsByHashesHashingTest
-  :: forall (m :: Type -> Type)
-   . MonadEffect m
-  => MonadThrow Error m
-  => m Unit
-getDatumsByHashesHashingTest = do
-  datums <- Map.toUnfoldable <<< unwrap <$> readGetDatumsByHashesSample
-  for_ (datums :: Array _) \(hash /\ datum) -> do
-    (datumHash <$> datum) `shouldEqual` Right (Just hash)
+-- TODO Add GetTxByHash
```
amirmrad, line 33: Is TODO resolved?

src/Internal/Contract/QueryBackend.purs
```diff
@@ -0,0 +1,156 @@
...
+--------------------------------------------------------------------------------
+-- QueryBackends
+--------------------------------------------------------------------------------
+
+-- | A generic type to represent a choice of backend with a set of fallback
+-- | backends when an operation is not supported by the default.
+data QueryBackends (backend :: Type) =
+  QueryBackends backend (Map QueryBackendLabel backend)
+
+-- Functor breaks this datatype...
```
rynoV, line 41: What does this mean?

src/Internal/Contract/QueryBackend.purs
```diff
@@ -0,0 +1,156 @@
...
+
+mkBlockfrostBackendParams :: ServerConfig -> QueryBackends QueryBackendParams
+mkBlockfrostBackendParams = mkSingletonBackendParams <<< BlockfrostBackendParams
+  <<< { blockfrostConfig: _ }
+
+mkBackendParams
+  :: QueryBackendParams
+  -> Array QueryBackendParams
+  -> Effect (QueryBackends QueryBackendParams)
+mkBackendParams defaultBackend' backends =
```
rynoV, line 73: These `mk*BackendParams` functions return `QueryBackends`, might be clearer to rename them to `mk*QueryBackends`. Or `mk*Backend` might make sense because we refer to `QueryBackends QueryBackend` as `backend` [here](https://github.com/Plutonomicon/cardano-transaction-lib//blob/5857e5eb826b27ba484c6ef36ee5ff3b8484a072/src/Internal/Contract/Monad.purs#L166 "ContractEnv type")

src/Internal/Contract/QueryBackend.purs
```diff
@@ -0,0 +1,156 @@
...
+  case length backends + 1 /= numUniqueBackends of
+    true ->
+      throw "mkBackendParams: multiple configs for the same service"
+    false ->
+      pure $ QueryBackends defaultBackend' $
+        foldl (\mp b -> Map.insert (backendLabel b) b mp) Map.empty backends
```
rynoV, line 79: Strange style when if-then-else will do

src/Internal/Contract/Monad.purs
```diff
@@ -0,0 +1,400 @@
...
+--------------------------------------------------------------------------------
+-- Contract
+--------------------------------------------------------------------------------
+
+-- | The `Contract` monad is a newtype wrapper over `ReaderT` on `ContractEnv`
+-- | over asynchronous effects, `Aff`. Throwing and catching errors can
+-- | therefore be implemented with native JavaScript `Effect.Exception.Error`s
+-- | and `Effect.Class.Console.log` replaces the `Writer` monad. `Aff` enables
+-- | the user to make effectful calls inside this `Contract` monad.
+newtype Contract (a :: Type) = Contract (ReaderT ContractEnv Aff a)
```
rynoV, line 99: Contract no longer has `Alt` or `Plus` instances, is this intended? If so we should remember to add it to the changelog

rynoV (changes requested):
A few things, still reviewing

src/Internal/Contract/Monad.purs
```diff
@@ -0,0 +1,400 @@
...
+  , logLevel :: LogLevel
+  , walletSpec :: Maybe WalletSpec
+  , customLogger :: Maybe (LogLevel -> Message -> Aff Unit)
+  , suppressLogs :: Boolean
+  , hooks :: Hooks
+  , wallet :: Maybe Wallet
+  , usedTxOuts :: UsedTxOuts
+  -- ledgerConstants are values that technically may change, but we assume to be
+  -- constant during Contract evaluation
+  , ledgerConstants ::
```
Luis-omega, line 179: In principle, we shouldn't have problems with this unless someone tries to use a test blockchain with dynamic changes in protocol parameters. So we should at least add a little warning in the documentation of some functions like `waitUntilSlot`

Luis-omega:
@jy14898 about the exports, I like what you did for them prefixing them with `Export`, still we have in #56 the following 


`If the goal of an import is just to re-export, use import Foo.Bar (baz) as X, not import Foo.Bar (baz) as Foo.Bar`

At first, I thought it was referring `X` as in your sense of `use whatever name is clear,` but while reviewing the code I found some cases when we really do `import SomethingModule(Name1, ... ) as X` using literally `X` as the name for the import.

test/OgmiosDatumCache.purs
jy14898: Not quite the same, the AffInterface only checks for any response from the server, not a parse of the success case. That being said, we only used it for transaction confirmation, and now we don't use it at all, due to Kupo replacing it. We should strip out all ODC stuff imo

src/Internal/Contract/QueryBackend.purs
jy14898: Addressed by https://github.com/Plutonomicon/cardano-transaction-lib/pull/1314

src/Internal/Contract/QueryBackend.purs
jy14898: I think this is not longer relevant since https://github.com/Plutonomicon/cardano-transaction-lib/pull/1314

src/Internal/Contract/QueryBackend.purs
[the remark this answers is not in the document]
jy14898: Addressed by https://github.com/Plutonomicon/cardano-transaction-lib/pull/1314

src/Internal/Contract/QueryBackend.purs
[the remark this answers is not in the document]
jy14898: Yes, it's not the case the user would switch backends during runtime, but for fallbacks https://mlabs-corp.slack.com/archives/C02QXGRGAPK/p1669375865559569

src/Internal/Contract/Monad.purs
klntsky: Why no `Alt` anymore?

test/OgmiosDatumCache.purs
jy14898: I can't remember my intention, probably to move it into Test.Ctl.QueryM. However, we no longer need ODC so I'll be removing this code

src/Internal/Contract/Monad.purs
jy14898: Oversight, this has been added back

src/Internal/QueryM.purs
```diff
@@ -49,6 +44,7 @@ module Ctl.Internal.QueryM
 
 import Prelude
 
+-- import Ctl.Internal.QueryM.Kupo (isTxConfirmed)
```
rynoV, line 47: Can we remove this?

rynoV (changes requested):
Some more ODC references to possibly remove/deal with:

<details>
<summary>Resolved</summary>
```
src/Internal/Types/PlutusData.purs:49:-- Ogmios Datum Cache Json format
examples/Datums.purs:9:-- |   * start the block-fetcher for `ogmios-datum-cache`, e.g.
Makefile:3:.PHONY: run-dev run-build check-format format run-datum-cache-postgres-console
Makefile:52:run-datum-cache-postgres-console:
nix/test-nixos-configuration.nix:41:  services.ogmios-datum-cache = {
nix/default.nix:131:                  pkgs.ogmios-datum-cache
nix/default.nix:245:  #  - `ogmios-datum-cache`
nix/default.nix:268:          ogmios-datum-cache
nix/default.nix:350:          ogmios-datum-cache
nix/default.nix:373:        export OGMIOS_DATUM_CACHE_PORT=10005
README.md:71:   - `ogmios-datum-cache` solves this problem
flake.nix:38:    ogmios-datum-cache.url = "github:mlabs-haskell/ogmios-datum-cache/862c6bfcb6110b8fe816e26b3bba105dfb492b24";
flake.nix:40:    # ogmios and ogmios-datum-cache nixos modules (remove and replace with the above after merging and updating)
flake.nix:42:    ogmios-datum-cache-nixos.url = "github:mlabs-haskell/ogmios-datum-cache/marton/nixos-module";
flake.nix:303:                ogmios-datum-cache =
flake.nix:304:                  inputs.ogmios-datum-cache.defaultPackage.${system};
flake.nix:529:          inputs.ogmios-datum-cache-nixos.nixosModules.ogmios-datum-cache
flake.nix:531:            services.ogmios-datum-cache.package =
flake.nix:532:              inputs.ogmios-datum-cache.packages.x86_64-linux."ogmios-datum-cache";
```
</details>

src/Internal/Contract/Monad.purs
Luis-omega: I mean that the most realistic case when this can be a problem for the users is when they are using a custom network and doing something like updating the network after the initialization of the Contract env. I don't think this would be a major problem but maybe we need to document that it could happen.

src/Internal/Contract/ApplyArgs.purs
Luis-omega: is now #1319

src/Internal/Contract/Monad.purs
jy14898: I think adding comments to functions which depend on a protocol parameters would be messy, the balancer is another example, there's quite a few more. I'm not sure how other tools deal with it, but I think we could at least add it to the general docs/faq of CTL. Note that this behaviour is already what happens in `develop`, so documentation about this doesn't depend on this PR

src/Internal/QueryM.purs
jy14898: Yeah, it was just a leftover from trying to import isTxConfirmed directly - it caused import cycles, so I just went with passing the function as an arg, similar to how we did with ODC's ws already

src/Internal/Contract/Monad.purs
Bradley-Heather: This is very much an edge case scenario and so I agree that commenting the actual code is not needed.

src/Internal/Wallet.purs
```diff
@@ -280,3 +314,97 @@ dummySign tx@(Transaction { witnessSet: tws@(TransactionWitnessSet ws) }) =
...
+  actionBasedOnWallet _.getWalletAddresses
+    (\kw -> pure $ pure $ Array.singleton (unwrap kw).address)
+    wallet
+
+signData :: Address -> RawBytes -> Wallet -> Aff (Maybe DataSignature)
+signData address payload wallet = do
+  actionBasedOnWallet
+    (\w conn -> w.signData conn address payload)
+    (\kw -> pure <$> (unwrap kw).signData payload)
+    wallet
```
rynoV, line 360:
```suggestion
signData :: Address -> RawBytes -> Wallet -> Aff (Maybe DataSignature)
signData address payload =
  actionBasedOnWallet
    (\w conn -> w.signData conn address payload)
    (\kw -> pure <$> (unwrap kw).signData payload)
```

src/Internal/Wallet.purs
```diff
@@ -280,3 +314,97 @@ dummySign tx@(Transaction { witnessSet: tws@(TransactionWitnessSet ws) }) =
...
+getUnusedAddresses wallet = fold <$> do
+  actionBasedOnWallet _.getUnusedAddresses
+    (\_ -> pure $ pure [])
+    wallet
+
+getChangeAddress :: Wallet -> Aff (Maybe Address)
+getChangeAddress wallet = do
+  actionBasedOnWallet _.getChangeAddress
+    (\kw -> pure $ pure (unwrap kw).address)
+    wallet
```
rynoV, line 341:
```suggestion
getChangeAddress =
  actionBasedOnWallet _.getChangeAddress
    (\kw -> pure $ pure (unwrap kw).address)
```

src/Internal/Wallet.purs
```diff
@@ -280,3 +314,97 @@ dummySign tx@(Transaction { witnessSet: tws@(TransactionWitnessSet ws) }) =
               "ed25519_sig1ynufn5umzl746ekpjtzt2rf58ep0wg6mxpgyezh8vx0e8jpgm3kuu3tgm453wlz4rq5yjtth0fnj0ltxctaue0dgc2hwmysr9jvhjzswt86uk"
           )
     )
+
+getNetworkId :: Wallet -> Aff NetworkId
+getNetworkId wallet = do
+  actionBasedOnWallet
+    (\w -> intToNetworkId <=< _.getNetworkId w)
+    (\kw -> pure (unwrap kw).networkId)
+    wallet
```
rynoV, line 323:
```suggestion
getNetworkId =
  actionBasedOnWallet
    (\w -> intToNetworkId <=< w.getNetworkId)
    (\kw -> pure (unwrap kw).networkId)
```

src/Internal/Wallet.purs
```diff
@@ -280,3 +314,97 @@ dummySign tx@(Transaction { witnessSet: tws@(TransactionWitnessSet ws) }) =
...
+ownPubKeyHashes :: Wallet -> Aff (Array PubKeyHash)
+ownPubKeyHashes wallet = Array.catMaybes <$> do
+  getWalletAddresses wallet >>= traverse \address -> do
+    paymentCred <-
+      Helpers.liftM
+        ( error $
+            "Unable to get payment credential from Address"
+        ) $
+        addressPaymentCred address
+    pure $ stakeCredentialToKeyHash paymentCred <#> wrap
```
rynoV, line 388: Seems strange that we're dealing with stake credentials here. Seems to stem from `Internal.Serialization.Address` using `StakeCredential` to [represent both payment and delegation credentials](https://github.com/Plutonomicon/cardano-transaction-lib//blob/c2dde03df26b16501961851b959dba00faae06cf/src/Internal/Serialization/Address.purs#L372-L373). Maybe it's just different naming than I'm used to

rynoV (changes requested):
I wasn't sure how detailed we wanted the changelog, but I made notes for changes to the public interface.

package.json
```diff
@@ -34,6 +34,7 @@
     "big-integer": "1.6.51",
     "blakejs": "1.2.1",
     "bufferutil": "4.0.5",
+    "doctoc": "^2.2.1",
```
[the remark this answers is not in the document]
jy14898: I think this change is in develop now

src/Internal/Contract/WaitUntilSlot.purs
```diff
@@ -149,27 +157,30 @@ waitNSlots offset = do
...
-  futureTime <- liftEffect $ slotToPosixTime eraSummaries sysStart futureSlot
-    >>= hush >>> liftM (error "Unable to convert Slot to POSIXTime")
+  { systemStart } <- asks _.ledgerConstants
+  queryHandle <- getQueryHandle
+  eraSummaries <- liftAff $ queryHandle.getEraSummaries
+  futureTime <- liftEffect $
+    slotToPosixTime eraSummaries systemStart futureSlot
+      >>= hush >>> liftM (error "Unable to convert Slot to POSIXTime")
   -- We assume that a slot is 1000 milliseconds here.
+  -- TODO Don't
```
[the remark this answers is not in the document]
jy14898: Yeah it's based off of the previous line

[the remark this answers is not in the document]
jy14898: I mean it's not a good comment lol

src/Internal/Wallet.purs
jy14898: Originates from https://github.com/Emurgo/cardano-serialization-lib/commit/9e86bff63c1c87c7fc6736c2a7c620f9c2b7fae8
We reuse the names CSL uses, and CSL seems to prefer using types defined in the CDDL. I'm not sure if that's the right choice on their end but whatever

src/Contract/Config.purs
```diff
@@ -32,7 +32,9 @@ import Ctl.Internal.Contract.Hooks (emptyHooks)
 import Ctl.Internal.Contract.Monad (ContractParams)
 import Ctl.Internal.Contract.QueryBackend
   ( QueryBackendParams(CtlBackendParams, BlockfrostBackendParams)
-  , mkBlockfrostBackendParams
+  -- TODO Export once the following is stable
+  -- https://github.com/Plutonomicon/cardano-transaction-lib/issues/1118
+  -- , mkBlockfrostBackendParams
```
rynoV, line 37: I think we need to do the same for `BlockfrostBackendParams`

rynoV (changes requested):

src/Internal/Contract/WaitUntilSlot.purs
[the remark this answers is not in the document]
rynoV: Oh I see haha

src/Internal/QueryM/Kupo.purs
```diff
@@ -405,6 +400,54 @@ instance DecodeAeson KupoScriptRef where
...
+  show = genericShow
+
+instance DecodeAeson KupoUtxoSlot where
+  decodeAeson = decodeAeson >>> map (slot >>> KupoUtxoSlot)
+    where
+    slot :: { created_at :: { slot_no :: Slot } } -> Slot
+    slot = _.created_at.slot_no
+
+unwrapKupoUtxoSlot :: KupoUtxoSlot -> Slot
+unwrapKupoUtxoSlot (KupoUtxoSlot slot) = slot
```
rynoV, line 422: Why not just add a `Newtype` instance?

src/Internal/QueryM/Kupo.purs
```diff
@@ -405,6 +400,54 @@ instance DecodeAeson KupoScriptRef where
...
+        TypeMismatch "Hexadecimal String"
+      metadata <- flip note (fromBytes ba) $
+        TypeMismatch "Hexadecimal encoded Metadata"
+      -- Conversion should always succeed, so use it as the `Just`
+      pure $ KupoMetadata $ hush $ convertGeneralTransactionMetadata metadata
+    [] -> Right $ KupoMetadata Nothing
+    _ -> Left $ TypeMismatch "Singleton or Empty Array"
+
+unwrapKupoMetadata :: KupoMetadata -> Maybe GeneralTransactionMetadata
+unwrapKupoMetadata (KupoMetadata mbMetadata) = mbMetadata
```
rynoV, line 449: Same as above

rynoV (dismissed):

src/Internal/QueryM/Kupo.purs
jy14898: Wasn't really worth it, the type only exists to attach decoding instances to, if we use unwrap we have to give a type hint

jy14898: eg 

```
unwrapKupoUtxoSlot a
```
vs
```
unwrap (a :: KupoUtxoSlot)
```

src/Contract/Wallet/KeyFile.purs
```diff
@@ -29,8 +37,21 @@ import Node.Path (FilePath)
 -- |
 -- | **NodeJS only**
 mkKeyWalletFromFiles
-  :: FilePath -> Maybe FilePath -> Aff Wallet.Wallet
+  :: FilePath -> Maybe FilePath -> Contract KeyWallet
```
klntsky, line 40: This function should be turned back to the `Aff` type. The below one can be removed.

It was in `Aff` because a wallet is needed to initialize the `Contract`, so users are expected to call it before `Contract`.

klntsky (changes requested):

src/Contract/Wallet/KeyFile.purs
jy14898: > a wallet is needed to initialize the `Contract`

Not really true, you can have no wallet in a contract and use `withKeyWallet` once the wallet is constructed, although perhaps not obvious to users

This change occurred after I made `KeyWallet` require a `NetworkId` upon construction. This made code elsewhere more streamlined, but also conceptually, why would we need a key wallet to be polymorphic to network after construction?

Keeping it in `Contract` helps avoid issues with wrong `NetworkId`s

klntsky: > why would we need a key wallet to be polymorphic to network after construction?

Because it's just a wrapped Ed25519 private key . What exactly do we need the network ID for? If it's just to tag it, I'm strongly against this change.

> Not really true, you can have no wallet in a contract and use withKeyWallet once the wallet is constructed

OK, but then there's another argument: we should not ask the user to provide more arguments than we need to perform the computation. `Contract` environment is not only the network ID parameter, but also the runtime handles, that we should not use. We don't want to force the users to initialize it just to get a KeyWallet (e.g. for later use in Contract, or just to call `signData`).

jy14898: Okay, I was thinking `KeyWallet` existed mostly for consumption in `Contract`, so it made sense to make it similar to other wallets (which are also bound to a network).

> If it's just to tag it, I'm strongly against this change.

It's not just as a tag, it also removes the need to pass in the network id every time we want the key wallets `address` or to `signData`.

I thought the `Aff` version would be sufficient to handle the non-contract cases, but I'm happy to revert the change

test/Plutip/Utils.purs
```diff
@@ -10,31 +10,25 @@ import Contract.Monad (Contract)
...
   )
 import Control.Monad.Reader (asks)
 import Ctl.Internal.Types.UsedTxOuts (TxOutRefCache)
-import Data.Maybe (Maybe(Just), isNothing)
 import Data.Newtype (unwrap)
 import Effect.Class (liftEffect)
-import Effect.Exception (throw)
 import Effect.Ref as Ref
 
+-- TODO: Get everything we can about a tx and confirm them
```
uhbif19, line 21: So this is functionality killed and waiting to be redone with new metadata function?

jy14898: Yeah it no longer verifies the transaction. I don't think it's a critical piece of code though. What exactly it should do isn't really clear by the name

src/Contract/Wallet/KeyFile.purs
jy14898: Reverted

pb99u069:
We cleaned up the `testClusterDir` in `startPostgresServer` on sigint, but since postgres no longer exists I moved the cleanup into `startKupo`. hope thats ok

jy14898:
> We cleaned up the `testClusterDir` in `startPostgresServer` on sigint, but since postgres no longer exists I moved the cleanup into `startKupo`. hope thats ok

Thanks!

examples/Datums.purs
```diff
@@ -1,6 +1,7 @@
--- | An example of fetching datums from `ogmios-datum-cache`. Helpful to test
--- | out the datum-cache integration
+-- | An example of fetching datums from `Kupo` or `Blockfrost`. Helpful to test
+-- | out the backend integration
 -- |
+-- TODO Rewrite this example or remove it
```
uhbif19, line 4: @jy14898 Why rewrite?

jy14898: It assumes a certain state of the chain + datum indexer. I suppose as an example it's fine, but as a test I don't think it's great

uhbif19:
@klntsky @jy14898 As far as I understand now `runPlutipContract` uses Ctl Backend inside while construction `ContractEnv`, so maybe we need to parametrise it by used backend, to reuse this function for Blockfrost testing?

uhbif19:
@klntsky @kirill-havryliuk Also, I think all `aesonX` helper functions use wrong `TypeMismatch` error messages

```
aesonObject = caseAesonObject (Left (TypeMismatch "Expected Object"))
```

Also looks like many of them they are duplicated in different modules. They maybe moved to `purescript-aeson` I think.

src/Internal/Contract/QueryBackend.purs
```diff
@@ -0,0 +1,69 @@
...
+getBlockfrostBackend :: QueryBackend -> Maybe BlockfrostBackend
+getBlockfrostBackend (CtlBackend _ backend) = backend
+getBlockfrostBackend (BlockfrostBackend backend _) = Just backend
+
+--------------------------------------------------------------------------------
+-- QueryBackendParams
+--------------------------------------------------------------------------------
+
+data QueryBackendParams
+  = CtlBackendParams CtlBackendParams (Maybe BlockfrostBackendParams)
```
uhbif19, line 53: Why do we need this `Maybe` param?

src/Internal/QueryM.purs
```diff
@@ -405,159 +291,6 @@ instance Parallel (QueryMExtended r ParAff) (QueryMExtended r Aff) where
   sequential :: QueryMExtended r ParAff ~> QueryMExtended r Aff
```
uhbif19, line 291: Looks like:

```
type QueryEnv (r :: Row Type) =
  { config :: QueryConfig
  , runtime :: QueryRuntime
  , extraConfig :: { | r }
  }
  ```
  
  is the only place where `extraConfig` got left, and it is not used.
  
  Also `QueryMExtended` is not used anymore.
  
  Can we remove them all?

**Review by uhbif19 — APPROVED** [review]

File: src/Contract/Transaction.purs [hunk 33]
```diff
@@ -537,42 +491,44 @@ getTxFinalFee :: BalancedSignedTransaction -> BigInt
...
-  :: forall (r :: Row Type)
-   . Seconds
+  :: Seconds
   -> TransactionHash
-  -> Contract r Unit
-awaitTxConfirmedWithTimeout timeout = wrapContract
-  <<< AwaitTx.awaitTxConfirmedWithTimeout timeout
-  <<< unwrap
+  -> Contract Unit
+awaitTxConfirmedWithTimeout = Contract.awaitTxConfirmedWithTimeout
```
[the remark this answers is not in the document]
rynoV: I'm guessing we're doing it this way to keep all the documentation in one place where it's easier to find

src/Internal/Contract/QueryBackend.purs
rynoV: The idea is to allow for the possibility of a "backup" backend, for any operations the main backend can't perform

src/Contract/Transaction.purs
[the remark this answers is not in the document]
jy14898: I'm not against removing them, making sure documentation (comments) is preserved. We have quite a lot of functions defined like this now, mostly because it was easier to transform them directly

src/Contract/Scripts.purs
```diff
@@ -51,3 +69,24 @@ import Ctl.Internal.Types.TypedValidator
...
   ) as TypedValidator
+import Data.Either (Either)
+import Data.Map (Map)
+import Data.Map as Map
+import Data.Maybe (Maybe)
+import Data.Tuple (Tuple(Tuple))
+import Effect.Aff.Class (liftAff)
+
+-- | Retrieve a `ScriptRef` given the hash
+getScriptByHash :: ScriptHash -> Contract (Either ClientError (Maybe ScriptRef))
```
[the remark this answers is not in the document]
jy14898: Maybe, I don't like throwing away error information though. I suppose either way in this case the user can't move as they presumably needed the script to submit some transaction? To be honest I'm not even sure of the use 

It only got added due to the query handle being inspired by those of Kupo. We don't use the query handle internally (maybe we should, like we do with datum?)

src/Contract/Wallet.purs
```diff
@@ -69,56 +64,20 @@ import Ctl.Internal.Wallet.Spec
...
-signData
-  :: forall (r :: Row Type)
-   . Address
-  -> RawBytes
-  -> Contract r (Maybe DataSignature)
-signData address dat = wrapContract (QueryM.signData address dat)
-
-getWallet :: forall (r :: Row Type). Contract r (Maybe Wallet)
-getWallet = wrapContract QueryM.getWallet
+getNetworkId :: Contract NetworkId
```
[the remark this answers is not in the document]
jy14898: The one here is left-over from when we didn't enforce during contract setup that the wallet network id matched that of the contract. I guess we kept it for API compatibility but we should remove it now due to all other changes

src/Internal/QueryM/ServerConfig.purs
```diff
@@ -1,11 +1,9 @@
 module Ctl.Internal.QueryM.ServerConfig
```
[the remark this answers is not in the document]
jy14898: Ideally yes, just hasn't happened yet

src/Internal/QueryM.purs
jy14898: Not exactly, `QueryM` and `ParQueryM` are aliases of `QueryMExtended`, however we can remove `r` and `extraConfig`. I also renamed `QueryMExtended` to `QueryMT` to indicate it is a monad transformer of some base monad; `Aff` for `QueryM` and `ParAff` for `ParQueryM`.

jy14898: We could make `QueryM` and `ParQueryM` `data` instead of `type`, but none of this faces the user and it's not going to save us much

uhbif19: `ParQueryM` is not used too, as far as I see.

uhbif19:
@errfrom @jy14898 @klntsky Do you know where these stacking tests error CI comes from? I found two kinds of them - timeouts and wrong result:

https://hercules-ci.com/accounts/github/Plutonomicon/derivations/%2Fnix%2Fstore%2Fps5fw22wdabi28idyrsjsr2cjnwy3p46-ctl-staking-test.drv/log?via-job=54c4ae20-7479-42ef-b0b7-f1d59c2449e1
https://hercules-ci.com/accounts/github/Plutonomicon/derivations/%2Fnix%2Fstore%2Fps5fw22wdabi28idyrsjsr2cjnwy3p46-ctl-staking-test.drv/log?via-job=54c4ae20-7479-42ef-b0b7-f1d59c2449e1

src/Internal/QueryM.purs
jy14898: > But ParQueryM is not used too, as far as I see.

Good point

jy14898: > Good point

Nevermind, `ParQueryM` is used indirectly at https://github.com/Plutonomicon/cardano-transaction-lib/blob/cad5b9c7341c33d4df6e3a926148cf4cd38d03b0/src/Internal/QueryM/Kupo.purs#L302

We could still remove `ParQueryM` and unwrap `QueryM` at that point to use `Aff`'s `parTraverse`, and then wrap again, but it doesn't really matter

jy14898:
@uhbif19 I got a notification that you mentioned me, but I cannot find the comment anywhere here. Did you delete it, or is it hidden in a conversation? The github app notification on my phone says it's about staking tests but I can't see the full comment nor click the links

uhbif19:
@jy14898 I was asking about failing stacking tests in CI, but then find out PR on that and deleted my question. Thanks for your reaction!

src/Internal/Contract/QueryBackend.purs
uhbif19: But PR does not actually use it? Is there some plans to use it in the future?

jy14898:
It looks like logging suppression has broken at some point, I'll take a look into it and also run some full tests today

PR #1261, opened 2022-11-23
add mote to Scaffold.Main test

### LINKED ISSUE #1241: Use Mote in template test suite
https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/plutip-testing.md#testing-with-mote

Add this ^ link to the comments near the code.

pb99u069:
Closes #1241.

### Pre-review checklist


templates/ctl-scaffold/test/Main.purs
```diff
@@ -6,22 +6,32 @@ import Contract.Prelude
 
 import Contract.Config as Contract.Config
 import Contract.Monad as Contract.Monad
+import Contract.Test.Plutip (PlutipTest, testPlutipContracts)
 import Contract.Test.Plutip as Contract.Test.Plutip
-import Contract.Wallet as Contract.Wallet
+import Ctl.Internal.Test.TestPlanM (TestPlanM, interpretWithConfig)
```
jy14898, line 11: I thought I'd take a quick look before it's ready for review - we don't export interpretWithConfig, so we really shouldn't be using it in the template. We could export it, but ideally the way CTL internally tests and the way users test should both be able to change freely - if we export it, we have promises to keep.

I think we should copy https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/src/Internal/Test/TestPlanM.purs as `ctl-scaffold/test/Plan.purs` and drop the `TestPlanM` alias in the copy. I'd be tempted to not use our custom version of `consoleReporter` either, and use the built in one from `purescript-spec`.

pb99u069: Thanks for the tip. as of now, we only need `interpretWithConfig`, but I left `interpret` and `interpretWithTimeout` in Plan.purs in case users want to use it. or should I remove it?

klntsky: No, we don't want to copy it. We want to make it a part of our API. Move `interpretWithConfig` and the like to `Contract.Test.Mote`. Our `consoleReporter` should go to `Contract.Test.Mote.ConsoleReporter` and should be used by default (it is better).

`interpretWithTimeout`  is not needed at all. I think we should drop it entirely, since it just sets a single config field.

pb99u069:
I noticed that `make format` doesn't work in the template because we don't have `fd` in the shell. added it, hope in the right place

templates/ctl-scaffold/test/Main.purs
```diff
@@ -6,22 +6,33 @@ import Contract.Prelude
...
 import Contract.Test.Plutip as Contract.Test.Plutip
-import Contract.Wallet as Contract.Wallet
 import Data.BigInt as BigInt
 import Data.UInt as UInt
+import Effect.Aff (Milliseconds(Milliseconds))
+import Mote (test)
 import Scaffold as Scaffold
+import Test.Spec.Runner (defaultConfig)
 
+-- Run with `spago test --main Test.Scaffold.Main`
```
klntsky, line 19: `npm run test`

klntsky (changes requested):

templates/ctl-scaffold/test/Main.purs
pb99u069: I also changed this in `Test.Ctl.Plutip`, from where I copied the command.

flake.lock
```diff
@@ -6047,11 +6047,11 @@
...
       "flake": false,
       "locked": {
-        "lastModified": 1668598393,
-        "narHash": "sha256-n1oob0LjJdfkkfWTa5GI6aAKG0xYRIAD63KIPr1E9nY=",
+        "lastModified": 1668678914,
+        "narHash": "sha256-XsbAFyUPmevGuoShEFlOVHt/7fFIpyCQuhulIrNzv80=",
         "owner": "CardanoSolutions",
         "repo": "kupo",
-        "rev": "c0312e5850a528e6ecebb350b3c54e7ab92c55ef",
+        "rev": "c9bc18d99f9e8af1840a265907db82b180d5a4d8",
```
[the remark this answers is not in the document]
pb99u069: wrong? I saw in your [PR](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1270) that you changed the hash in both packages.dhall and flake.nix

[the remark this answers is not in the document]
pb99u069: ahh wait, that's in flake.lock. no idea where this change is coming from, I'll try to find out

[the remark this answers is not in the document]
klntsky: Just kill it I think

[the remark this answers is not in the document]
pb99u069: I deleted and recreated the lock-file when I got the `could not find module` error (and before I realized that the solution is to update the commit hash in the templates packages.dhall). The new [commit](https://github.com/CardanoSolutions/kupo/commit/c9bc18d99f9e8af1840a265907db82b180d5a4d8) seems to just contain a minor fix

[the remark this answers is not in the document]
pb99u069: with kill it you mean reset it to the old rev?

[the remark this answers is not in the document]
klntsky: > with kill it you mean reset it to the old rev?

yes

flake.lock
```diff
@@ -2997,11 +2997,11 @@
     "flake-compat_4": {
       "flake": false,
       "locked": {
-        "lastModified": 1650374568,
-        "narHash": "sha256-Z+s0J8/r907g149rllvwhb4pKi8Wam5ij0st8PwAh+E=",
+        "lastModified": 1668681692,
+        "narHash": "sha256-Ht91NGdewz8IQLtWZ9LCeNXMSXHUss+9COoqu6JLmXU=",
```
klntsky, line 3001: It's still modified. Just copy the contents from the target branch.

pb99u069: sorry! should have had a better look at the other changes, thought they were ok. It's reset to the old version now

uhbif19 (approved):

klntsky (changes requested):
Please update the changelog, since there are additions to the public API

klntsky (approved):

klntsky:
@pb99u069 please fix the conflicts

PR #1264, opened 2022-11-24
CIP-30 optional args support

### LINKED ISSUE #1132: Support optional parameters (CIP-30)
These [CIP-30](https://cips.cardano.org/cips/cip30/) methods have optional parameters the users might want to utilize, but CTL does not allow specifying them:

- `api.getUtxos(amount: cbor\ = undefined, paginate: Paginate = undefined): Promise\`
- `api.getCollateral(params: { amount: cbor\ }): Promise\`
- `api.getUsedAddresses(paginate: Paginate = undefined): Promise\`
- `api.signTx(tx: cbor\, partialSign: bool = false): Promise\>`

We should add support for pagination (create a new type), amounts and `partialSign = false`. Public interface functions should be extended with extra parameters (wrapped in `Maybe`)

uhbif19:
Closes #1132.

* `partialSign` does nothing, cuz I am not sure what it is supposed to do in mock.
* `getUtxos` in our code filters out collateral outputs - I am not sure if that is correct, cuz CIP-30 seems to say nothing about this.

### Pre-review checklist


src/Contract/Address.purs
```diff
@@ -180,9 +189,12 @@ getWalletAddressesWithNetworkTag = do
 -- | Throws on `Promise` rejection by wallet, returns `Nothing` if no collateral
 -- | is available.
 getWalletCollateral
-  :: forall (r :: Row Type). Contract r (Maybe (Array TransactionUnspentOutput))
-getWalletCollateral = do
-  mtxUnspentOutput <- wrapContract QueryM.getWalletCollateral
+  :: forall (r :: Row Type)
+   . Maybe Value
+  -> Contract r (Maybe (Array TransactionUnspentOutput))
+getWalletCollateral amount = do
```
klntsky, line 195: We should **require** an amount, according to CIP-30 (please remove `Maybe`). The amount value should be provided as a `Coin`.

> The amount parameter is required, specified as a string (BigNumber) or a number, 

We should use a string in FFI

> and the maximum allowed value must be agreed to be something like 5 ADA

This is of course a bad wording in the standard, but we should also enforce the 5 ADA maximum.

src/Contract/Address.purs
```diff
@@ -141,8 +145,13 @@ getWalletAddress = head <$> getWalletAddresses
 -- | Get all the `Address`es of the browser wallet.
 getWalletAddresses
   :: forall (r :: Row Type). Contract r (Array Address)
-getWalletAddresses = do
-  addresses <- wrapContract QueryM.getWalletAddresses
+getWalletAddresses = getWalletAddressesPaginated Nothing
+
+-- | Get all the `Address`es of the browser wallet with pagination.
+getWalletAddressesPaginated
+  :: Maybe Paginate -> forall (r :: Row Type). Contract r (Array Address)
```
klntsky, line 152: The paginated version should require a `Paginate` argument (remove `Maybe`)

examples/Wallet.purs
```diff
@@ -12,11 +12,11 @@ contract = do
   log "Address:"
   log <<< show =<< getWalletAddresses
   log "Collateral:"
-  log <<< show =<< getWalletCollateral
+  log <<< show =<< getWalletCollateral Nothing
   log "Balance:"
   log <<< show =<< getWalletBalance
```
klntsky, line 17: Please add paginated calls here too

klntsky (changes requested):

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -206,3 +242,30 @@ mkCip30Mock pKey mSKey = do
...
+-- Helpers
+
+foreign import raisePaginateError :: forall a. Int -> Effect a
+
+foreign import raiseAPIError :: forall a. Effect a
+
+paginateArray :: forall a. Maybe Paginate -> Array a -> Effect (Array a)
+paginateArray (Just { page, limit }) xs = process $ pages !! page
+  where
+  chunkBy :: forall a. Int -> Array a -> Array (Array a)
```
klntsky, line 255: This is O(n^2). Since there may be thousands of utxos, it is a problem. Do not use `drop`&`take` (it's `O(n+m)`, where `n` and `m` are lengths of the resulting arrays), recursively iterate over `Int` indices instead and use `Array.slice`. That way each element of the original array will "participate" only in 1 `slice` call.

klntsky (changes requested):

src/Contract/Address.purs
uhbif19: getWalletAddressesPaginated - is more general version used in `getWalletAddresses`.

klntsky: @uhbif19 yep, please do

src/Contract/Address.purs
uhbif19: But this is not CIP-30 interface, but our internal interface, there Maybe is useful.

klntsky: OK I agree. Please make sure the semantics for `Nothing` are well described in the doc-comment

uhbif19: You were right, we can just use `mempty` as Nothing here.

klntsky (changes requested):

src/Internal/Types/BigNum.purs
```diff
@@ -116,6 +120,17 @@ fromString = _fromString maybeFfiHelper
...
 
+foreign import data StringOrNumber :: Type
+
+foreign import _fromStringOrNumber
+  :: MaybeFfiHelper -> StringOrNumber -> Maybe BigNum
+
+fromStringOrNumber :: StringOrNumber -> Maybe BigNum
+fromStringOrNumber = _fromStringOrNumber maybeFfiHelper
+
+stringToStringOrNumber :: String -> StringOrNumber
```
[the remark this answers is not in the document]
uhbif19: The point is that CIP-30 standard uses String or Number to represent BigNumber amount argument.

After writing that code I found out that we use untagged union lib. I think it may be suited better than writing separate type?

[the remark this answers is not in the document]
uhbif19: Also this is strange to me, why CIP-30 uses BigNumber instead of BigInteger here.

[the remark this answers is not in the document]
klntsky: It can't use BigNumber. It's CSL internal type. Maybe a typo in the spec.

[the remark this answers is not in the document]
klntsky: I see, they don't accept a BigNumber, they mean that the representation of the number in the string is the same as BigNumber string representation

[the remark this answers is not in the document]
klntsky: Please remove this type and just use String in the CIP-30 FFI code. The interface should accept a `Coin`.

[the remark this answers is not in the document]
uhbif19: Sorry, I mixed up, I meant BigNum, not BigNumber.

klntsky (changes requested):

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 - Running plutip servers attaches on SIGINT handlers and therefore node will not exit by default. ([#1231](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1231)).
 - `TestPlanM`, `interpret` and `interpretWithConfig` are now public in `Contract.Test.Mote` and our custom `consoleReporter` in `Contract.Test.Mote.ConsoleReporter`. ([#1261](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1261)).
+- CIP-30 got optional args support on level of `Cip30Mock` and `Cip30Wallet` module. `getWalletAddresses` and `getValue` got `Maybe Paginate` argument, `getCollateral` got amount argument of type `Coin` and `getValue` got amount argument of type `Value`. ([#1261](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1261)
```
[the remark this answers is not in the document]
uhbif19: I do not understand you.

src/Internal/Undefinable.purs
```diff
@@ -0,0 +1,20 @@
+module Ctl.Internal.Undefinable
```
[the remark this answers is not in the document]
klntsky: > So we should support it somehow.

Yes, by using different type annotations. For example, the UndefinedOr type:  https://plutonomicon.github.io/cardano-transaction-lib/Untagged.Union.html#t:UndefinedOr

CHANGELOG.md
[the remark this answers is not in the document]
klntsky: It's getWalletAddressesPaginated

src/Internal/Types/BigNum.purs
```diff
@@ -35,6 +35,7 @@ import Data.UInt (UInt)
 import Data.UInt (fromInt, fromString, toString) as UInt
 import Partial.Unsafe (unsafePartial)
 import Type.Row (type (+))
+import Unsafe.Coerce (unsafeCoerce)
```
jy14898, line 38: Unused import (and dependency as far as I can tell)

jy14898 (changes requested):

src/Internal/Plutus/Conversion/Value.purs
```diff
@@ -61,6 +62,9 @@ fromPlutusValue plutusValue =
     $ nonAdaTokenMap <#> \(cs /\ tokens) ->
         Plutus.getCurrencySymbol cs /\ Map.fromFoldable (unwrap tokens)
 
+fromPlutusCoin :: Plutus.Coin -> Types.Coin
+fromPlutusCoin = Types.Coin <<< Plutus.getLovelace
+
```
klntsky, line 67: We need a conversion function in the opposite direction as well (`toPlutusCoin`)

Bradley-Heather (changes requested):

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -210,3 +274,34 @@ mkCip30Mock pKey mSKey = do
...
+
+-- Helpers
+
+foreign import raisePaginateError :: forall a. Int -> Effect a
+
+foreign import _catchPaginateError
+  :: forall a. MaybeFfiHelper -> Effect a -> Effect (Maybe a)
+
+catchPaginateError :: forall a. Effect a -> Effect (Maybe a)
+catchPaginateError = _catchPaginateError maybeFfiHelper
```
klntsky, line 286: It is not used anywhere, right? If so, please remove.

klntsky (changes requested):

src/Contract/Address.purs
```diff
@@ -141,8 +145,17 @@ getWalletAddress = head <$> getWalletAddresses
 -- | Get all the `Address`es of the browser wallet.
 getWalletAddresses
   :: forall (r :: Row Type). Contract r (Array Address)
-getWalletAddresses = do
-  addresses <- wrapContract QueryM.getWalletAddresses
+getWalletAddresses = getWalletAddressesGeneral Nothing
+
+-- | Get all the `Address`es of the browser wallet with pagination.
+getWalletAddressesPaginated
+  :: Paginate -> forall (r :: Row Type). Contract r (Array Address)
```
jy14898, line 152: Typically we lift forall's as high as possible

```suggestion
  :: forall (r :: Row Type). Paginate -> Contract r (Array Address)
```

src/Internal/QueryM/Utxos.purs
```diff
@@ -117,15 +118,22 @@ getWalletBalance = do
...
+getWalletUtxos :: Maybe Value -> Maybe Paginate -> QueryM (Maybe UtxoMap)
+getWalletUtxos value paginate = do
   asks (_.runtime >>> _.wallet) >>= map join <<< traverse case _ of
+<<<<<<< HEAD
+    Nami wallet -> common wallet
+    Gero wallet -> common wallet
+    Flint wallet -> common wallet
+    Eternl wallet -> common wallet
+    Lode wallet -> common wallet
+=======
```
jy14898, line 136: Merge conflict

jy14898 (changes requested):

src/Internal/Wallet/Cip30Mock.purs
uhbif19: It is used in test: https://github.com/Plutonomicon/cardano-transaction-lib/pull/1264/files#diff-826108a8e51ac7cdb18bef882b2860b322b5bde19916c95705dd5ef5ef54e8a0R39

uhbif19: For use in our PS API this is catched in Cip30.js

src/Contract/Address.purs
uhbif19: Thanks!

src/Internal/Wallet/Cip30Mock.purs
klntsky: Can we move it to the test module then?

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -109,20 +133,31 @@ withCip30Mock (KeyWallet keyWallet) mock contract = do
...
 
+type CollateralParams = { amount :: String }
+
 type Cip30Mock =
   { getNetworkId :: Effect (Promise Int)
-  , getUtxos :: Effect (Promise (Array String))
-  , getCollateral :: Effect (Promise (Array String))
+  , getUtxos ::
+      Fn2 (UndefinedOr String) (UndefinedOr Paginate)
+        (Promise (Nullable.Nullable (Array String)))
```
klntsky, line 142: Please remove Nullable dependency. untagged-union provides a `HasRuntimeType Null` instance, which makes it possible to use `Null |+| a` (very similar to  `UndefinedOr` from `untagged-union` that we already use).

klntsky (changes requested):

src/Internal/Wallet/Cip30Mock.purs
uhbif19: Then FFI will be here and we must export it. Which looks even stranger than current state to me.

src/Internal/Wallet/Cip30Mock.js
```diff
@@ -1,5 +1,29 @@
...
+class APIError {}
+
+exports.raiseAPIError = function () {
+  throw new APIError();
+};
+
+exports._catchPaginateError = maybeffi => action => () => {
+  try {
+    return maybeffi.just(action());
+  } catch (PaginateError) {
```
jy14898, line 22: I don't think this is standard syntax?

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -210,3 +274,48 @@ mkCip30Mock pKey mSKey = do
 
 -- returns an action that removes the mock.
 foreign import injectCip30Mock :: String -> Cip30Mock -> Effect (Effect Unit)
+
+-- Helpers
+
+foreign import raisePaginateError :: forall a. Int -> Effect a
```
jy14898, line 280: kind annotations missing for foralls

src/Internal/Wallet/Cip30Mock.js
```diff
@@ -1,5 +1,29 @@
...
+  constructor(maxSize) {
+    this.maxSize = maxSize;
+  }
+}
+
+exports.raisePaginateError = maxSize => () => {
+  throw new PaginateError(maxSize);
+};
+
+class APIError {}
```
jy14898, line 13: Doesn't follow CIP30, missing info and code fields

jy14898 (changes requested):

src/Internal/Wallet/Cip30Mock.purs
uhbif19: Why there should be kind annotation? Looks like regular Type, no?

klntsky: It's unfortunately something from MLabs code style we inherited. It should be `forall (a :: Type). `

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -210,3 +274,48 @@ mkCip30Mock pKey mSKey = do
...
+  where
+  utxoAmount (TransactionUnspentOutput { output }) = outputAmount output
+  outputAmount (TransactionOutput { amount }) = amount
+  sumAmount = fold $ map utxoAmount values
+
+-- Support NullOr
+
+foreign import data Null :: Type
+
+foreign import null :: Null
```
klntsky, line 313: Use null from `Literals.Null` - it has a RuntimeType instance that makes it possible to use it with `OneOf`.

klntsky (changes requested):

src/Internal/Wallet/Cip30Mock.purs
uhbif19: You mean this one? Looks like it implemented just the same.

https://github.com/rowtype-yoga/purescript-literals/blob/v1.0.2/src/Literals/Null.purs#L13-L13

klntsky: Except of the instance that is defined in untagged-union source

uhbif19: Checked - still cannot use `oneOf`. I do not think that instance I wanted to use uses `RuntimeType` somehow.

klntsky: You forgot the `Castable a (OneOf Null a)` :

```

maybeToNullOr :: forall a. Castable a (OneOf Null a) => Maybe a -> NullOr a
maybeToNullOr (Just x) = asOneOf x
maybeToNullOr Nothing = asOneOf null
```

src/Internal/Wallet/Cip30.js
```diff
@@ -2,27 +2,32 @@
...
 
 exports._getChangeAddress = conn => () => conn.getChangeAddress();
 
 exports._getRewardAddresses = conn => () => conn.getRewardAddresses();
 
-exports._signTx = txHex => conn => () =>
+exports._signTx = txHex => _partialSign => conn => () =>
   conn.signTx(txHex, true).catch(e => {
     throw JSON.stringify(e);
   });
```
jy14898, line 33: You asked about partialSign, and how the mock can implement it. From what I understand, `partialSign` dictates (when set to true) that it's ok for there to be [required signers](https://input-output-hk.github.io/cardano-js-sdk/classes/_cardano_sdk_core.CSL.TransactionBody.html#required_signers) missing from the returned vkeys in the [witness set](https://input-output-hk.github.io/cardano-js-sdk/classes/_cardano_sdk_core.CSL.TransactionWitnessSet.html). When false, there must be no missing signers. The behaviour we already implement is if `partialSign = true` for the key wallet.

jy14898 (changes requested):

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -155,49 +211,69 @@ mkCip30Mock pKey mSKey = do
...
-        -- Convert to CSL representation and serialize
-        cslUtxos <- traverse (liftEffect <<< convertTransactionUnspentOutput)
-          $ Map.toUnfoldable nonCollateralUtxos <#> \(input /\ output) ->
-              TransactionUnspentOutput { input, output }
-        pure $ (byteArrayToHex <<< unwrap <<< toBytes) <$> cslUtxos
-    , getCollateral: fromAff do
+        let
+          xUtxos = Map.toUnfoldable nonCollateralUtxos <#> \(input /\ output) ->
+            TransactionUnspentOutput { input, output }
+        let
```
rynoV, line 217: Extra `let`s

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -155,49 +211,69 @@ mkCip30Mock pKey mSKey = do
...
+          amountValue = DSV.convertValue =<< fromBytes =<< hexToCborBytes =<<
+            (uorToMaybe amount)
+        if (not hasEnoughAmount amountValue xUtxos) then pure $ maybeToNullOr
```
rynoV, line 220: Strange style, works because of [this](https://github.com/purescript/purescript-prelude/blob/68f8012bc2309d9bf5832cdf7316ad052d586905/src/Data/HeytingAlgebra.purs#L71), didn't know about this. I'm guessing this generates more complicated JS code than `not $ hasEnoughAmount amountValue xUtxos`, as the instances have to be resolved recursively for the function instance.

rynoV: Also I don't think the parens are necessary

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -155,49 +211,69 @@ mkCip30Mock pKey mSKey = do
...
-        cslUnspentOutput <- liftEffect $ traverse
-          convertTransactionUnspentOutput
-          collateralUtxos
-        pure $ (byteArrayToHex <<< unwrap <<< toBytes) <$>
-          cslUnspentOutput
+        amountValue <- liftEffect $ convertAmount amount
+        if gt amountValue maxCollateralAmount then liftEffect $
+          raiseInvalidRequestError "Amount value is bigger than allowed 5 ADA"
+        else pure unit
+        if (not $ hasEnoughAmount (Just amountValue) collateralUtxos) then
```
rynoV, line 240: Don't think parens are necessary

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -155,49 +211,69 @@ mkCip30Mock pKey mSKey = do
...
         collateralUtxos <- getCollateralUtxos utxos
-        cslUnspentOutput <- liftEffect $ traverse
-          convertTransactionUnspentOutput
-          collateralUtxos
-        pure $ (byteArrayToHex <<< unwrap <<< toBytes) <$>
-          cslUnspentOutput
+        amountValue <- liftEffect $ convertAmount amount
+        if gt amountValue maxCollateralAmount then liftEffect $
+          raiseInvalidRequestError "Amount value is bigger than allowed 5 ADA"
+        else pure unit
```
rynoV, line 239: `when` can be used instead

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -155,49 +211,69 @@ mkCip30Mock pKey mSKey = do
...
+        if (not $ hasEnoughAmount (Just amountValue) collateralUtxos) then
+          liftEffect $ raiseInvalidRequestError
+            "Cannot find collateral to match required amount"
+        else do
+          cslUnspentOutput <- liftEffect $ traverse
+            convertTransactionUnspentOutput
+            collateralUtxos
+          pure $ (cborBytesToHex <<< toBytes) <$> cslUnspentOutput
```
rynoV, line 247: `when` can be used instead of `if then else` to reduce indentation

src/Internal/Wallet/Cip30Mock.purs
```diff
@@ -155,49 +211,69 @@ mkCip30Mock pKey mSKey = do
...
     , getUnusedAddresses: fromAff $ pure []
     , getChangeAddress: fromAff do
         (unwrap keyWallet).address config.networkId <#>
           byteArrayToHex <<< unwrap <<< toBytes
     , getRewardAddresses: fromAff do
         (unwrap keyWallet).address config.networkId <#> \address ->
-          [ byteArrayToHex $ unwrap $ toBytes address ]
-    , signTx: \str -> unsafePerformEffect $ fromAff do
-        txBytes <- liftMaybe (error "Unable to convert CBOR") $ hexToByteArray
+          [ (cborBytesToHex <<< toBytes) address ]
```
rynoV, line 266: To follow style guide

```suggestion
          [ cborBytesToHex $ toBytes address ]
```

rynoV (changes requested):

src/Internal/Wallet/Cip30.purs
```diff
@@ -238,22 +274,35 @@ foreign import _getNetworkId
...
+  amountBigNumStr <- fromJustEff "Incorrect Coin value" $
+    coinToBigNumberStr amount
+  _getCollateral maybeFfiHelper conn amountBigNumStr `catchError`
     \_ -> throwError $ error "Wallet doesn't implement `getCollateral`."
+  where
+  coinToBigNumberStr coin =
+    BigNum.toString <$>
+      ( BigNum.fromBigInt $
+          getLovelace coin
+      )
```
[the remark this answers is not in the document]
uhbif19: > The amount parameter is required, specified as a string (BigNumber) or a number

says CIP-30,  it looks like BigNum to me

src/Internal/Wallet/Cip30.js
```diff
@@ -2,27 +2,32 @@
 
 exports._getNetworkId = conn => () => conn.getNetworkId();
 
-exports._getUtxos = maybe => conn => () =>
-  conn.getUtxos().then(res => (res === null ? maybe.nothing : maybe.just(res)));
+exports._getUtxos = maybe => conn => amount => paginate => () =>
+  conn
+    .getUtxos(amount, paginate)
+    .then(res => (res === null ? maybe.nothing : maybe.just(res)))
+    .catch(_ => maybe.nothing);
```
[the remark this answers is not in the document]
uhbif19: I do not understand what you mean. catches below use throw, which is not applicable here, as far as I understand.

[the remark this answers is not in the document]
rynoV: I'm referring to when we `catch` the errors and return `Nothing` without logging what was caught. If we could use our existing logging functionality to log the errors (or not, depending on the Contract config) instead of silencing them completely, I think that would be better for debugging

src/Internal/Wallet/Cip30.purs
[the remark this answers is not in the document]
rynoV: Ah I see. The signature they give uses the type `cbor<Coin>`, which based on [here](https://cips.cardano.org/cips/cip30/#cbort) I would've thought was an int as `coin` is defined [here](https://github.com/input-output-hk/cardano-ledger/blob/0738804155245062f05e2f355fadd1d16f04cd56/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L248), but that `coin` isn't capitalized, so afaict `Coin` is undefined in the CIP30 spec. The part you quoted seems to be the closest to a definition we have, so I think this is fine

src/Internal/QueryM/Utxos.purs
```diff
@@ -117,41 +118,43 @@ getWalletBalance = do
...
+    Flint wallet -> common wallet
+    Eternl wallet -> common wallet
+    Lode wallet -> common wallet
+    NuFi wallet -> common wallet
     KeyWallet _ -> do
-      mbAddress <- getWalletAddresses <#> head
+      mbAddress <- getWalletAddresses paginate <#> head
       map join $ for mbAddress utxosAt
   where
+  common wallet = liftAff $
```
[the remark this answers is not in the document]
uhbif19: This leads to dependency cycle. Also there are also functions without signatures in lets in this module.

uhbif19:
@jy14898 Please check if I modified `template/ctl-scaffold/packages.dhall` correctly.

src/Internal/Wallet/Cip30Mock.purs
uhbif19: How so?  If have two branches, while when has only one.

jy14898: Raising an error will cause the second clause to never be evaluated

jy14898:
@uhbif19 Looks good, we'll also need to push the templates commit of CTL at some point but we can leave that for later

examples/Wallet.purs
```diff
@@ -12,11 +12,13 @@ contract = do
   log "Address:"
   log <<< show =<< getWalletAddresses
   log "Collateral:"
-  log <<< show =<< getWalletCollateral
+  log <<< show =<< getWalletCollateral mempty
```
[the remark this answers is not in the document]
uhbif19: Now correct?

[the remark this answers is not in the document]
jy14898: Well here and everywhere else we use `getWalletCollateral`

src/Internal/Wallet/Cip30.purs
```diff
@@ -98,7 +119,7 @@ mkCip30WalletAff
 mkCip30WalletAff walletName enableWallet = do
   wallet <- toAffE enableWallet
   -- Ensure the Nami wallet has collateral set up
-  whenM (isNothing <$> getCollateral wallet) do
+  whenM (isNothing <$> getCollateral wallet mempty) do
```
jy14898, line 122: A reminder that we should replace these memptys with 5ada. You could use `Ctl.Internal.BalanceTx.Collateral.Select.minRequiredCollateral` (except in places with import cycles or examples)

jy14898 (changes requested):

src/Internal/Wallet/Cip30Mock.purs
uhbif19: As far as I understand this style (not writing else, ) is not enforced by our style guide and I am myself stand against that style, so I will not change that.

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 
+- CIP-30 got optional args support on level of `Cip30Mock` and `Cip30Wallet` module API is changed to support them. `getWalletAddresses` got version `getWalletAddressesPaginated`. `getValue` got `Maybe Paginate` argument. `getCollateral` got amount argument of type `Coin` and `getValue` got amount argument of type `Value`. ([#1264](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1264)
```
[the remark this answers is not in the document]
uhbif19: > Why do we have one function which is suffixed with Paginated, while another accepts a Maybe?

That was suggestion of @klntsky  to have version without Maybe: https://github.com/Plutonomicon/cardano-transaction-lib/pull/1264#discussion_r1033553271

Not `Paginated` version is saved to simplify refactoring, as far as I remember.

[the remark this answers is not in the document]
uhbif19: `getValue` is probably my error. I think I meant `getWalletUtxos`.

src/Internal/Wallet/Cip30.js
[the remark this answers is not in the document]
klntsky: We don't care about errors at this stage

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -199,7 +199,8 @@ balanceTxWithConstraints unbalancedTx constraintsBuilder = do
   setTransactionCollateral :: Address -> Transaction -> BalanceTxM Transaction
   setTransactionCollateral changeAddr transaction = do
     collateral <-
-      liftEitherQueryM $ note CouldNotGetCollateral <$> getWalletCollateral
+      liftEitherQueryM $
+        note CouldNotGetCollateral <$> getWalletCollateral mempty
```
klntsky, line 203: mempty = 0? It's not what we want

klntsky (changes requested):

uhbif19:
Closed as stale.

PR #1285, opened 2022-12-02
Organize `FromBytes` and `ToBytes` (version 2)

### LINKED ISSUE #560: Improve code organization (`from_bytes` / `to_bytes`)

rynoV:
Closes #560. Replaces #1073 to build on top of the new `develop`.

### Pre-review checklist


src/Internal/Hashing.purs
```diff
@@ -53,13 +54,16 @@ foreign import sha3_256HashHex :: ByteArray -> String
 
 datumHash :: Datum -> Maybe DataHash
 datumHash =
-  map (wrap <<< hashPlutusData) <<< convertPlutusData <<< unwrap
+  map (wrap <<< unwrap <<< toBytes <<< hashPlutusData)
+    <<< convertPlutusData
+    <<<
+      unwrap
```
klntsky, line 60:
```suggestion
    <<< unwrap
```

src/Internal/QueryM/Pools.purs
```diff
@@ -97,8 +102,11 @@ getPubKeyHashDelegationsAndRewards pkh = do
   stringRep =
     ed25519KeyHashToBech32Unsafe "stake_vkh" $ unwrap $ unwrap pkh
 
+  -- byteHex :: String
+  -- byteHex = byteArrayToHex $ unwrap $ toBytes $ unwrap $ unwrap pkh
+
```
klntsky, line 107: Leftover?

test/Serialization/Hash.purs
```diff
@@ -67,14 +67,14 @@ suite = do
...
-    $ wrap
     $ hexToByteArrayUnsafe
         scriptHashHex
   let
     scrhB32 = scriptHashToBech32Unsafe "stake_vkh" scrh
     mScrhB32 = scriptHashToBech32 "stake_vkh" scrh
-    scrhBts = scriptHashToBytes scrh
-    scrhFromBytes = scriptHashFromBytes scrhBts
+    scrhBts = toBytes scrh
+    -- TODO: use fromBytes instead?
```
klntsky, line 76: Does it work? If so, yes.

klntsky (changes requested):

rynoV:
@klntsky changes since your review: I added `FromBytes` instances to make them consistent with the `toBytes` types: `Address`, `Ed25519KeyHash`, `ScriptHash`, `Redeemers`, `TransactionBody`, and `TransactionOutput`.

I also added `Mint` as a `toBytes` type to make it consistent with `FromBytes`. `PublicKey` may also be added to `SerializationData`, but for some reason CSL doesn't give it a `to_bytes` method (currently checking on Slack if I should make an issue with CSL).

klntsky:
@rynoV PublicKey does not have a cbor representation, because it is never a part of a Tx. We should not have toBytes/fromBytes for it.

rynoV:
@klntsky Does `VRFKeyHash` have its own "toBytes" function ([vrfKeyHashToBytes](https://github.com/Plutonomicon/cardano-transaction-lib//blob/2c5995581664b155d6d4a43e8e60089e391ebe96/src/Internal/Types/VRFKeyHash.purs#L39)) because of [this](https://github.com/Plutonomicon/cardano-transaction-lib//blob/2c5995581664b155d6d4a43e8e60089e391ebe96/src/Internal/Serialization/ToBytes.purs#L59) (to_bytes for this type doesn't give CborBytes)?

rynoV:
Does it make sense to provide `toBytes` for `COSEKey` as well (to replace [bytesFromCoseKey](https://github.com/Plutonomicon/cardano-transaction-lib//blob/2c5995581664b155d6d4a43e8e60089e391ebe96/src/Internal/Wallet/Cip30/SignData.js#L112)), even though it's not from CSL? It should work since the `to_bytes` for `COSEKey` just returns a byte array, and `CBORBytes` at runtime is a byte array.

rynoV:
> We should not have toBytes/fromBytes for it.

Realized the `FromBytes` instance was being used in a test here [here](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1285/commits/6588226bbff07e9041161b230412afa0ef2c3ecc#diff-f81e2b5788f7e1b6a1fd2a0282c0c588243b3deca2df72cb161fe88e0ecb026aR131)

src/Internal/QueryM/Pools.purs
pb99u069: This one (and also line 83/84) is a bit of a mystery to me. The commented out versions produce a `No type class instance` error while the two step solutions (calculating first ScriptHash/Ed2551KeyHash and then serialize) work fine. afaics there should be no difference between the two..

rynoV:
I was trying to see if we could get rid of the `to/from_bytes` usages in `Serialization/Address.js`, but unfortunately I don't think there's a way without causing cyclic dependency issues, using orphan instances, or putting everything in one big file. But I have changed it so that `Address.purs`'s to/fromBytes functions are only used within the module.

There's still the issue of the `base/enterprise/pointer/rewardAddressBytes` functions, but I don't think they can be replaced by `toBytes` without checking the types at runtime, or patching CSL to add `to_bytes` methods for those classes. **However if we were to bring back the `ToBytes` class we could make this work**, but I'm also not entirely sure what the reasons against using a class were [here](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1073#discussion_r993598706). For the `FromBytes` case, we could replace the individual functions with instances of `FromBytes`, but I think it's better to keep it consistent.

src/Internal/Deserialization/FromBytes.purs
```diff
@@ -1,155 +1,135 @@
...
 import Type.Row (type (+))
 
 -- | Calls `from_bytes` method for the appropriate type
 class FromBytes a where
   fromBytes' :: forall (r :: Row Type). ByteArray -> E (FromBytesError + r) a
 
-instance FromBytes DataHash where
-  fromBytes' = _fromBytesDataHash eh
+instance FromBytes Address where
+  fromBytes' = _fromBytes "Address" fromBytesErrorHelper
```
Luis-omega, line 50: Why do we need to explicitly pass `fromBytesErrorHelper` to all these calls to `_fromBytes`?

src/Internal/QueryM/Pools.purs
rynoV: It's a mystery to me as well, seems like a purescript oddity. I'm going to ask about it in Slack, but deleting these comments for now as this is working

test/Serialization/Hash.purs
rynoV: Currently my understanding is that `scriptHashFromBytes`, `ed25519KeyHashFromBytes`, `vrfKeyHashFromBytes`, and `vrfKeyHashToBytes` are all meant to be separate functions because `to/from_bytes` for hashes doesn't use CBOR, but just returns raw bytes instead, so for now I'm just leaving these functions alone.

src/Internal/Deserialization/FromBytes.purs
rynoV: That's a good point, we can avoid repetition, I'll make the change

src/Internal/Serialization/ToBytes.purs
```diff
@@ -17,32 +28,47 @@ import Ctl.Internal.Serialization.Types
...
+  |+| Transaction
+  |+| TransactionBody
+  |+| TransactionHash
+  |+| TransactionOutput
+  |+| TransactionUnspentOutput
+  |+| TransactionWitnessSet
+  |+| Value
+  |+| VRFKeyHash
+
+-- and more as needed
```
Bradley-Heather, line 63:
```suggestion
-- add more as needed
```

src/Internal/Serialization/ToBytes.purs
```diff
@@ -17,32 +28,47 @@ import Ctl.Internal.Serialization.Types
...
-  -> ByteArray
+foreign import _toBytes
+  :: forall a. a -> ByteArray
+
+toBytes
+  :: forall a
+   . Castable a SerializationData
+  => a
+  -> CborBytes
+toBytes = CborBytes <<< _toBytes
```
Bradley-Heather, line 74: We're wrapping and then immediately unwrapping at multiple call sites so would something like below not make more sense? 
```suggestion
toBytes
  :: forall a
   . Castable a SerializationData
  => a
  -> CborBytes
toBytes = CborBytes <<< _toBytes

toRawBytes
  :: forall a
   . Castable a SerializationData
  => a
  -> ByteArray
toRawBytes = _toBytes
```

Bradley-Heather, line 74: You're wrapping then unwrapping immediately at multiple call sites so why not something along the lines of? 

```suggestion
toBytes
  :: forall a
   . Castable a SerializationData
  => a
  -> CborBytes
toBytes = CborBytes <<< _toBytes

toRawBytes
  :: forall a
   . Castable a SerializationData
  => a
  -> ByteArray
toRawBytes = _toBytes
```

src/Internal/QueryM/Pools.purs
```diff
@@ -79,10 +80,11 @@ getValidatorHashDelegationsAndRewards skh = do
...
   stringRep = scriptHashToBech32Unsafe "script" $ unwrap skh
 
+  sh :: ScriptHash
+  sh = unwrap skh
+
   byteHex :: String
-  byteHex =
-    byteArrayToHex <<< unwrap <<< scriptHashToBytes <<< unwrap $
-      skh
+  byteHex = byteArrayToHex $ unwrap $ toBytes sh
```
klntsky, line 87: Doesn't it work without specifying a type for a new variable?

src/Internal/QueryM/Pools.purs
```diff
@@ -97,8 +99,8 @@ getPubKeyHashDelegationsAndRewards pkh = do
...
 
+  ed :: Ed25519KeyHash
+  ed = unwrap $ unwrap pkh
+
   byteHex :: String
-  byteHex =
-    byteArrayToHex <<< unwrap <<< ed25519KeyHashToBytes <<< unwrap $
-      unwrap
-        pkh
+  byteHex = byteArrayToHex $ unwrap $ toBytes ed
```
klntsky, line 106: Same as above

src/Internal/Serialization.purs
```diff
@@ -554,8 +545,13 @@ convertProposedProtocolParameterUpdates
 convertProposedProtocolParameterUpdates ppus =
   newProposedProtocolParameterUpdates containerHelper =<<
     for (Map.toUnfoldable $ unwrap ppus) \(genesisHash /\ ppu) -> do
-      Tuple <$> newGenesisHash (unwrap genesisHash) <*>
-        convertProtocolParamUpdate ppu
+      Tuple
+        <$>
+          ( fromJustEff "Failed to convert genesis hash" $ fromBytes
+              (wrap $ unwrap genesisHash)
```
klntsky, line 551: Please use `$` where possible

```suggestion
          ( fromJustEff "Failed to convert genesis hash" $ fromBytes
              $ wrap $ unwrap genesisHash
```

klntsky (changes requested):

src/Internal/QueryM/Pools.purs
rynoV: No, but I'm changing it to use a type annotation instead of a variable. Seems like some oddity with the type checker, I'll link a discussion in slack

src/Internal/Deserialization/FromBytes.purs
```diff
@@ -1,155 +1,152 @@
...
+fromBytesEffect :: forall (a :: Type). FromBytes a => CborBytes -> Effect a
+fromBytesEffect bytes =
+  case fromBytes bytes of
+    Nothing -> throw "from_bytes() call failed"
+    Just a -> pure a
 
-foreign import _fromBytesTransactionWitnessSet
-  :: forall (r :: Row Type)
-   . ErrorFfiHelper r
+fromBytes''
```
klntsky, line 138: Let's rename it to `fromBytesImpl`

src/Internal/Deserialization/FromBytes.purs
```diff
@@ -1,155 +1,152 @@
...
-foreign import _fromBytesEd25519Signature
-  :: forall (r :: Row Type)
-   . ErrorFfiHelper r
+foreign import _fromBytes
+  :: forall (r :: Row Type) (a :: Type)
+   . String
+  -> ErrorFfiHelper r
   -> ByteArray
-  -> E r Ed25519Signature
+  -> E r a
```
klntsky, line 152: Order of arguments can be changed, to avoid `flip` above.

src/Internal/Serialization/ToBytes.purs
```diff
@@ -17,32 +28,47 @@ import Ctl.Internal.Serialization.Types
...
   , TransactionWitnessSet
-  , VRFKeyHash
   , Value
   )
 import Ctl.Internal.Types.ByteArray (ByteArray)
+import Ctl.Internal.Types.CborBytes (CborBytes(CborBytes))
+import Untagged.Castable (class Castable)
 import Untagged.Union (type (|+|))
 
+type SerializationData = Address
```
klntsky, line 38: `SerializableData` is probably a better name

klntsky (changes requested):
Thank you and sorry for not reviewing it sooner. I just left a few notes on naming.

klntsky (approved):
All good! Thanks!

klntsky (changes requested):
Except that the tests are failing in CI

rynoV:
@klntsky My bad forgot to reorder the params on the JS side as well

klntsky (approved):
Can be merged as long as CI passes

PR #1305, opened 2022-12-08
Add `submitTxFromConstraints` instead of `Helpers.buildBalanceSignAndSubmitTx`

### LINKED ISSUE #821: Add `submitTxConstraintsWith`
Quoting @ngua 

> We probably should add something like `buildBalanceSignAndSubmit` (we even use that in `Examples.AlwaysSucceeds`) but with a better name. We can use `submitTxConstraintsWith` probably

uhbif19:
Closes #821.

### Pre-review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 
 ### Added
 
+* Add `submitTxConstraintsWith` and `submitTxConstraintsWithReturningFee` in `Contract.Transaction`. `submitTxConstraintsWith` build a transaction that satisfies the constraints, then submit it to the network. It is analog for same function in Plutus and replaces `Helpers.buildBalanceSignAndSubmitTx`.
+
 ### Changed
 
 - Running plutip servers attaches on SIGINT handlers and therefore node will not exit by default. ([#1231](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1231)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 
 ### Added
 
-* Add `submitTxConstraintsWith` and `submitTxConstraintsWithReturningFee` in `Contract.Transaction`. `submitTxConstraintsWith` build a transaction that satisfies the constraints, then submit it to the network. It is analog for same function in Plutus and replaces `Helpers.buildBalanceSignAndSubmitTx`.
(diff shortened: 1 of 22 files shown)


examples/SendsToken.purs
```diff
@@ -12,14 +12,17 @@ import Contract.Log (logInfo')
 import Contract.Monad (Contract, launchAff_, liftedM, runContract)
 import Contract.ScriptLookups as Lookups
 import Contract.Scripts (MintingPolicy)
-import Contract.Transaction (TransactionHash, awaitTxConfirmed)
+import Contract.Transaction
+  ( TransactionHash
+  , awaitTxConfirmed
+  , submitTxConstraintsWith
```
klntsky, line 18: Let's rename it to `submitTxFromConstraints`, because technically constraints are not submitted anywhere.

klntsky (changes requested):

examples/SendsToken.purs
klntsky: Propose a better name if you wish, because this is also not ideal

uhbif19: @klntsky I do not even understand what `submitTxConstraintsWith` supposed to mean (with what? with Lookups?), but this is name taken from Plutus.

Bradley-Heather: I think `submitTxFromConstraints` is fine. It basically does what it says. Given `Constraints` it will build balance sign and eventually submit a Tx

klntsky: > this is name taken from Plutus.

Wow. Good to know we're not the only ones bad at naming.

Anyway, let's rename and put a new entry in `doc/plutus-comparison.md`

CHANGELOG.md
```diff
@@ -36,6 +36,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
+* Add `submitTxFromConstraints` and `submitTxFromConstraintsReturningFee` in `Contract.Transaction`. `submitTxFromConstraints` build a transaction that satisfies the constraints, then submit it to the network. It is analog for `submitTxConstraintsWith` function in Plutus and replaces `Helpers.buildBalanceSignAndSubmitTx`.
```
errfrom, line 39:
```suggestion
* Add `submitTxFromConstraints` and `submitTxFromConstraintsReturningFee` in `Contract.Transaction`. `submitTxFromConstraints` builds a transaction that satisfies the constraints, then submits it to the network. It is analog for `submitTxConstraintsWith` function in Plutus and replaces `Helpers.buildBalanceSignAndSubmitTx`.
```

errfrom (approved):
LGTM, except for one minor thing

PR #1312, opened 2022-12-09
Switch CIP25 metadata keys to use strings

### LINKED ISSUE #1309: CTL's CIP25 metadata encoding is considered invalid by Blockfrost
Blockfrost recently added validation for CIP25 metadata, and will not return metadata from the `assets/:asset:` endpoint unless the metadata is valid. Blockfrost uses [getOnchainMetadata](https://github.com/blockfrost/openapi/blob/466f0b7f511f856f3a7f48abe5c05226219b100e/src/functions/metadata.ts#L28) on [this line](https://github.com/blockfrost/blockfrost-backend-ryo/blob/10d89a2584b0d36c26a929e89bf30030caf234df/src/routes/assets/index.ts#L71) to validate the metadata.

CTL's encoding of CIP25 metadata adds a `0x` prefix to the `policy_id` and `asset_name` keys, I believe this is what is invalidating the metadata for blockfrost.

This is an example of metadata accepted by blockfrost (not encoded by CTL): https://preview.cexplorer.io/tx/2ffd85c1d2c4d9f316bfe6954db804cf27285d9848f4dfb7a52ab524cf8db48d/metadata#data

This is an example of metadata encoded by CTL and not accepted by blockfrost:
https://preview.cexplorer.io/tx/b2749c9b60d3c4d4f9592b51cee31d7129ab4faaabd58a5d0ccc18d5b5111d6e/metadata#data

The CTL revision that encoded the offending metadata is quite old (`09540ea3915be20e5095b3b6f2418ddd712eb58e`), but afaict there hasn't been any changes to the metadata encoding since then.

An argument could be made that Blockfrost is not following the [CIP25 specification](https://cips.cardano.org/cips/cip25/), since the spec says "raw bytes" should be used for `policy_id` and `asset_name` for version 2. However it also doesn't define what "raw bytes" are, so we may just want to conform to Blockfrost.

rynoV:
Closes #1309.

This is confirmed to fix the issue of Blockfrost not accepting the CIP25 metadata through testing in Seabug using the [`seabug/main`](https://github.com/Plutonomicon/cardano-transaction-lib/tree/seabug/main) branch of CTL. [This](https://preview.cexplorer.io/tx/1e48e0891c3d597b5c8973f7a507bcb7d009a217fa6ed9675d5e13289d869845/metadata#data) is an example of the new CIP25 metadata encoding, and the corresponding response returned from Blockfrost from https://cardano-preview.blockfrost.io/api/v0/assets/1e4409ba69fb38887c23d15a766476384085b66a755530934938abfeabcdef:

```json
{
  "asset": "1e4409ba69fb38887c23d15a766476384085b66a755530934938abfeabcdef",
  "policy_id": "1e4409ba69fb38887c23d15a766476384085b66a755530934938abfe",
  "asset_name": "abcdef",
  "fingerprint": "asset1a63v0u52tvacvp60tyykn0chjfsgtqg3p8a9am",
  "quantity": "1",
  "initial_mint_tx_hash": "1e48e0891c3d597b5c8973f7a507bcb7d009a217fa6ed9675d5e13289d869845",
  "mint_or_burn_count": 1,
  "onchain_metadata": {
    "name": "Allium",
    "image": "ipfs://k2cwuee3arxg398hwxx6c0iferxitu126xntuzg8t765oo020h5y6npn",
    "description": "From pixabay"
  },
  "onchain_metadata_standard": "CIP25v2",
  "metadata": null
}
```

Note that the `seabug/main` branch is significantly behind `develop`, however the only difference between the changes on that branch and on this PR's branch is the import locations.

### Pre-review checklist

### DIFF
--- a/src/Internal/Metadata/Cip25/V2.purs
+++ b/src/Internal/Metadata/Cip25/V2.purs
@@ -224,6 +224,26 @@ metadataEntryDecodeAeson policyId assetName =
     pure $
       wrap { policyId, assetName, name, image, mediaType, description, files }
 
+-- | Encode the entry's policy id to the string used as the metadata
+-- | key
+encodePolicyIdKey :: Cip25MetadataEntry -> String
+encodePolicyIdKey (Cip25MetadataEntry { policyId }) =
+  rawBytesToHex $ scriptHashToBytes $ unwrap policyId
+
+-- | Decode the CIP25 policy id key
+decodePolicyIdKey :: String -> Maybe MintingPolicyHash
+decodePolicyIdKey = map wrap <<< scriptHashFromBytes <=< hexToRawBytes
+
+-- | Encode the entry's asset name to the string used as the metadata
+-- | key
+encodeAssetNameKey :: Cip25MetadataEntry -> String
+encodeAssetNameKey (Cip25MetadataEntry { assetName }) =
+  byteArrayToHex $ getTokenName $ unwrap assetName
+
+-- | Decode the CIP25 asset name key
+decodeAssetNameKey :: String -> Maybe Cip25TokenName
+decodeAssetNameKey = map wrap <<< mkTokenName <=< hexToByteArray
+
 newtype Cip25Metadata = Cip25Metadata (Array Cip25MetadataEntry)
 
 derive instance Generic Cip25Metadata _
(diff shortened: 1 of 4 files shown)


klntsky (changes requested):
Looks like ToData/FromData instances haven't been updated in the same manner.

Please add a golden test for your example as well to `Test.Ctl.Metadata.Cip25`

rynoV:
I left the `DecodeAeson` instance alone as it already works for the updated metadata format. It does accept more than our format as it expects a utf-8 string for the asset name, instead of just hex, but I'm not sure what this instance is used for and I don't want to make an unnecessary breaking change

src/Internal/Metadata/Cip25/V2.purs
```diff
@@ -142,6 +143,7 @@ instance DecodeAeson Cip25V2 where
       2 -> pure Cip25V2
       _ -> Left $ TypeMismatch "Cip25V2"
 
+-- Why not an instance of ToMetadata?
```
klntsky, line 146: It's not useful alone, so it was done that way to prevent confusion between `Cip25Metadata` and `Cip25MetadataEntry` (the users do not need to deal with data-representations of a single entry, because the standard only specifies the encoding for Cip25Metadata, that is very different from our representation in PS (k/v object vs. an array)). Please remove this comment, or write a new one based on this clarification.

klntsky (changes requested):
Other than that, looks good.

klntsky (approved):

Issue #1120, opened 2022-10-17
Implement query stubs for blockfrost backend handle.


klntsky:
Implementation of Blockfrost query backend should be done in parallel by multiple developers. Implement query stubs for blockfrost backend handle that will be created in #1119

uhbif19:
@klntsky Do this need something more than just record with `undefined` in fields? Or with undefined in functions in separate file?

It seems to me like proper stub will be hard to make work with integration tests, and I do not understand what may be other needs for it.

errfrom:
@klntsky @uhbif19 I too don't understand what is the point of writing query stubs when we can already proceed to implementing the actual queries.

PR #1314, opened 2022-12-11
Refactor `QueryBackend` and `QueryBackendParams`

errfrom:
### Pre-review checklist

### DIFF
--- a/src/Internal/Contract/Monad.purs
+++ b/src/Internal/Contract/Monad.purs
@@ -54,6 +54,7 @@ import Ctl.Internal.QueryM
   , mkOgmiosWebSocketAff
   , underlyingWebSocket
   )
+-- TODO: Move/translate these types into Cardano
 import Ctl.Internal.QueryM.Ogmios (ProtocolParameters, SystemStart) as Ogmios
 import Ctl.Internal.QueryM.ServerConfig (ServerConfig)
 import Ctl.Internal.Serialization.Address (NetworkId)
@@ -192,7 +193,7 @@ mkContractEnv params = do
     b2 <- parallel do
       wallet <- buildWallet
       pure $ merge { wallet }
-    -- compose the sub-builders together
+    -- Compose the sub-builders together
     in b1 >>> b2 >>> merge { usedTxOuts }
 
   pure $ build envBuilder constants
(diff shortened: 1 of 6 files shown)


src/Internal/Contract/Monad.purs
```diff
@@ -60,11 +54,7 @@ import Ctl.Internal.QueryM
   , mkOgmiosWebSocketAff
   , underlyingWebSocket
   )
--- TODO Move/translate these types into Cardano
```
rynoV, line 63: Does this TODO have a resolution or an issue?

src/Internal/Contract/Monad.purs
```diff
@@ -197,7 +187,7 @@ mkContractEnv params = do
     b1 <- parallel do
       backend <- buildBackend logger params.backendParams
       -- Use the default backend to fetch ledger constants
```
rynoV, line 189: Comment seems redundant

src/Internal/Contract/Monad.purs
```diff
@@ -257,11 +248,11 @@ getLedgerConstants
        , systemStart :: Ogmios.SystemStart
        }
 getLedgerConstants logger = case _ of
-  CtlBackend { ogmios: { ws } } -> do
+  CtlBackend { ogmios: { ws } } _ -> do
     pparams <- getProtocolParametersAff ws logger
     systemStart <- getSystemStartAff ws logger
     pure { pparams, systemStart }
-  BlockfrostBackend _ -> undefined
+  BlockfrostBackend _ _ -> undefined
```
rynoV, line 255: Will this `undefined` not be an issue? For example [here](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1314/files#diff-cd9604d9a86ee84c0def99f84ddea1aa930a7b511a2ada6493fdea9e61ced7cfR190)

src/Internal/Contract/Monad.purs
```diff
@@ -279,18 +270,20 @@ walletNetworkCheck envNetworkId wallet = do
...
+stopContractEnv { backend } = liftEffect do
+  traverse_ stopCtlRuntime (getCtlBackend backend)
+  traverse_ stopBlockfrostRuntime (getBlockfrostBackend backend)
+  where
+  stopCtlRuntime :: CtlBackend -> Effect Unit
+  stopCtlRuntime { ogmios, odc } =
+    stopWebSocket odc.ws *> stopWebSocket ogmios.ws
+
+  stopBlockfrostRuntime :: BlockfrostBackend -> Effect Unit
+  stopBlockfrostRuntime = undefined
```
rynoV, line 283: Why include this if it doesn't do anything?

rynoV (changes requested):
Just a few things, but I agree this is a better way of doing it

src/Internal/Contract/Monad.purs
Luis-omega: I think it's okay as long as we don't export to the `BlockfrostBackend` to public interface yet. After all the point of #1119  is to  be able to work this in #1120

Luis-omega (approved):
I like this more, is less configurable but I don't think we have to worry about more backends in a while.

src/Internal/Contract/Monad.purs
rynoV: Ah I see, thanks for the info

src/Internal/Contract/Monad.purs
errfrom: Not yet, restored this TODO

src/Internal/Contract/Monad.purs
errfrom: But we still use the default backend, not the optional one

jy14898 (approved):
Yeah I prefer this design, it's always nicer when your invariants are directly encoded in the datatype

src/Internal/Contract/QueryBackend.purs
```diff
@@ -136,21 +41,34 @@ type BlockfrostBackend =
...
   }
 
-data QueryBackend
-  = CtlBackend CtlBackend
-  | BlockfrostBackend BlockfrostBackend
+getCtlBackend :: QueryBackend -> Maybe CtlBackend
+getCtlBackend (CtlBackend backend _) = Just backend
+getCtlBackend (BlockfrostBackend _ backend) = backend
+
+getBlockfrostBackend :: QueryBackend -> Maybe BlockfrostBackend
```
uhbif19, line 48: Why do we need those? And `Maybe` parameter at all?

PR #1320, opened 2022-12-13
Add missing collateral fields to `ProtocolParamUpdate`

### LINKED ISSUE #1296: ProtocolParamUpdate missing vasil collateral fields
An oversight, both the CSL (`src/Internal/Serialization/Types.purs`) and our Cardano (`src/Internal/Cardano/Types/Transaction.purs`) `ProtocolParamUpdate` types are missing `collateral_percentage` and `max_collateral_inputs` and the associated functions to work with them.

pb99u069:
Closes #1296.

### Pre-review checklist


test/Fixtures.purs
```diff
@@ -321,6 +321,8 @@ proposedProtocolParameterUpdates1 = ProposedProtocolParameterUpdates $
         , maxBlockExUnits: Just
             { mem: BigInt.fromInt 1, steps: BigInt.fromInt 1 }
         , maxValueSize: Just $ UInt.fromInt 1
+        , collateralPercentage: Nothing -- Just $ UInt.fromInt 140
+        , maxCollateralInputs: Nothing -- Just $ UInt.fromInt 10
```
pb99u069, line 325: @klntsky I can't figure out why the unit test `Deserialization is inverse to serialization #4` fails when I set the two new fields to a `Just` value. The error msg is `Cannot deserialize bytes`. meanwhile, if the fields are `Nothing`, the test passes

klntsky: @pb99u069 use `spy` function to debug this (you can detect where decoding fails by inserting it in a pure expression)

jy14898: If you print the error message CSL returns in `fromBytes`, it seems confused and thinks the value we pass as a collateral_percentage is a key in some map. So either we're doing something wrong on the to_bytes/CSL building side, or CSL has a bug

pb99u069: The functions `set_collateral_percentage` and `set_max_collateral_inputs` [here](https://github.com/Emurgo/cardano-serialization-lib/blob/fa2a9c94bcd1e3a7353852915a0f2e666ec0027f/rust/src/lib.rs#L2690) differ from all the other setters, as they don't use `clone()` at the end. Maybe that's our problem?

jy14898: I saw that too but there are other cases that aren't cloned, `max_tx_size` for example

klntsky: @jy14898 
> it seems confused and thinks the value we pass as a collateral_percentage is a key in some map

Could you please post the message for me to find where is it thrown from?

jy14898: Definitely a bug with CSL:

```
> ppu = lib.ProtocolParamUpdate.new()
ProtocolParamUpdate { ptr: 1668248 }
> ppu.set_collateral_percentage(42)
undefined
> ppu.collateral_percentage()
42
> lib.ProtocolParamUpdate.from_bytes(ppu.to_bytes()).collateral_percentage()
undefined
```

jy14898: @klntsky I used the following to always print the error, as we hush it in purescript. But the error is misleading due to nested decoding, the crux of the problem is as I posted above

```diff
--- a/src/Internal/Deserialization/FromBytes.js
+++ b/src/Internal/Deserialization/FromBytes.js
@@ -11,6 +11,7 @@ const fromBytes = name => helper => bytes => {
   try {
     return helper.valid(lib[name].from_bytes(bytes));
   } catch (e) {
+    console.log(name + ".from_bytes() raised " + e);
     return helper.error(name + ".from_bytes() raised " + e);
   }
 };
```

klntsky: Made an issue
https://github.com/Emurgo/cardano-serialization-lib/issues/564

pb99u069: https://github.com/Emurgo/cardano-serialization-lib/pull/565

klntsky: Yep, but the release hasn't been cut yet (I hope it will be soon)

jy14898 (approved):
LGTM. Not sure if we need a changelog entry for this, given nobody filed a bug for it

jy14898 (changes requested):
Actually, we need to update the template too

package.json
```diff
@@ -34,6 +34,7 @@
     "@emurgo/cardano-serialization-lib-browser": "11.2.1",
     "@emurgo/cardano-serialization-lib-nodejs": "11.2.1",
     "base64-js": "^1.5.1",
+    "@noble/secp256k1": "^1.7.0",
```
pb99u069, line 37: It's strange that this dependency was missing. I guess we need it?

pb99u069:
when I run `npm i -packages-lock-only` in the template I get the warning that there are 9 vulnerabilites. Which is strange as we have the same dependencies as in the main package.json. Any idea why this could be @jy14898?

jy14898:
@pb99u069 Maybe we dealt with these package-lock issues for CTL but not the template, we'd need to diff the locks to find out where they don't agree

klntsky (approved):
Thanks for debugging this tricky one!

jy14898 (approved):

Issue #1311, opened 2022-12-09
Tx chaining doesn't work with Eternl after switching to Kupo for `getUtxoByOref`


errfrom:
In order to make transaction chaining work with Eternl, we use a hacky solution in which we wait for `additionalUtxos` to be confirmed by the wallet before signing a transaction. 
https://github.com/Plutonomicon/cardano-transaction-lib/blob/a83dff3465f67c8d7b339bd5925a1440c5aab977/src/Internal/QueryM/Sign.purs#L56-L63

After switching to Kupo for utxos-related queries, `getUtxoByOref`, which is used by `walletWaitForInputs`, now fails to request utxos produced by yet unconfirmed transactions.

Solution: Update `walletWaitForInputs` to not request full utxos and just use output references to compare with entries returned by the wallet.

klntsky:
Will be addressed by https://github.com/Plutonomicon/cardano-transaction-lib/issues/1301

PR #1326, opened 2022-12-14
Eternl wallet: respect UTxO locking #1301

### LINKED ISSUE #1301: Eternl wallet: respect UTxO locking
Eternl UTxO locking feature (possibility for the user to prevent specific UTxOs from being spent by marking them as "locked" and thus hiding them from CIP-30 `getUtxos` output) is not compatible with our hack to make Eternl "see" the Utxos we are spending, namely, `waitForWalletInputs` function. See https://github.com/Plutonomicon/cardano-transaction-lib/issues/1282

We need to change our approach, specifically we should use Eternl (CIP-30) query layer as a source of truth.

In the balancer, whenever we query for utxos at own addresses, if the wallet is Eternl, we must use `getWalletUtxos` (CIP-30) instead of `utxosAt` (Kupo). Instead of waiting for Eternl to synchronize with our query layer, we must do the opposite: check that utxos that are seen by `utxosAt` are a superset of `getWalletUtxos` call output - it should be done on an earlier stage [(in the balancer)](https://github.com/Plutonomicon/cardano-transaction-lib/blob/2b6ea92809d72e93764f52acc42f1d1cdda444cb/src/Internal/BalanceTx/BalanceTx.purs#L161)

Note that we use `Constraints._srcAddresses` that may contain addresses not controlled by Eternl. So, we must first separate these addresses into those that are used by the wallet (CIP-30 `getUsedAddresses`) and those that are not. The latter group should be treated using our query layer (utxosAt) the same way as before.

Luis-omega:
Closes #1301  and #1311 

### Pre-review checklist


Luis-omega:
I tested it by hand in the e2e browser since the `npm run e2e-test` command keep failing all the test. All the test passed in the browser. What remains is to refactor the code as right now is a mess.

src/Internal/QueryM/Sign.purs
```diff
@@ -5,32 +5,23 @@ module Ctl.Internal.QueryM.Sign
...
-import Effect.Aff (delay, error)
+import Data.Maybe (Maybe(Just))
+import Data.Newtype (unwrap)
+import Data.Traversable (for_)
 import Effect.Aff.Class (liftAff)
 import Effect.Class (liftEffect)
-import Effect.Exception (throw, try)
+import Effect.Exception (try)
 
 signTransaction
```
Luis-omega, line 26: I'm tempted to just remove this module and put `signTransaction` under `QueryM` and use `actionBasedOnWallet` to make it shorter.

Luis-omega:
I'm still having problems running `e2e-test` locally. Also, I don't know if we can lock UTxOs using Eternl API to make a test specific to the issue, perhaps we can control the browser to lock some UTxOs if they don't have one. For now, I'm satisfied with the pr passing all the current e2e-test in ci.

Luis-omega:
I was curious why the ci test are passing but not the local ones, it seems that ci only uses `keywallet as nami` for e2e test.

```
E2E tests
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:OneShotMinting (7.23s)
127.0.0.1 - - [16/Dec/2022 07:19:00] "GET /?plutip-nami-mock:SignMultiple HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:19:00] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:19:00] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:SignMultiple (7.09s)
127.0.0.1 - - [16/Dec/2022 07:19:07] "GET /?plutip-nami-mock:AlwaysSucceeds HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:19:07] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:19:07] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:AlwaysSucceeds (9.16s)
127.0.0.1 - - [16/Dec/2022 07:19:16] "GET /?plutip-nami-mock:AlwaysSucceedsV2 HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:19:16] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:19:16] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:AlwaysSucceedsV2 (8.04s)
127.0.0.1 - - [16/Dec/2022 07:19:24] "GET /?plutip-nami-mock:AlwaysMints HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:19:25] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:19:25] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:AlwaysMints (7.24s)
127.0.0.1 - - [16/Dec/2022 07:19:31] "GET /?plutip-nami-mock:Pkh2Pkh HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:19:32] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:19:32] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:Pkh2Pkh (7.16s)
127.0.0.1 - - [16/Dec/2022 07:19:39] "GET /?plutip-nami-mock:SendsToken HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:19:39] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:19:39] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:SendsToken (9.18s)
127.0.0.1 - - [16/Dec/2022 07:19:48] "GET /?plutip-nami-mock:MintsMultipleTokens HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:19:48] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:19:48] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:MintsMultipleTokens (7.08s)
127.0.0.1 - - [16/Dec/2022 07:19:55] "GET /?plutip-nami-mock:Schnorr HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:19:56] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:19:56] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:Schnorr (9.79s)
127.0.0.1 - - [16/Dec/2022 07:20:05] "GET /?plutip-nami-mock:ECDSA HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:20:06] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:20:06] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:ECDSA (9.85s)
127.0.0.1 - - [16/Dec/2022 07:20:16] "GET /?plutip-nami-mock:ApplyArgs HTTP/1.1" 200 -
127.0.0.1 - - [16/Dec/2022 07:20:16] code 404, message File not found
127.0.0.1 - - [16/Dec/2022 07:20:16] "GET /favicon.ico HTTP/1.1" 404 -
  ✓︎ plutip:http://localhost:4008/?plutip-nami-mock:ApplyArgs (7.47s)
Summary
11/11 tests passed
```
So moving this pr again to draft

Luis-omega:
Current state: 
- Most test now shows the same error for eternl e2e-test as in `develop`.
- `OneShotMinting` and `ECDSA` examples have it's assertions broken.

I will try to fix the problem with eternl test in this pr, but it seems that the main changes for #1301 are done and working, so opening for reviews again.

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -190,6 +213,24 @@ balanceTxWithConstraints unbalancedTx constraintsBuilder = do
       =<< maybe (liftQueryM QueryM.getChangeAddress) (pure <<< Just)
       =<< asksConstraints Constraints._changeAddress
 
+  getWalletUtxosAt :: Array Address -> QueryM (Either BalanceTxError UtxoMap)
+  getWalletUtxosAt inWalletAddresses = do
+    filteredWalletUtxos <-
+      maybe (Map.empty)
```
amirmrad, line 219: I believe the parens around `Map.empty` are redundant.

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -190,6 +213,24 @@ balanceTxWithConstraints unbalancedTx constraintsBuilder = do
...
+        ( Map.filter
+            (\txout -> elem (_.address (unwrap txout)) inWalletAddresses)
+        ) <$> getWalletUtxos
+    utxosAtWalletAddresses <- parallelUtxosAt inWalletAddresses
+    pure $ case Map.isSubmap filteredWalletUtxos <$> utxosAtWalletAddresses of
```
amirmrad, line 224: I'm running into an issue with the logic here.
Specifically, now that `waitForInputs` is no longer used, when running AlwaysSucceeds with a newly initialized wallet (or wallet with 1 spendable UTxO) there is a scenario, after the initial payment to the alwaysSucceedScript, in which Eternl hasn't yet updated it's UTxO list and therefore `filteredWalletUtxos` is no longer a subset of `utxosAtWalletAddress` and so the example fails.

_I'm currently investigating further, will add more details and logs as I go forward_

src/Internal/QueryM/Sign.purs
```diff
@@ -44,57 +35,10 @@ signTransaction tx = do
...
       pure $ Just (tx # _witnessSet <>~ witnessSet)
 
--- | Waits till all provided inputs of a given transaction appear in the UTxO
--- | set provided by the wallet.
--- | This is a hacky solution to the problem of Eternl not seeing UTxOs that
--- | hasn't been fully confirmed at the moment of a `sign()` call.
--- | Since it can't detect UTxO origin, it can't decide which of the private
--- | keys to use for signing. As a result, we get `MissingVKeyWitnesses`.
-walletWaitForInputs :: Array TransactionInput -> QueryM Unit
-walletWaitForInputs txInputs = do
```
amirmrad, line 66: Could you please remind of why we decided that this was no longer necessary?

src/Internal/BalanceTx/BalanceTx.purs
Luis-omega: Yes, this is the bug I have been tracing all weekend, I came to the conclusion that using `getWalletUtxos` in the balancer once before beginning, has the effect of allowing the first test in the e2e suit to pass and I was suspecting it because it forced the wallet to update it' utxos. This problem is affecting `develop` right now.

Sadly it only works for the fist contract that is run in the suite, the rest of them fail after that, so I need to find proper way to force the update.

Luis-omega: I would try to add a waiting loop in the balancer and see if it helps

Luis-omega: It didn't help, I did something like `run this function once every second for 20 seconds and keep the last returned result`. It seems that the problem may be that the addresses are outdated. For some reason, I keep seeing that every transaction generates a new address for the wallet and all the assets that aren't locked in the contract are moved to the new address. While constructing the transaction the addresses are outdated and so the new address with all the assets isn't part of the return of `asksConstraints Constraints._srcAddresses` in the balancer, effectively hiding the utxos in this algorithm. I would test it tomorrow.

src/Internal/QueryM/Sign.purs
amirmrad: What if we waited for all the outputs of a Tx that belong to self show up in `getWalletUtxos` at the `awaitTxConfirmed` stage?

src/Internal/BalanceTx/BalanceTx.purs
amirmrad: With regards to the new addresses, from what I've seen, `getChangeAddress` generates a new address on every call. But the address is not returned from `getUsedAddresses` until the wallet can see the outputs of the Tx.

Any wallet that is multi-address will probably do this, I know Yoro certainly does.

src/Internal/QueryM/Sign.purs
klntsky: No guarantee that the utxos are not spent in another Contract or app.

src/Internal/BalanceTx/BalanceTx.purs
Luis-omega: tanks! I was wondering if I misunderstood how addresses work.

Luis-omega:
Status: 

I put two loops waiting until the utxos and address are updated and it seems to work: 

- In the browser we still have some errors happening but none of them related to the previous problem, and most of the tests are successful now.
- in the terminal, we get a timeout, so we need to augment the timeout for eternl wallet, I don't know how to do that, I would figure it out tomorrow. @klntsky Can you give me some pointers on how todo it please?

The only error that I think I can't solve is 

```
  ✗ nami:http://localhost:4008/?nami:SignMultiple:

  Error: Error: "Insufficient Utxos for 2 transactions"
```

but it seems to be a common problem in all wallets.

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -190,6 +201,68 @@ balanceTxWithConstraints unbalancedTx constraintsBuilder = do
...
+    => Int
+    -> m (Either b c)
+    -> m (Either b c)
+  repeatN attempts action =
+    if attempts <= 0 then action
+    else do
+      result <- action
+      case result of
+        Left _ -> do
+          logInfo' $ "waiting repeatN: " <> show attempts
```
Luis-omega, line 256: I would like to preserv logs of the wait but I need to change them to `Trace` or `Debug`

I would refactor `repeatN` function anyway.

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -551,3 +624,8 @@ logTx msg utxos (Transaction { body: body'@(TxBody body) }) =
     , "Output Value: " <> show (Array.foldMap getAmount body.outputs)
     , "Fees: " <> show body.fee
     ]
+
+-- The same logInfo' as in Contract.Log
```
Luis-omega, line 628: Temporal hack, but it's quite possible to become a permanent hack.

Luis-omega:
after runing the tests with differents timeouts, the only persisting problem is the following error : 

```
  ✗ eternl:http://localhost:4008/?eternl:OneShotMinting:

  Error: Error: `submit` call failed. Error from Ogmios: [{"missingVkWitnesses":["4d149ae923cdce83dad20c83c1bb82b5da3d0940e057dd05fe1a5438"]}]
```

it's happening on 3 examples but the rest of them are working right now with eternl!

Renegatto:
> after runing the tests with differents timeouts, the only persisting problem is the following error :
> 
> ```
>   ✗ eternl:http://localhost:4008/?eternl:OneShotMinting:
> 
>   Error: Error: `submit` call failed. Error from Ogmios: [{"missingVkWitnesses":["4d149ae923cdce83dad20c83c1bb82b5da3d0940e057dd05fe1a5438"]}]
> ```
> 
> it's happening on 3 examples but the rest of them are working right now with eternl!

I had the same error when tried to run contract while my CTL runtime had not been completely synchronized yet.
Maybe the reason of your error is somewhat similar 🤷.

klntsky:
We don't need this PR anymore after https://github.com/Plutonomicon/cardano-transaction-lib/pull/1440 - the problem should be fixed

PR #1328, opened 2022-12-14
Add a note on Nix caches

klntsky:


rynoV (approved):

uhbif19 (approved):

Bradley-Heather (approved):
Good idea!

doc/getting-started.md
```diff
@@ -25,6 +26,12 @@ This guide will help you get started writing contracts with CTL. Please also see
 
 ## Prerequisites
 
+### Setting up Nix caches
+
+Nix caches improve build speeds by allowing to download pre-built dependencies instead of compiling them from source.
+
+Follow [these instructions](https://github.com/input-output-hk/cardano-ledger#nix-cache) to add the caches.
```
uhbif19, line 33: As far as I understand, without MLabs private caches there will be still many builds happening.

klntsky: I don't think so, `Push+pull caches` are set to `public-plutonomicon` in CI config.

klntsky: Obviously, it does not guarantee that we don't need to use mlabs cachix

klntsky: I added public-plutonomicon cache

klntsky: I asked what to do with mlabs cachix internally

PR #1342, opened 2022-12-16
Use safe matching of tagged CSL types via CSL's exported kinds where possible

jy14898:
Wanted to make converting from CSL Metadata 'free' (Not in `Maybe`/`Either`) for the kupo backend and got carried away

Other than `PlutusData` (which could be done in the plutusdata pr), we now use `.kind()` and the appropriate enums exported by CSL everywhere.

I also noticed we do conversions from `Int` to `UInt` in places where the flowtypes say CSL outputs `Number`s, but actually CSL internally uses `u32` (`UInt` in purescript) and it's a limitation for the flowtype generation that causes them to be represented as `Number`s. There's probably a lot more cases of these, but it requires looking at CSLs source to know

### Pre-review checklist


klntsky (approved):

klntsky:
We can merge it as long as there are no more build warnings

jy14898:
Thanks

PR #1354, opened 2022-12-21
Disallow selecting inputs with inline datums/script refs when plutusv1 scripts are used during balancing

### LINKED ISSUE #1349: Re-think wallet utxos with datum/script references
Currently there are two issues I see:
  1. We support paying into wallets with datum via `mustPayToPubKey(Address)WithDatum`, but no way of spending such utxos manually via `mustSpendPubKeyOutput`, as it checks there is no datum (possibly from legacy models about Ogmios).
  2. In the presence of such utxos in the users wallet, the balancer may select them. This is an issue for inline datum, as it will make it impossible to construct the `ScriptContext` for any PlutusV1 script. We should kick into a PlutusV1 compatible selection mode when necessary, only selecting inputs which will make a valid transaction.

This may also be an issue with script references too

jy14898:
Closes #1349.

### Pre-review checklist


klntsky (approved):

jy14898:
Don't merge just yet, there's another case I'd like to test

rynoV (approved):

klntsky (approved):

PR #1360, opened 2022-12-22
Fix concurrency problem in staking tests affecting CI

klntsky:


jy14898 (changes requested):
Other than these changes, LGTM

test/Plutip/Staking.purs
```diff
@@ -113,8 +113,7 @@ suite = do
   -- We must never select this pool, because it retires at the third epoch
   -- (this is Plutip internal knowledge)
   -- https://github.com/mlabs-haskell/plutip/blob/7f2d59abd911dd11310404863cdedb2886902ebf/src/Test/Plutip/Internal/Cluster.hs#L692
-  retiringPoolId <- liftEffect
-    $ maybe (throw "unable to decode poolId bech32") pure
+  retiringPoolId <- liftEffect $ liftM (error "unable to decode poolId bech32")
```
jy14898, line 116: I don't think this `liftEffect` is necessary, due to `liftM` being over any `m` with `MonadError Error m`

jy14898 (approved):
LGTM, small comment but not important

rynoV (approved):

rynoV (commented):
Suggestion for changelog:

```
- Module `Contract.Transaction` now exports `mkPoolPubKeyHash` and `poolPubKeyHashToBech32` for bech32 string <-> `PoolPubKeyHash` conversion
```

zmrocze:
The spawn and configuration of this pool is handled at plutip level already: 
https://github.com/mlabs-haskell/plutip/blob/7f2d59abd911dd11310404863cdedb2886902ebf/src/Test/Plutip/Internal/Cluster.hs#L690
so the change could have been made there

Issue #1365, opened 2022-12-26
Implement Blockfrost queries based on OpenAPI spec [Metatask]


errfrom:
This issue lists all the Blockfrost queries that we need to implement, divided into logical blocks so that the work can be done by multiple developers in parallel. Feel free to assign yourself **(please put your username next to the task you are working on)**.

All work should be based on [#1260 Make Contract query backend configurable](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1260).
***
### 1. Get utxos at address / by output reference (@errfrom https://github.com/Plutonomicon/cardano-transaction-lib/pull/1366)
- [x] `utxosAt` - https://docs.blockfrost.io/#tag/Cardano-Addresses/paths/~1addresses~1{address}~1utxos/get
- [x] `getUtxoByOref` - https://docs.blockfrost.io/#tag/Cardano-Transactions/paths/~1txs~1{hash}~1utxos/get
***
### 2. Get transaction metadata, Check transaction confirmation status (@jy14898 https://github.com/Plutonomicon/cardano-transaction-lib/pull/1368)
- [x] `getTxMetadata` - https://docs.blockfrost.io/#tag/Cardano-Transactions/paths/~1txs~1{hash}~1metadata/get
- [x] `isTxConfirmed` - https://docs.blockfrost.io/#tag/Cardano-Transactions/paths/~1txs~1{hash}/get
***
### 3. Submit / evaluate a transaction (@jy14898 https://github.com/Plutonomicon/cardano-transaction-lib/pull/1376)
- [x] `submitTx` - https://docs.blockfrost.io/#tag/Cardano-Transactions/paths/~1tx~1submit/post
- [x] `evaluateTx` - https://docs.blockfrost.io/#tag/Cardano-Utilities/paths/~1utils~1txs~1evaluate/post
***
### 4. Get datums / scripts by hash (@uhbif19 #1378)
- [x] `getDatumByHash` - https://docs.blockfrost.io/#tag/Cardano-Scripts/paths/~1scripts~1datum~1{datum_hash}~1cbor/get
- [x] `getScriptByHash`:
  - get `timelock` script by hash - https://docs.blockfrost.io/#tag/Cardano-Scripts/paths/~1scripts~1{script_hash}~1json/get
  - get `plutus` script by hash - https://docs.blockfrost.io/#tag/Cardano-Scripts/paths/~1scripts~1{script_hash}~1cbor/get
***
[...]

PR #1368, opened 2022-12-27
Blockfrost metadata and confirmation

jy14898:
Part of #1365

### Pre-review checklist


src/Internal/QueryM/Kupo.purs
```diff
@@ -156,19 +165,25 @@ isTxConfirmedAff config (TransactionHash txHash) = runExceptT do
...
-      -- `KupoMetadata` will fail parsing if there is more than one response
-      pure $ unwrapKupoMetadata metadata
+      kupoMetadata <- ExceptT $
+        lmap GetTxMetadataClientError <<< handleAffjaxResponse <$>
+          kupoGetRequest
+            endpoint
+      case unwrapKupoMetadata kupoMetadata of
+        Nothing -> throwError GetTxMetadataMetadataEmptyOrMissingError
+        Just metadata
+          | Map.isEmpty (unwrap metadata) -> throwError
```
uhbif19, line 184: Why we need to handle empty Map case?

jy14898: To make the API the same for blockfrost and kupo.

Blockfrost always gives a response for metadata, regardless of whether it actually had the optional metadata field attached. Kupo on the other hand returns exactly what the transaction contained,

The alternative could be to always make the kupo version return an empty metadata on the case where there was none, but I think if users are requesting metadata from a transaction, they probably expect some metadata.

jy14898: Not sure if I explained that well, maybe this makes it clearer:

No metadata:
blockfrost: []
kupo: []

Empty metadata:
blockfrost: []
kupo: [ ... ]

With metadata:
blockfrost: [ ... ]
kupo: [ ... ]

src/Internal/Service/Blockfrost.purs
```diff
@@ -88,14 +88,14 @@ runBlockfrostServiceM backend = flip runReaderT serviceParams
 --------------------------------------------------------------------------------
 
 data BlockfrostEndpoint
-  = GetTransaction TransactionHash
-  | GetTransactionMetadata TransactionHash
+  = Transaction TransactionHash
```
uhbif19, line 91: Why change from `Get`?

klntsky (approved):
Looks good

klntsky (approved):
Looks good to me

PR #1371, opened 2022-12-29
Add framework for implementing Blockfrost queries

errfrom:
### Pre-review checklist


src/Internal/Service/Blockfrost.purs
```diff
@@ -0,0 +1,95 @@
...
+blockfrostGetRequest endpoint = ask >>= \params -> liftAff do
+  Affjax.request $ Affjax.defaultRequest
+    { method = Left GET
+    , url = mkHttpUrl params.blockfrostConfig <> realizeEndpoint endpoint
+    , responseFormat = Affjax.ResponseFormat.string
+    , headers =
+        maybe mempty (\apiKey -> [ Affjax.RequestHeader "project_id" apiKey ])
+          params.blockfrostApiKey
+    }
+
```
jy14898, line 74: Can we add this while we're at it?

```suggestion
blockfrostPostRequest
  :: BlockfrostEndpoint
  -> MediaType
  -> Maybe RequestBody
  -> BlockfrostServiceM (Either Affjax.Error (Affjax.Response String))
blockfrostPostRequest endpoint mediaType mbContent = ask >>= \params -> liftAff do
  Affjax.request $ Affjax.defaultRequest
    { method = Left POST
    , url = mkHttpUrl params.blockfrostConfig <> realizeEndpoint endpoint
    , content = mbContent
    , responseFormat = Affjax.ResponseFormat.string
    , headers =
        [ Affjax.ContentType mediaType ] <>
        maybe mempty (\apiKey -> [ Affjax.RequestHeader "project_id" apiKey ])
          params.blockfrostApiKey
    }
```

src/Internal/Service/Error.purs
```diff
@@ -0,0 +1,89 @@
...
+import Data.Newtype (class Newtype)
+import Data.Show.Generic (genericShow)
+
+--------------------------------------------------------------------------------
+-- ClientError
+--------------------------------------------------------------------------------
+
+data ClientError
+  = ClientHttpError Affjax.Error
+  | ClientHttpResponseError ServiceError
```
jy14898, line 31: Can we add the status code to `ClientHttpResponseError`, like I have in my PR? https://github.com/Plutonomicon/cardano-transaction-lib/blob/a4b545e29e9e8f9a37823f1f37699f72660a6158/src/Internal/QueryM.purs#L390

I needed it to check for 404 errors to not have them be a `ClientError`, but a richer error or the result type (`isTxConfirmed` for eg will check for 404 at the `/txs/{hash}` endpoint and return `false`)

It's likely other endpoints may need to check for 404 too

jy14898 (changes requested):

src/Internal/Service/Blockfrost.purs
jy14898: Actually I guess it could be `Maybe (MediaType /\ RequestBody)`, not critical though

src/Internal/Service/Error.purs
```diff
@@ -0,0 +1,97 @@
...
+  = ServiceBlockfrostError BlockfrostError
+  | ServiceOtherError String
+
+derive instance Generic ServiceError _
+
+instance Show ServiceError where
+  show = genericShow
+
+newtype BlockfrostError = BlockfrostError
+  { statusCode :: Int
```
uhbif19, line 81: If this is always in `ClientHttpResponseError`, then status code is probably always already recorded.

src/Internal/Service/Error.purs
```diff
@@ -0,0 +1,97 @@
...
+  show (ClientOtherError err) =
+    "(ClientOtherError "
+      <> err
+      <> ")"
+
+--------------------------------------------------------------------------------
+-- ServiceError
+--------------------------------------------------------------------------------
+
+data ServiceError
```
uhbif19, line 71: I have a couple of thoughts about this.

First - it looks like all `ServiceOterError` is used by `handleAffjaxResponse` in Kupo. So maybe it should be `KupoError`? Or add `service` field to it, to track which service has failed.
Second - do we plan to actually handle such errors or just crash? If we handle them - then `ClientError` should probably be parametrized by type of `ServiceError` instead of using ADT cases, cuz only one case seems possible, and using cases complicates handling errors.

uhbif19 (approved):

uhbif19:
I think this closes #1120

src/Internal/Service/Error.purs
jy14898: >  parametrized by type of `ServiceError`

I thought about doing that too, but ideally users write contracts which do not care about the choice of backend, which would then lead to type variables everywhere to abstract over the backend... probably not worth it

src/Internal/Service/Error.purs
jy14898: Hmm, I originally requested for `StatusCode` to be added to `ClientHttpResponseError` so that I could know the status code for the blockfrost backend, but perhaps now it's not necessary.

Doesn't hurt to keep both though, `BlockfrostError` might exist away from the `ServiceError` at some point in the future?

PR #1375, opened 2023-01-02
Better event fetching code for e2e tests

kirill-havryliuk:
### Pre-review checklist

Mostly, refactoring of subscribeToBrowserEvents function.


src/Internal/Helpers.purs
```diff
@@ -282,3 +289,40 @@ concatPaths a b =
...
 
 infixr 5 concatPaths as <</>> -- </> is taken
+
+-- | Runs two `Aff` actions concurrently
+-- | Get resolved by the first which either resolves or throws
+-- | Expected properties:
+-- | race x y = race y x
+-- | race x (race y z) = race (race x y) z
+-- | race never x = x
+race :: forall (a :: Type). Aff a -> Aff a -> Aff a
```
jy14898, line 299: How does this differ from:

```purescript
race f g = sequential (parallel (try f) <|> parallel (try g)) >>= liftEither
```

kirill-havryliuk: Indeed. This should be the same. Nice catch.

src/Internal/Helpers.purs
```diff
@@ -282,3 +290,23 @@ concatPaths a b =
...
+-- | Runs multiple `Aff` actions concurrently
+-- | raceMany [] = never
+-- | raceMany [f] = f
+-- | raceMany [f, g, h] = race (race f g) h
+raceMany
+  :: forall (a :: Type) (f :: Type -> Type). Foldable f => f (Aff a) -> Aff a
+raceMany = foldl race never
+
+delaySec :: Seconds -> Aff Unit
+delaySec (Seconds seconds) = delay $ Milliseconds (seconds * 1000.0)
```
jy14898, line 312: There's also

```suggestion
delaySec :: Seconds -> Aff Unit
delaySec seconds = delay $ convertDuration seconds
```

which helps with avoiding these conversion constants

src/Internal/Test/E2E/Feedback/Node.purs
```diff
@@ -3,100 +3,71 @@
...
+        allLogs <- Ref.read logs
+        Console.log allLogs
+
+        let errStr = "Page error occured: " <> message err <> "\n" <> allLogs
+        AVarSync.kill (error errStr) eventAVar
+
+  let
+    -- Asyncronously gets browser events and pushes
+    -- to the `eventVAar` one by one.
+    watcher = fix \this -> do
```
jy14898, line 62: FWIW You don't need to fix under monadic recursion for `Aff`, ie you can use `watcher` directly and recursively here (and in the other places you have used fix, if you gave them names)

kirill-havryliuk: Sure. That's the matter of taste. I prefer to use `fix`.

kirill-havryliuk: This is just less verbose, imho.

foo bar baz quux = something *> foo bar baz quux 
foo bar baz quux = fix \this -> something *> this
foo bar baz quux = let this = foo bar baz quux in something *> this

src/Internal/Helpers.purs
kirill-havryliuk: That's a good one!

src/Internal/Test/E2E/Runner.purs
```diff
@@ -303,26 +296,73 @@ testPlan opts@{ tests } rt@{ wallets } =
...
-                      Sign -> rethrow someWallet.sign
+                  handler event = do
+                    case event of
+                      ConfirmAccess -> do
+                        handler =<< raceMany
+                          [ someWallet.confirmAccess *> never
+                          , waitNextEvent
+                          , delaySec (Seconds 100.0) *>
+                              throwError (error nextAfterConfirmAccess)
+                          ]
```
klntsky, line 310: This effectively sets 100 seconds timeout for every test. It's not what we want, the test suite should control the timeouts, not the testing engine.

src/Internal/Test/E2E/Runner.purs
```diff
@@ -303,26 +296,73 @@ testPlan opts@{ tests } rt@{ wallets } =
...
+                      Sign -> do
+                        handler =<< raceMany
+                          [ someWallet.sign *> never
+                          , waitNextEvent
+                          , delaySec (Seconds 100.0) *>
+                              throwError (error nextAfterSign)
```
klntsky, line 316: Same

src/Internal/Test/E2E/Runner.purs
```diff
@@ -303,26 +296,73 @@ testPlan opts@{ tests } rt@{ wallets } =
...
+    subscribeToBrowserEvents page \waitNextEvent -> do
+      let
+
+        handler event =
+          case event of
+            ConfirmAccess -> handler =<< raceMany
+              [ waitNextEvent
+              , delaySec (Seconds 10.0) *> throwError
+                  (error nextAfterConfirmAccess)
+              ]
```
klntsky, line 354: Same

src/Internal/Test/E2E/Runner.purs
```diff
@@ -303,26 +296,73 @@ testPlan opts@{ tests } rt@{ wallets } =
...
+            Sign -> handler =<< raceMany
+              [ waitNextEvent
+              , delaySec (Seconds 10.0) *> throwError (error nextAfterSign)
```
klntsky, line 357: Same

klntsky (changes requested):

PR #1377, opened 2023-01-03
Implement `getChainTip`, `getEraSummaries`, and `getSystemStart` Blockfrost queries

errfrom:
Part of https://github.com/Plutonomicon/cardano-transaction-lib/issues/1365

### Observations
1. Blockfrost and Ogmios return different `systemStart` values (tested on `preview`): 
Ogmios: `(SystemStart (DateTime (Date (Year 2022) October (Day 25)) (Time (Hour 0) (Minute 0) (Second 0) (Millisecond 0))))`
Blockfrost: `(SystemStart (DateTime (Date (Year 2022) October (Day 25)) (Time (Hour 10) (Minute 0) (Second 0) (Millisecond 0))))` (Fixed in https://github.com/blockfrost/blockfrost-backend-ryo/pull/53)
2. Blockfrost and Ogmios return different `end` values for the last `eraSummaries` entry, which is probably OK since the end time of the current era is unknown (see https://github.com/blockfrost/blockfrost-backend-ryo/blob/c6ca87db072f10b08e1e75ff9a7436b112972a7e/src/routes/network/index.ts#L136-L160 for reference on how the last summary entry is constructed in Blockfrost)

### Pre-review checklist


jy14898:
> Blockfrost and Ogmios return different systemStart values (tested on preview)

Blockfrost is straight up wrong AFAIU, it should be providing the value from the genesis file at https://book.world.dev.cardano.org/environments/preview/shelley-genesis.json

klntsky:
@errfrom could it be a timezone problem?

errfrom:
@klntsky Blockfrost returns the `systemStart` as a Unix timestamp (which is `1666692000` in this case). This should correspond to `Tue Oct 25 2022 10:00:00 GMT+0000`, so I doubt this is a timezone issue.

src/Contract/Time.purs
```diff
@@ -111,11 +109,12 @@ getCurrentEpoch = do
     $ BigInt.toString (bigInt :: BigInt.BigInt)
 
 -- | Get `EraSummaries` as used for Slot arithemetic.
--- | Details can be found https://ogmios.dev/api/ under "eraSummaries" query
+-- | Details can be found https://ogmios.dev/api/ under "eraSummaries" query.
```
klntsky, line 112: Is it fair to point to Ogmios docs here, given that the code is backend-agnostic?

errfrom: I think it is a good reference for the user, and Blockfrost response has an identical structure. 
Maybe we should specify a link to the Blockfrost endpoint spec as well?

klntsky: Ideally we should reference neither, because it is our own domain type that just happens to correspond to both Blockfrost's and Ogmios' ones, but since there would be no documentation otherwise, let's mention both, but form the message in a way that suggests that it is not a documentation for something we provide, but an external source.

E.g.:

More info can be found in Ogmios or Blockfrost docs (*links*), currently we use the same data type definition

jy14898 (commented):
Generally LGTM, except one comment about aeson instances

test/Plutus/Time.purs
```diff
@@ -78,8 +82,9 @@ relSlotFixture = RelSlot $ BigInt.fromInt 12855
 currentEpochFixture :: Epoch
 currentEpochFixture = Epoch $ BigInt.fromInt 58326646
 
-systemStartFixture :: SystemStart
-systemStartFixture = SystemStart "2019-07-24T20:20:16Z"
+systemStartFixture :: OgmiosSystemStart
```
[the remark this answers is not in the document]
errfrom: I think it should be reflected in the CHANGELOG that `getEraSummaries` now returns the new `EraSummaries`, and the user may want to wrap it in `OgmiosEraSummaries` in order to use `EncodeAeson` functionality.

Maybe we should create an issue about revisiting these round-trip tests and remove them if they prove to be redundant.

jy14898 (approved):
LGTM. There's an issue with log supression in the plutip tests, however (failed in CI)

PR #1384, opened 2023-01-06
Wrap CSL  to eliminate need for explicit pointer deallocation

### LINKED ISSUE #102: Figure out a way to `free` CSL objects
Unfortunately allocating memory with CSL requires us to `free` it afterwards. We might need to write some sort of pseudo garbage collector to register objects upon allocation and then free them in a monadic eliminator (e.g. `runQueryM`)

This is lower-priority and we have some time to think of the best way to approach this

amirmrad:
Closes #102  .

### Pre-review checklist


test/Wallet/Cip30/SignData.js
```diff
@@ -8,6 +8,7 @@ if (typeof BROWSER_RUNTIME != "undefined" && BROWSER_RUNTIME) {
   lib = require("@emurgo/cardano-message-signing-nodejs");
   csl = require("@emurgo/cardano-serialization-lib-nodejs");
 }
+lib = require("csl-runtime-gc")(lib);
```
jy14898, line 11: Needs to be done on both lib and csl

src/Internal/ApplyArgs.js
```diff
@@ -9,6 +9,7 @@ if (typeof BROWSER_RUNTIME != "undefined" && BROWSER_RUNTIME) {
   lib = require("@emurgo/cardano-serialization-lib-nodejs");
   apply_args = require("apply-args-nodejs");
```
jy14898, line 10: Also to apply args?

src/Internal/ApplyArgs.js
```diff
@@ -9,6 +9,7 @@ if (typeof BROWSER_RUNTIME != "undefined" && BROWSER_RUNTIME) {
   lib = require("@emurgo/cardano-serialization-lib-nodejs");
```
jy14898, line 9: What if we made a js file `src/Internal/CSL-wrapped.js` which imports and wraps the needed CSL, and then we import that from all foreign code? Then it's trivial to check all foreign imports use the wrapped one via a grep which searches for the unwrapped one

klntsky:
Closing in favor of https://github.com/Plutonomicon/cardano-transaction-lib/pull/1386

src/Internal/ApplyArgs.js
amirmrad: But won't spago move all the js files around during build? Using relative path requires doesn't seem straightforward.

amirmrad:
Reopening since wasm bindgen approach failed with errors when testing. Falling back to wrapper solution.

klntsky (approved):

PR #1386, opened 2023-01-11
Use mlabs-owned CSL fork with GC

### LINKED ISSUE #102: Figure out a way to `free` CSL objects

klntsky:
Closes #102 

### Pre-review checklist


klntsky:
This approach failed because the compilation option does not introduce code to cover the case of accidental double `free()`s. Our wrapper, introduced in https://github.com/Plutonomicon/cardano-transaction-lib/pull/1384 does.

PR #1389, opened 2023-01-11
Re-implement test utils

klntsky:
### Pre-review checklist


src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,600 @@
...
+  | cond = pure unit
+  | otherwise = tellFailure failure
+
+assertContractMaybe
+  :: forall (a :: Type)
+   . ContractAssertionFailure
+  -> Maybe a
+  -> ContractAssertion a
+assertContractMaybe msg =
+  maybe (liftEffect $ throw $ show msg) pure
```
errfrom, line 266: If such an assertion fails, does this mean that subsequent assertions won't be executed?

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,600 @@
...
+  wrapAssertion :: ContractCheck a -> ContractAssertion a -> ContractAssertion a
+  wrapAssertion assertion acc = do
+    run /\ finalize <- assertion acc
+    E.try run >>= case _ of
+      Left failure -> do
+        finalize
+        throwError failure
+      Right success -> do
+        finalize
+        pure success
```
errfrom, line 318:
```suggestion
    E.try run >>= \res -> finalize *> either throwError pure res
```

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,600 @@
...
+        _ -> ""
+      errorReport =
+        String.trim
+          ( errorStr <> "\n\n" <>
+              show (ContractAssertionFailures $ Array.fromFoldable failures)
+          ) <> "\n"
+    -- error trace from the exception itself will be appended here
+    liftEffect $ throwError $ error errorReport
+  where
+  reportException error = "\n\nAn exception has been thrown: \n\n" <> show error
```
errfrom, line 307:
```suggestion
  reportException :: Error -> String
  reportException error = "\n\nAn exception has been thrown: \n\n" <> show error
```

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,600 @@
...
+  failures <- liftEffect $ Ref.read ref
+  if null failures then either
+    (liftEffect <<< throwError <<< error <<< reportException)
+    pure
+    eiResult
+  else do
+    let
+      errorStr = case eiResult of
+        Left error -> reportException error
+        _ -> ""
```
errfrom, line 298:
```suggestion
      errorStr = either reportException (const mempty) eiResult
```

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,600 @@
...
+-- | Accepts an array of checks and interprets them into a `Contract`.
+runChecks
+  :: forall (a :: Type)
+   . Array (ContractCheck a)
+  -> ContractAssertion a
+  -> Contract a
+runChecks assertions contract = do
+  ref <- liftEffect $ Ref.new Nil
+  eiResult <- E.try $
+    flip runReaderT ref go
```
errfrom, line 288:
```suggestion
  eiResult <- E.try $ runReaderT wrappedContract ref
```

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,600 @@
...
+  -> (a -> Value -> Value -> ContractAssertion Unit)
+  -> ContractCheck a
+assertValueDeltaAtAddress addr check contract = do
+  valueBefore <- valueAtAddress' addr
+  ref <- liftEffect $ Ref.new Nothing
+  let
+    finalize = do
+      valueAfter <- valueAtAddress' addr
+      liftEffect (Ref.read ref) >>= case _ of
+        Nothing -> pure unit -- tellFailure $ CustomFailure "Contract did not run"
```
errfrom, line 401: This comment is slightly misleading, since AFAIU the result of the contract isn't stored only in case of an exception, so the contract did actually run, but without success. Although this presupposition may be false with user-written assertions.

errfrom (changes requested):
Partial review

src/Contract/Test/Assert.purs
jy14898:
```suggestion
    E.try run >>= \res -> finalize *> liftEither res
```

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,600 @@
...
+import Data.String.Common (joinWith) as String
+import Data.Tuple.Nested (type (/\), (/\))
+import Effect.Class (liftEffect)
+import Effect.Exception (error, throw, try)
+import Effect.Ref (Ref)
+import Effect.Ref as Ref
+import Type.Proxy (Proxy(Proxy))
+
+-- | Monad allowing for accumulation of assertion failures.
+type ContractAssertion (a :: Type) =
```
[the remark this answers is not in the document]
klntsky: Would not be consistent with `Contract`, that's why I removed the `M`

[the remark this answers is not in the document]
klntsky: In general I don't like this Haskell folklore, naming the type (by its purpose, ideally) has nothing to do with the fact it is a monad.

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,600 @@
...
+
+instance Show a => Show (ExpectedActual a) where
+  show (ExpectedActual expected actual) =
+    " (Expected: " <> show expected <> ", Actual: " <> show actual <> ")"
+
+--------------------------------------------------------------------------------
+-- Different types of assertions, Assertion composition, Basic functions
+--------------------------------------------------------------------------------
+
+-- | An check that can run some initialization code before the `Contract` is run
```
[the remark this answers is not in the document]
klntsky: Nice catch

src/Contract/Test/Assert.purs
klntsky: No, the finalization parts will still be executed.

src/Contract/Test/Assert.purs
klntsky: Indeed

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,631 @@
...
+        Just res -> f res
+  pure $ run /\ finalize
+
+assertContract
+  :: ContractAssertionFailure
+  -> Boolean
+  -> ContractAssertion Unit
+assertContract failure cond
+  | cond = pure unit
+  | otherwise = tellFailure failure
```
jy14898, line 287: I like `unless`, up to you

```suggestion
assertContract failure cond = unless cond $ tellFailure failure
```

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,631 @@
...
+          \oldHook -> Just \tx -> do
+            -- ignore possible exception from the old hook
+            void $ try $ oldHook tx
+            submitHook tx
+
+    finalize :: ContractAssertion Unit
+    finalize = do
+      exUnits <- liftEffect $ Ref.read ref
+      assertContract (MaxExUnitsExceeded (ExpectedActual maxExUnits exUnits))
+        (maxExUnits >= exUnits)
```
jy14898, line 405: This is a bug, we must compare the fields independently. For eg, the ord instance on records does the following:

```
> { a: 3, b: 1000 } > { a: 10, b: 20 }  
> false
```

src/Contract/Test/Assert.purs
```diff
@@ -0,0 +1,631 @@
...
+    let
+      actual :: BigInt
+      actual = valueToCoin' valueAfter - valueToCoin' valueBefore
+
+      unexpectedLovelaceDelta :: ContractAssertionFailure
+      unexpectedLovelaceDelta =
+        UnexpectedLovelaceDelta addr (ExpectedActual expected actual)
+
+    assertContract unexpectedLovelaceDelta (comp actual expected)
+    pure unit
```
jy14898, line 461:
```suggestion
```

jy14898 (changes requested):
Looks fine to me, I think my only comments are on code that didn't change from before actually

src/Contract/Test/Assert.purs
klntsky: Thank you!

jy14898 (approved):
LGTM

errfrom (approved):
LGTM, well done! My only concern was related to `assertContractMaybe` interrupting the execution of other assertions (especially in the context of `assertTxHasMetadata`), but since all skipped assertions are displayed to the user, I'm fine with that.

Issue #1359, opened 2022-12-22
Introduce blockfrost to devShell


klntsky:
Motivation is: we want to make blockfrost usable with Plutip. For that we need the binary to be available in the devShell, as well as everything we need to start it.

The Ask:

- add the binary to nix-shell environment
- TBD

aciceri:
Just adding the `blockfrost` binary to the `devShell` is not sufficient since it also requires a running `cardano-db-sync` (which itself requires a running Cardano node). Why can't we just use `nix run .#ctl-runtime-blockfrost` (as in #1358) that starts `docker-compose` under the hood ?
@klntsky

klntsky:
@aciceri because we need to use blockfrost that connects to a cluster that is started from purescript. Could you please put `cardano-db-sync` to PATH as well

aciceri:
@klntsky I think I raised the same concern weeks ago and you gave me this same exact answer, I had to write it down...
By the way just implemented in #1395 (binary name in `PATH` is `blockfrost-backend-ryo`, not just `blockfrost`).

klntsky:
#1395

PR #1395, opened 2023-01-13
CTL runtime with Blockforst (`nix run .#ctl-runtime-blockfrost`)

### LINKED ISSUE #1358: Wrap blockfrost as a runtime service
We need a blockfrost runtime separate from `ctl-runtime` in nix. It should start blockfrost and make its ports available.

aciceri:
Closes #1358 and #1359.
Adds a new CTL runtime that includes `blockfrost` (run with `nix run .#ctl-runtime-blockfrost`). Also add the binary `blockfrost` to `devShell`.

### Pre-review checklist

- [X] Any new API features or modification of existing behavior is covered as defined in our [test plan](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/test-plan.md) *Not involving APIs*


nix/runtime.nix
```diff
@@ -20,13 +20,26 @@ rec {
       tag = "1.35.4";
     };
     ogmios = { port = 1337; };
+    postgres = {
```
brainrake, line 23: please add docker image tag option

nix/runtime.nix
```diff
@@ -20,13 +20,26 @@ rec {
...
+      # User-facing port on host machine.
+      # Can be set to null in order to not bind postgres port to host.
+      # Postgres will always be accessible via `postgres:5432` from
+      # containers.
```
brainrake, line 27: rm comment

brainrake (changes requested):

nix/runtime.nix
aciceri: I will use a Nix generated Docker image to improve reproducibility.

aciceri:
@klntsky now if you run `nix run .#ctl-runtime-blockfrost` it starts `arion` with two more containers: `blockfrost` and `cardano-db-sync` (which is needed by `blockfrost`).
It should work with other networks too. I've seen that now one can also override the `cardano-configurations` flake input (for using custom networks I guess), in that case `blockfrost` will not work.

I've also updated `postgres` and added some `TODO`s near `ogmios`, `kupo` and `cardano-node` images definitions since I would like to define them using specific packages from flake inputs and not docker image + tag from dockerhub. But if you agree I would do this in a separate PR. I can already do this next tomorrow or next week.

aciceri:
Have I also to update the changelog?

klntsky:
@aciceri I will take care of the changelog

klntsky:
> I would like to define them using specific packages from flake inputs and not docker image + tag from dockerhub

@aciceri 
So they will be fetched from some nix cache? I'm fine with any approach that does not force us to build images.

aciceri:
> So they will be fetched from some nix cache? I'm fine with any approach that does not force us to build images.

In theory we should have also this nice corollary. But the main reason because I wanted to do that is for having same binaries built from the flake inputs everywhere in the flake, also in containers.

PR #1403, opened 2023-01-20
Fix error passing from websocket

przembot:
Closes # .

`Event` is passed where `Error` was expected.
Reading message field from `ErrorEvent`
https://github.com/pladaria/reconnecting-websocket/blob/05a2f7cb0e31f15dff5ff35ad53d07b1bec5e197/events.ts#L11
Or would it be better to pass `event.error.toString()`?

### Pre-review checklist


src/Internal/JsWebSocket.js
```diff
@@ -45,7 +45,7 @@ exports._onWsConnect = ws => fn => () => {
 
 exports._onWsError = ws => fn => () => {
   const listener = function (event) {
-    fn(event.toString())();
+    fn(event.message)();
```
klntsky, line 48: This code assumes that the thing being thrown is indeed an `Error`, but in fact any object can be thrown from JS, hence the `toString`. Note that when a message property is set properly in a class that extends Error, it gets propagated into a toString method:


```
> class A extends Error {
... constructor() {
..... super();
..... this.message = "hiii"
..... }
... }
undefined
> throw new A()
Thrown:
{ [Error: hiii] message: 'hiii' }
```

We could accept this change, but with one addition: we should use `message` only for `event instanceof Error` and only if it is a string (`typeof event.message === 'string'`), and `toString` for everything else.

klntsky (changes requested):

src/Internal/JsWebSocket.js
przembot: For some context - been debugging websocket connection issues and `event.toString()` evaluated to `[object Object]` without fixing:
https://github.com/Plutonomicon/cardano-transaction-lib/blob/87dfb2964fbdd3c8afbf22cbe6ec65e28246d2d1/src/Internal/QueryM.purs#L895-L897
Event handling on `ws` side rewraps the error with `EventError` and attaches original `Error` as `error` field of the `EventError`
https://github.com/websockets/ws/blob/3c0acc90587cde21f45c7c6fa15e234257c87d05/lib/event-target.js#L223
https://github.com/websockets/ws/blob/ea761933702bde061c2f5ac8aed5f62f9d5439ea/doc/ws.md#event-error

But given it's still JS, I'd do the instance checks for the `EventType` and fallback to `toString` if that sound reasonable to you

klntsky: It's OK to specialize it for that error layout with `EventError.error`, but we also need a fallback case and to detect the error type correctly.

przembot: Updated fallback logic

klntsky (approved):
LGTM, thanks!

CHANGELOG.md
```diff
@@ -63,6 +63,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 
 - `ConnectToNuFi` now reexported in `Contract.Wallet` ([#1435](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1435))
+- Error message from `reconnecting-websocket` is passed correctly in `Ctl.Internal.JsWebSocket._onWsError` ([#1403](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1403))
```
klntsky, line 66:
```suggestion
- Better error reporting for WebSocket errors  ([#1403](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1403))
```

PR #1417, opened 2023-01-30
Fix redeemer indexing problem in staking constraints

klntsky:
There is a problem with redeemers for stake validators: redeemers were not. in fact, reindexed in the balancer. The index in the redeemer with `Cert` type should point to the corresponding certificate in the array of certificates.

### Pre-review checklist

### DIFF
--- a/src/Internal/Types/ScriptLookups.purs
+++ b/src/Internal/Types/ScriptLookups.purs
@@ -249,14 +249,14 @@ import Ctl.Internal.Types.UnbalancedTransaction
   , emptyUnbalancedTx
   )
 import Data.Array (cons, filter, mapWithIndex, partition, toUnfoldable, zip)
-import Data.Array (singleton, union, (:)) as Array
+import Data.Array (length, singleton, union, (:)) as Array
 import Data.Bifunctor (lmap)
 import Data.BigInt (BigInt, fromInt)
 import Data.Either (Either(Left, Right), either, isRight, note)
 import Data.Foldable (foldM)
 import Data.Generic.Rep (class Generic)
 import Data.Lattice (join)
-import Data.Lens (non, (%=), (%~), (.=), (.~), (<>=))
+import Data.Lens (non, view, (%=), (%~), (.=), (.~), (<>=))
 import Data.Lens.Getter (to, use)
 import Data.Lens.Iso.Newtype (_Newtype)
 import Data.Lens.Record (prop)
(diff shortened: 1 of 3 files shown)


src/Internal/Types/ScriptLookups.purs
```diff
@@ -1321,29 +1321,30 @@ processConstraint mpsMap osMap = do
           ( scriptHashCredential $ unwrap $ plutusScriptStakeValidatorHash
               plutusScript
           )
+      index <- lift $ addCertificate cert
```
jy14898, line 1324: We have two styles of indexing in this file, maybe it would be better to follow how it's done for `mint` redeemers and call a `reindexStakeRedeemers` as we do with `reindexMintRedeemers` after building the Redeemer?

jy14898: I'm not sure which is better, however

jy14898 (commented):
For reference, as I was interested in where this behaviour is specified:

>  - indexof is a helper function that finds the index of a given certificate, value, input, or
> withdrawal in a list, finite map, or set of such objects. The implementation is not given
> directly, but we give a high-level description of how it relies on the underlying ordering.
> For each of these, it computes the index as follows:
> 
>    - certificates in the DCert list are indexed in the order in which they arranged in the
>    (full, unfiltered) list of certificates inside the transaction
> 
>    - the index of a reward account ract in the reward withdrawals map is the index of ract
>    as a key in the (unfiltered) map. The keys of the Wdrl map are arranged in the order
>    defined on the RewardAcnt type, which is a lexicographical (abbrv. lex) order on the
>    pair of the Network and the Credential.
> 
>    - the set of inputs is an ordered set (according to the order defined on the type TxIn) -
>    this also is the order in which the elements of the set are indexed (lex order on the
>    pair of TxId and Ix). All inputs of a transaction are included in the set being indexed
>    (not just the ones that point to a Plutus script UTxO)
> 
>    - the set of policy IDs is ordered according to the order defined on PolicyID (lex). The
>    index of a PolicyID in this set of policy IDs is computed according to this order. Note
>    that at the use site, the set of policy IDs passed to indexof is the (unfiltered) domain of
>    the Value map in the mint field of the transaction.
> 
> Note that in all the orderings above, long and short lex are one in the same, since for each
> of the ordered types, all the terms of that type have the same length.

Page 14-15 of [alonzo-ledger.pdf](https://github.com/input-output-hk/cardano-ledger/releases/latest/download/alonzo-ledger.pdf)


Do we have the same bug for `Reward` redeemers? To me it seems so

src/Internal/Types/ScriptLookups.purs
klntsky: No, we don't need to reindex `Cert` redeemers because we know the exact order of `Cert`s at the time of construction.

klntsky:
> Do we have the same bug for Reward redeemers? To me it seems so 

Yes, we do

klntsky:
@jy14898 let's merge this since it's a strict improvement. I made a separate issue for `Reward` redeemers due to organizational reasons.

jy14898 (approved):

PR #1438, opened 2023-02-23
bundlePursProject: don't crash if output directory exists

brainrake:
Fixes mlabs-haskell/atrium#108

### Pre-review checklist


klntsky (approved):
LGTM. It fixes the problem when `dist` is already present in the source repo, right? If so, can we change the name to something less probable to already exist as well, or better yet make it not collide with the sources at all?

brainrake:
`dist` is created during build in our example. for further customization, `buildPhase` can be overridden

PR #1478, opened 2023-03-31
Add storage for spent UTxO index

### LINKED ISSUE #1427: Consider maintaining the list of spent UTxOs as a global variable or even save it to LocalStorage
We [maintain a list of spent UTxOs](https://github.com/Plutonomicon/cardano-transaction-lib/blob/fcdd234cfe71345990f09eb1d6b4e2274faa2405/src/Internal/QueryM.purs#L340) in `ContractEnv`, but they only persist in a given value, and are not shared globally.
 
- If the user refreshes the page, they are lost
- If the user accesses another tab, they are not propagated there
- If the app is written such that there are many `ContractEnv`s, e.g. a JS SDK simply calls `runContract` every time something needs to be done, these values are not shared. This has already caused problems on the user side, e.g. see [this code](https://github.com/mlabs-haskell/singularitynet/blob/501e62df18c788a974b26dd17a3b114928ef5954/frontend/src/SdkApi.purs#L391)

We should make this piece of state as global and persistent as possible.

There are corner cases:

- LocalStorage can fill up with time. Solution: add timestamps and prune everything older than 1 day AND keep only N recent entries
- Same wallet can be used in multiple browsers

klntsky:
Closes #1427 

### Pre-review checklist


klntsky:
Closing because this idea is unfeasible. Weak guarantees for a huge increase in complexity + hard to test.

PR #1484, opened 2023-04-06
add FromBytes for Vkeywitness and Vkeywitnesses

emiflake:
This PR adds two instances for `FromBytes`: `Vkeywitness` and `Vkeywitnesses`. These instances can be made because CSL exports `.from_bytes` for these types.

### Pre-review checklist

### DIFF
--- a/src/Internal/Deserialization/FromBytes.purs
+++ b/src/Internal/Deserialization/FromBytes.purs
@@ -38,6 +38,8 @@ import Ctl.Internal.Serialization.Types
   , TransactionUnspentOutput
   , TransactionWitnessSet
   , Value
+  , Vkeywitness
+  , Vkeywitnesses
   )
 import Ctl.Internal.Types.ByteArray (ByteArray)
 import Ctl.Internal.Types.CborBytes (CborBytes)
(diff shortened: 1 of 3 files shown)


src/Internal/Deserialization/FromBytes.purs
```diff
@@ -133,6 +135,12 @@ instance FromBytes Value where
 instance FromBytes VRFKeyHash where
   fromBytes' = fromBytesImpl "VRFKeyHash"
 
+instance FromBytes Vkeywitness where
+  fromBytes' = fromBytesImpl "Vkeywitness"
+
+instance FromBytes Vkeywitnesses where
```
klntsky, line 141: This type does not have from_bytes method

klntsky (changes requested):
As a rule for any changes, please add at least one test that triggers the code.

src/Internal/Deserialization/FromBytes.purs
emiflake: Both of these do have `from_bytes`. I have tested `Vkeywitnesses`, so that one has one for sure, but the other one does too -- see the [typings](https://www.npmjs.com/package/@emurgo/cardano-serialization-lib-nodejs?activeTab=code) at line 8495. Or [the source](https://github.com/Emurgo/cardano-serialization-lib/blob/cac56c5b1493aeec2c15d71533f5f6252221f604/rust/src/crypto.rs#L551-L584).

emiflake:
> As a rule for any changes, please add at least one test that triggers the code.

Sure, I can add an example test.

src/Internal/Deserialization/FromBytes.purs
klntsky: True, I was looking at an older version that we use.

klntsky (approved):
I added the test.

PR #1519, opened 2023-06-27
Add `RefInputUnchecked` that allows referencing only with `TransactionInput`

### LINKED ISSUE #1518: Simplify `InputWithScriptRef`
`InputWithScriptRef` doesn't need to hold `TransactionUnspentOutput`. Rather, it can just hold `TransactionInput` since transaction doesn't require full copy of reference scripts.

SeungheonOh:
Closes #1518 

### Pre-review checklist

### DIFF
--- a/src/Internal/ProcessConstraints.purs
+++ b/src/Internal/ProcessConstraints.purs
@@ -170,7 +170,7 @@ import Ctl.Internal.Types.Transaction (TransactionInput)
 import Ctl.Internal.Types.TxConstraints
   ( DatumPresence(DatumWitness, DatumInline)
   , InputConstraint(InputConstraint)
-  , InputWithScriptRef(RefInput, SpendInput)
+  , InputWithScriptRef(RefInput, RefInputUnchecked, SpendInput)
   , OutputConstraint(OutputConstraint)
   , TxConstraint
       ( MustBeSignedBy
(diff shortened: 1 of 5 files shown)


src/Internal/ProcessConstraints.purs
```diff
@@ -574,19 +578,20 @@ checkRefNative
   :: forall (a :: Type)
    . InputWithScriptRef
   -> ConstraintsM a (Either MkUnbalancedTxError Boolean)
-checkRefNative scriptRef = pure $ note (WrongRefScriptHash Nothing) $ isNative
-  (unwrap (unwrap uout).output).scriptRef
+checkRefNative scriptRef =
+  case scriptRef of
+    RefInput ref -> isNative ref
+    SpendInput ref -> isNative ref
+    RefInputUnchecked _ -> pure $ Right false
```
klntsky, line 585: This is not going to work. It could as well be native.

klntsky (changes requested):

src/Internal/ProcessConstraints.purs
SeungheonOh: Made it so that when `RefInputUnchecked` is given, it will just bypass the native script check altogether

SeungheonOh:
We builds a registry for reference scripts, and having registry, we know for a fact that our reference inputs(without txOut resolved) is valid. Having option to use script reference without getting them resolved opens up some room for optimization. 

Since the name, `RefInputUnchecked`, implies that there will be no "checks" I think it's fine to skip check for native script as well

PR #1528, opened 2023-08-16
Add support for generic CIP30 wallets by name

danielfarrelly:
### Pre-review checklist

### DIFF
--- a/src/Internal/Wallet/Cip30.js
+++ b/src/Internal/Wallet/Cip30.js
@@ -6,15 +6,27 @@ exports._getUtxos = maybe => conn => () =>
   conn.getUtxos().then(res => (res === null ? maybe.nothing : maybe.just(res)));
 
 exports._getCollateral = maybe => conn => () =>
-  conn.experimental
-    .getCollateral()
-    .then(utxos =>
-      utxos !== null && utxos.length ? maybe.just(utxos) : maybe.nothing
-    );
+  /* Notes regarding the quirks of various wallets:
+
+     Yoroi will throw an error if no amount argument is provided, and will
+     break if the following expression is written as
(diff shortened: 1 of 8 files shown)


src/Internal/Wallet/Cip30.js
```diff
@@ -6,15 +6,27 @@ exports._getUtxos = maybe => conn => () =>
...
+     break if the following expression is written as
+     (conn.getCollateral || conn.experimental.getCollateral)("5000000") due to
+     JavaScript object binding
+
+     Typhon will throw an error if the amount argument is not a string
+
+     Nami only provides `getCollateral` under the experimental API
+  */
+  (typeof conn.getCollateral === "function"
+    ? conn.getCollateral("5000000")
```
klntsky, line 21: This string value is probably incorrect. In the spec, it must be cbor-string of Coin type, not numeric literal.

klntsky (changes requested):

klntsky:
We should pass the amount parameter from purescript. If we hardcode it, it must be hardcoded on PS side and then processed into cbor as usual

klntsky:
@danielfarrelly 

No, we must require it, because this is what CIP-30 tells us to do:

https://cips.cardano.org/cips/cip30/#apigetcollateralparamsamountcborcoinpromisetransactionunspentoutputnull

> The function takes a required object with parameters. With a single required parameter for now: amount. (NOTE: some wallets may be ignoring the amount parameter, in which case it might be possible to call the function without it, but this behavior is not recommended!). Reasons why the amount parameter is required: [...]

The current behavior is a bug.

klntsky:
I'm taking this PR over

klntsky:
Re-created here: #1536

PR #1538, opened 2023-09-25
scriptCurrencySymbol no longer returns Maybe

### LINKED ISSUE #1491: `scriptCurrencySymbol` doesn't need to return `Maybe`
THIS IS A DETAIL

Function
```
scriptCurrencySymbol :: MintingPolicy -> Maybe CurrencySymbol
scriptCurrencySymbol = mpsSymbol <<< mintingPolicyHash
```
from [here](https://github.com/Plutonomicon/cardano-transaction-lib/blob/10a88faa2e6237aafc90568e3488f3421517af63/src/Internal/Scripts.purs#L101)
seemingly doesn't need to have `Maybe` type.

`Maybe` there stems from `mpsSymbol` checking if provided ByteArray is empty or correctly decodes as hash. We know that it's a hash, so 
```
scriptCurrencySymbol :: MintingPolicy -> CurrencySymbol
scriptCurrencySymbol = CurrencySymbol <<< mintingPolicyHash
```
has the same semantic as long as results of `mintingPolicyHash` correctly decode with `scriptHashFromBytes` (which they should).

kozross:
Closes #1491.

### Pre-review checklist

- [X] Any new API features or modification of existing behavior is covered as defined in our [test plan](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/test-plan.md)
- [X] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included


src/Internal/Plutus/Types/CurrencySymbol.purs
```diff
@@ -98,6 +99,10 @@ mkCurrencySymbol byteArr
   | otherwise =
       scriptHashFromBytes byteArr $> CurrencySymbol byteArr
 
+-- | Bypasses the usual checks of mkCurrencySymbol: use with extreme care!
+mkCurrencySymbolUnsafe :: ByteArray -> CurrencySymbol
+mkCurrencySymbolUnsafe = CurrencySymbol
+
```
klntsky, line 105: This function is not needed anymore.

klntsky (changes requested):

src/Internal/Plutus/Types/CurrencySymbol.purs
klntsky: Is there a reason not to use the constructor?

kozross: No reason really. I thought that the hidden constructor had some purpose, and having a function with `unsafe` in the name makes it clear that you're doing something dangerous.

klntsky (approved):

PR #1565, opened 2023-11-27
Fix `redeemerHash`

klntsky:
Looks like a typo was made in the definition long time ago. Seems to not be used, hence not catched


bladyjoker (approved):
Thanks!!!

PR #1521, opened 2023-06-29
Migrate to PureScript 0.15

errfrom:
Closes #79  
Closes #1430

1. Update `purs` to `v0.15.8`
2. Update `spago` to `v0.21.0`
3. Upgrade package-set
4. Remove deprecated `purescript-math` dependency
5. Remove every occurrence of using deprecated `MonadZero` type class
6. Upgrade `medea-ps` to PureScript 0.15 https://github.com/mlabs-haskell/medea-ps/pull/1
7. Drop `sequences` dependency https://github.com/mlabs-haskell/purescript-aeson/pull/18
8. Upgrade `purescript-bignumber` to PureScript 0.15 https://github.com/mlabs-haskell/purescript-bignumber/pull/3
9. Upgrade `purescript-aeson` to PureScript 0.15 https://github.com/mlabs-haskell/purescript-aeson/pull/17
10. Upgrade `noble-secp256k1` to PureScript 0.15 https://github.com/mlabs-haskell/purescript-noble-secp256k1/pull/3
11. Upgrade `toppokki` to PureScript 0.15 https://github.com/mlabs-haskell/purescript-toppokki/pull/1
12. Convert `cjs` exports to `esm` using [lebab](https://github.com/lebab/lebab), replace `export var` with `export const`
13. Fix `EncodeAeson` and `DecodeAeson` instances for `Ctl.Internal.Plutus.Types.AssocMap`
14. Remove `ToMetadata` instance for `(NonEmpty f a)` **TODO: Can it be rewritten instead?**
15. Use `AffjaxDriver` depending on the environment (node or browser)
16. Replace `SProxy` with kind-generic `Proxy`
17. Fix `ToData` / `FromData` instances for `Ctl.Internal.Types.Interval.Extended` and `Test.Data.Tree`
18. Use `Test.QuickCheck.Gen.frequency` with `NonEmptyArray`
19. Use `exists` from `Node.FS.Sync` (`Node.FS.Aff.exists` has been removed)
20. **Milestone reached: CTL compiles without errors**
21. Use `esm` imports
22. **FIXME**: Do not re-export `Undefined` from `Contract.Prelude` (remove `undefined` dependency from `spago.dhall`)
23. Fix CIP-30 Plutip tests, use `global` object in Node.js environment
24. **Milestone reached: All unit and Plutip tests pass**
25. **Add support for bundling using `esbuild`**
[...]

PR #1567, opened 2023-12-05
Fix Arion volumes

aciceri:


aciceri:
Closing because I directly pushed the commit to #1521

PR #1585, opened 2024-01-01
Add Functionality for Sharing Wallets between Plutip Tests

MitchyCola:
Closes # .

### Pre-review checklist

- [ ] Any new API features or modification of existing behavior are covered with tests
- [ ] The template (`templates/ctl-scaffold`) has been [updated](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/development.md#maintaining-the-template)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 
 Then consult with [the template's build scripts](./templates/ctl-scaffold/esbuild/) - also see the new [Makefile](./templates/ctl-scaffold/Makefile) setup and [NPM scripts](./templates/ctl-scaffold/package.json).
 
+- Sharing wallets between Plutip tests
+  - `runPlutipTestPlan` is a new function that executes a `ContractTestPlan`.
+  - `groupContractTestPlan` is a new function that uses `Mote.Monad.group` to combine `ContractTestPlan`s.
+  - `sameWallets` is a new function that creates a `ContractTestPlan` from a `UtxoDistribution` and a `TestPlanM` of the same wallets running different `Contract`s.
+
 ### Changed
 
 - PureScript compiler version has been updated to v0.15.8. ([#1521](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1521))
--- a/src/Contract/Test/Plutip.purs
+++ b/src/Contract/Test/Plutip.purs
@@ -1,10 +1,11 @@
 -- | This module contains everything needed for `Contract` testing in Plutip
 -- | environment.
 module Contract.Test.Plutip
-  ( testPlutipContracts
+  ( PlutipTest
   , defaultPlutipConfig
   , module X
-  , PlutipTest
+  , runPlutipTestPlan
+  , testPlutipContracts
   ) where
 
 import Prelude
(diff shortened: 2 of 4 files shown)


doc/plutip-testing.md
```diff
@@ -150,8 +150,49 @@ See the comments in the [`Ctl.Internal.Plutip.Server` module](../src/Internal/Pl
...
+To execute tests that share the same wallet state, the use of `Contract.Test.Plutip.runPlutipTestPlan` is suggested, which has a type of:
+
+```purescript
+runPlutipTestPlan
+  :: PlutipConfig
+  -> ContractTestPlan
+  -> TestPlanM (Aff Unit) Unit
+```
+
+`runPlutipTestPlan` uses the exact same logic to perform `testPlutipContracts`, except it forgoes the use of `execDistribution` to create a `ContractTestPlan` from a `TestPlanM ContratTest Unit`. Instead, a `ContractTestPlan` is expected as the second parameter of the function, meaning that the test plan will need to be created before using `runPlutipTestPlan`.
```
klntsky, line 163: Please rephrase it in a way that makes it clear to the user who does not know what `execDistribution` means. Just phrase it in a way that makes it clear when funds pre-allocation happens and how it is used.

src/Internal/Test/ContractTest.purs
```diff
@@ -60,6 +74,30 @@ newtype ContractTestPlan = ContractTestPlan
...
+instance Semigroup ContractTestPlan where
+  append
+    (ContractTestPlan runContractTestPlan)
+    (ContractTestPlan runContractTestPlan') =
+    do
+      runContractTestPlan \distr tests -> do
+        runContractTestPlan'
+          \distr' tests' -> ContractTestPlan \h -> h (distr /\ distr') do
+            mapTest (_ <<< fst) tests
+            mapTest (_ <<< snd) tests'
```
klntsky, line 86: I don't think this is a valid semigroup: `/\` is not associative. From the point of view of the user, sure. But not *actually*. We are using it only in one place, so please make it a function.

CHANGELOG.md
```diff
@@ -87,6 +87,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 Then consult with [the template's build scripts](./templates/ctl-scaffold/esbuild/) - also see the new [Makefile](./templates/ctl-scaffold/Makefile) setup and [NPM scripts](./templates/ctl-scaffold/package.json).
 
+- Sharing wallets between Plutip tests ([#1585](https://github.com/Plutonomicon/cardano-transaction-lib/pull/1585))
+  - `runPlutipTestPlan` is a new function that executes a `ContractTestPlan`.
+  - `groupContractTestPlan` is a new function that uses `Mote.Monad.group` to combine `ContractTestPlan`s.
```
klntsky, line 92:
```suggestion
  - `groupContractTestPlans` is a new function that uses `Mote.Monad.group` to combine `ContractTestPlan`s.
```

src/Internal/Test/ContractTest.purs
```diff
@@ -60,6 +74,30 @@ newtype ContractTestPlan = ContractTestPlan
...
+
+-- | Group `ContractTestPlans` together, so that they can be ran in the same Plutip instance
+groupContractTestPlans
```
klntsky, line 89: I question the usefulness of this function. It only allows two `ContractTestPlan`s to be joined, but requires a title. In real life scenarios, will users be forced to provide 2 titles to join 3 tests?

src/Contract/Test/Plutip.purs
```diff
@@ -41,6 +47,13 @@ testPlutipContracts
   -> MoteT Aff (Aff Unit) Aff Unit
 testPlutipContracts = Server.testPlutipContracts
 
+-- | Run `ContractTestPlan` tests in a single Plutip instance.
+runPlutipTestPlan
+  :: PlutipConfig
+  -> Server.ContractTestPlan
+  -> MoteT Aff (Aff Unit) Aff Unit
+runPlutipTestPlan = Server.runPlutipTestPlan
```
klntsky, line 55: Looks like a plain re-export?

doc/plutip-testing.md
```diff
@@ -150,8 +150,49 @@ See the comments in the [`Ctl.Internal.Plutip.Server` module](../src/Internal/Pl
...
+    group "Test Plan" do
+      test "Test 1"  \(alice /\ bob /\ charlie) -> do
+          ...
+
+      test "Test 2"  \(alice /\ bob /\ charlie) -> do
+          ...
+
+      test "Test 3" \(alice /\ bob /\ charlie) ->  do
+          ...
+```
```
klntsky, line 195: Please add a runnable example to the tests. Ideal location would be `test/Plutip/SameWallets.purs`. Then link this file here. For example contracts to use in test bodies, see `Ctl.Examples.Pkh2Pkh` - just plug it multiple times.

To actually run it in CI, plug the test suite into `test/Plutip.purs`.

src/Internal/Test/ContractTest.purs
klntsky: Maybe we want an array of tests, grouped under a common title?
But then, couldn't the user just use

```
 runPlutipTestPlan config do
  sameWallets distribution $
     group "foo" do
        test "bar" do ...
        test "baz" do ...
```

?

I see no advantage in using an array, so maybe we should just remove this function?

MitchyCola: That method works, and that is what I actually have been using. This is was an idea I had that I carried over, but I will remove this since I don't actually use it myself.

src/Contract/Test/Plutip.purs
MitchyCola: I just followed the exporting of `testPlutipContracts` as en example, should that just be a re-export as well?

klntsky (approved):
Thanks!

PR #1613, opened 2024-04-23
Revise ctl-scaffold

euonymos:
### Motivation

Many projects need to implement web API components with CTL that are supposed to be integrated within a website. A simple scaffold example is a must-have thing to promote CTL's usage. I am not sure about preserving other things that were here previously, let's discuss. Our project doesn't require them and I have doubts we can afford re-integrating them back within our budget.

### Questions and Tasks

- [ ] Do we need `liqwid-nix` indeed?
- [ ] Place an actual React app rather than just an empty page
- [ ] Hooks to build and push `dist`/NPM package?
- [ ] A test-suite example?
- [ ] Back-end service example (or another template for that?)
- [ ] Restore headless browser test suite?

### Pre-review checklist


klntsky:
> Do we need liqwid-nix indeed?

Context?

> Place an actual React app rather than just an empty page

I'm strongly against this. People should choose their frameworks as they wish. There's no point in maintaining react code here.

> Hooks to build and push dist/NPM package?

I would accept a PR for that.

> A test-suite example?

Plutip:

https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/templates/ctl-scaffold/test/Main.purs

> Back-end service example (or another template for that?)

Not every dApp has a backend. I'm strongly against this in CTL.

> Restore headless browser test suite?

It didn't go anywhere. Here's the entry point: 
https://github.com/Plutonomicon/cardano-transaction-lib/blob/eccf2705f4b6ce590b648a8f28ba095c36440433/templates/ctl-scaffold/package.json#L17

klntsky (changes requested):
Hey! What are you trying to achieve? I would not merge anything that makes the template more opinionated than it is, i.e. no react or any other particular framework.

Moreover, the template should be kept in sync with CTL's own setup. I would not merge this, because you are basically asking me and the future CTL devs to maintain twice more code.

Have you seen https://github.com/mlabs-haskell/the-plutus-scaffold ? This is a separate project created specifically to showcase CTL+(Plutus|plutarch)+react, I really think you should work with it instead.

klarkc:
CTL itself should not have a scaffold; I mean, in this repository, it should instead refer to example repositories, like [LovelaceAcademy/nix-templates](https://github.com/LovelaceAcademy/nix-templates), because you can use CTL in many different ways. Having a scaffold pushes the user to use it in a singular, optioned way that may not match their needs.

klntsky:
Closing this PR with the motivation stated above

PR #1641, opened 2024-08-06
Expose Testnet API

marcusbfs:
### Pre-review checklist

- [X] Any new API features or modification of existing behavior are covered with tests
- [X] The template (`templates/ctl-scaffold`) has been [updated](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/development.md#maintaining-the-template)

### DIFF
--- a/src/Contract/Test/Testnet.purs
+++ b/src/Contract/Test/Testnet.purs
@@ -26,13 +26,24 @@ import Ctl.Internal.Testnet.Contract
   , runTestnetTestPlan
   , testTestnetContracts
   ) as X
-import Ctl.Internal.Testnet.Types (Era(Babbage), TestnetConfig) as X
+import Ctl.Internal.Testnet.Types (Era(Babbage), TestnetConfig)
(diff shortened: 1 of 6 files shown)


src/Contract/Test/Testnet.purs
```diff
@@ -26,13 +26,13 @@ import Ctl.Internal.Testnet.Contract
   , runTestnetTestPlan
   , testTestnetContracts
   ) as X
-import Ctl.Internal.Testnet.Types (Era(Babbage), TestnetConfig)
+import Ctl.Internal.Testnet.Types (Era(Babbage), TestnetConfig) as X
```
klntsky, line 29: I think, if the goal is to re-export, you should export the other constructors too. In general, it's best to avoid using `X.` imports in the module (OK to import the same module  unqualified)

klntsky (changes requested):

klntsky (approved):

PR #1648, opened 2024-09-06
Release v9.3.0

errfrom:


package.json
marcusbfs, line 1: `@emurgo/cardano-serialization-lib-browser` v12.0.1 is out with a memory management fix. Might be worth to upgrade.

klntsky: we fix it on the cardano-types side. don't forget to remove it. 

https://github.com/mlabs-haskell/purescript-cardano-types/blob/71b204c7c9c83b8280ed3ed14837d76b8ac0a6fe/src/Cardano/Types/Mint.purs#L161

klntsky (approved):
LGTM

PR #1668, opened 2025-01-23
Ignore any tx evaluation errors if tx marked invalid

errfrom:
### Pre-review checklist

- [x] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Changed`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included


marcusbfs (approved):
Is it possible to add a test case for this?

errfrom:
@marcusbfs Only if we manage to reproduce an internal tx evaluation error somehow. I'm not sure it's worth the effort since I've encountered this issue with Blockfrost and it may not be as easily reproducible in the cardano-testnet environment.

PR #1680, opened 2025-04-08
Pluggable Balancers

### LINKED ISSUE #1621: Plugable balancer strategies
Some dApps may need to use specific balancing strategies that cannot be expressed using balancing constraints. In Cardano balancers can utilize different strategies most of which are designed to prevent dust, and they have their pros and cons. One example is a wallet used by a backend service to perform some administrative operations that we want to run in parallel. The ability to provide a custom strategy looks very attractive for cases like that.

errfrom:
Closes [Move balancer to a separate package with public API #1621
](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1621)

Extracted the default CTL transaction balancer into its own package: [CTL Tx Balancing & Coin Selection #2](https://github.com/mlabs-haskell/purescript-cardano-transaction-balancer/pull/2)

### Pre-review checklist

- [x] The changelog has been updated under the `## Unreleased` header, using the appropriate sub-headings (`### Added`, `### Changed`, `### Removed`, `### Fixed`), and the links to the appropriate issues/PRs have been included


marcusbfs (changes requested):
Looking great.  Two things:
- the test `Evaluation with additional UTxOs with native scripts` is failing
- A new version of the [packaget-set](https://github.com/mlabs-haskell/purescript-cardano-package-set) containing the balancer package must be released. Then, update CTL (and the scaffold template) to use it

marcusbfs (approved):
LGTM

src/Contract/Transaction.purs
```diff
@@ -304,8 +326,28 @@ balanceTxs unbalancedTxs =
       withUsedTxOuts <<< unlockTransactionInputs <<< _.transaction
     throwError e
 
+balanceMultipleTxs
```
KristianBalaj, line 331: Missing documentation comment for this function.

src/Contract/Transaction.purs
```diff
@@ -315,6 +357,18 @@ balanceAndLock { transaction, usedUtxos, balancerConstraints } = do
   void $ withUsedTxOuts $ lockTransactionInputs balancedTx
   pure balancedTx
 
+balanceAndLockUtxos
```
KristianBalaj, line 366: Also, add some comment to this one, too. Public-facing API should have documentation comments.

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -1,936 +1,92 @@
...
-  { strategy :: SelectionStrategy
-  , transaction :: Transaction
-  , changeAddress :: Address
-  , changeDatum :: Maybe OutputDatum
-  , allUtxos :: UtxoMap
-  , utxos :: UtxoMap
-  , miscFee :: BigInt -- can be negative (deregistration)
+type CtlBalancerContext =
+  { balancerConstraints :: BalancerConstraints
+  , extraUtxos :: UtxoMap
```
KristianBalaj, line 54: What's the meaning of `extraUtxos` field? I would prefer some documentation comment with this field.

I'd expect it means additional UTXOs, however, that has a different meaning and is set via the `balancerConstraints`. So I need further explanation for this... :thinking:

src/Internal/BalanceTx/BalanceTx.purs
```diff
@@ -1,936 +1,92 @@
...
-
-getRequiredValue
-  :: BigInt -> UtxoMap -> TransactionBody -> Either BalanceTxError Val
-getRequiredValue miscFee utxos txBody = do
-  getInputVal utxos txBody <#> \inputValue ->
-    ( outputValue txBody <> minFeeValue txBody <> Val miscFee Map.empty
+-- | Balances an unbalanced transaction using the specified balancer
+-- | constraints.
+defaultBalancerErr :: CtlBalancer BalanceTxError
+defaultBalancerErr transaction ctx = do
```
KristianBalaj: A nitpick:
I'd prefer a slightly different name, this has semantics of "Default balancer error" instead it should have a meaning of something like "Default balancer with error".

KristianBalaj (changes requested):
The default balancer is in the `Cardano.Transaction.Balancer` AFAIU. But, it sounds too general namespace for me. Instead, it should be something more specific IMO, that describes it better.

KristianBalaj (approved):

